PR #1823, opened 2012-10-12
"Show in Tree" command

peterflynn:
- General purpose ProjectManager.showInTree() API (any file/folder)
- "Show in Tree" command (current file)
- Utility for searching arrays (new CollectionUtils module): used in new code plus to clean up an older DocumentManager method


ychaikin:
"Show in File Tree" seems to be a something you have to go to over and over. Is there a way to make it a toggle, so it's always on if you want that way. Currently, the highlighting switching to the "Working Files", so it's no longer "in File Tree". To get it back there, you have to manually go to Navigate->Show in File Tree.

PR #3144, opened 2013-03-15
Keyboard Shortcut for JSLint

RodrigoEspinosa:
Shortcut for JSLint as "Ctrl-Shift-J"


sohailsiddiq:
'console' was used before it was defined. | console.log('Hello World!!!');
-- | --

redmunds:
Use `window.console.log('Hello World!!!');`

PR #3240, opened 2013-03-25
Wrap Selection with Tag

TomMalbran:
This adds the feature mentioned in this [Trello Card](https://trello.com/card/wrap-selection-w-html-tags/4f90a6d98f77505d7940ce88/811) in a similar way as Sublime Text 2 implements it.

Edit -> Wrap Selection with Tag or Alt-Shift-W adds a `<p>` and a `</p>` around the selection and sets the selection on the first `p`. Every edition made after this will synchronize the content between the open and close tag, using the content of the open tag up to the first space until the cursor leaves the open tag.


deepu105:
this extension seems to do it well

https://github.com/pedelman/brackets-surround

PR #4341, opened 2013-06-25
Call new shell api setModal to enforce the modality of a JS dialog.

RaymondLim:
This pull request goes with another one in brackets shell https://github.com/adobe/brackets-shell/pull/268.

Also remove all the hacks that we put in when switching to native menus. So after this pull request landed, we need to regress #2616, #3152, #3219 and #418.


RaymondLim:
Closing for now.

PR #5316, opened 2013-09-24
Guard against concurrent array mutation in Async.doInParallel

iwehrman:
If the array of items processed by `Async.doInParallel` is augmented concurrently, the returned master promise may neither resolve nor reject. This is because resolution and rejection happen only when the total number of resolved or rejected promises reaches the current size of the array of items, which, if augmented, is now larger than the total number of promises. Of course, failure of some sort is unsurprising with concurrent modification, but this simple change---counting down from the initial array length instead of counting up to the current array length---avoids this particular problem and is also slightly more efficient.

This PR also removes code that creates an array with the generated promises and then promptly throws it away.

### DIFF
--- a/src/utils/Async.js
+++ b/src/utils/Async.js
@@ -88,20 +88,17 @@ define(function (require, exports, module) {
      * @return {$.Promise}
      */
     function doInParallel(items, beginProcessItem, failFast) {
-        var promises = [];
         var masterDeferred = new $.Deferred();
         
         if (items.length === 0) {
             masterDeferred.resolve();
             
         } else {
-            var numCompleted = 0;
+            var numRemaining = items.length;
             var hasFailed = false;
             
             items.forEach(function (item, i) {
                 var itemPromise = beginProcessItem(item, i);
-                promises.push(itemPromise);
-                
                 itemPromise.fail(function () {
                     if (failFast) {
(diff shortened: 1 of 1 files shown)


busykai:
is this still on? for both cases in question (`doInParallel()` and `doSequentially()`) why not just make a copy of the array immediately?

peterflynn:
I started writing some unit tests for this and discovered that the solution here for `doInParallel()` isn't complete, since `Array.forEach()` itself is not safe with concurrent mutation either -- see http://jsfiddle.net/c130tkbm/ for some examples.

So I think the only real way to reliably fix this is to clone the array from the start.  That seems fine since the cost of cloning that would be far overshadowed by the overhead of creating Promises on each iteration.

Question is: do we want to bother?  Given that synchronous loops aren't safe, how likely is it for developers to assume these async variants are?

peterflynn:
For posterity, here's the draft unit tests: https://gist.github.com/peterflynn/17694e1d2f0e9d9151c9

peterflynn:
We talked about this in bug review yesterday and decided to close it for now -- this issue isn't something we necessarily must fix, and if we did the code changes would have to be different from these anyway (due to the Array.forEach issue).  Sorry this dragged on so long before reaching a decision :-/

PR #6317, opened 2013-12-24
Insert space after colon for CSS code hints

stowball:
Everybody writes a space after the colon in CSS properties, but Brackets doesn't insert them by default. This fixes that.


pi7dets:
Here, I found the guilty ones.  Just because Dreamweaver uses Brackets I can't upgrade to the new version for about 6 years because now there is always a space after a colon in CSS.
To put a space or not, this should be in the settings.
After all, maybe those who have been writing code for 2-3 years haven't noticed it, but I've been doing it since the first version of DW and all these new things with spaces are very annoying.
Thank you.

PR #6325, opened 2013-12-27
Resize QuickOpen results if the window size is changed, and add scroll bars

ishan1608:
These changes add a scroll-bar to the results of QuickOpen when there are more results that can fit in the screen.
It also moves the results horizontally with the QuickOpen input area.

This addresses issues #1156 and #1157

Result :
15a2214

### DIFF
--- a/src/search/QuickOpen.js
+++ b/src/search/QuickOpen.js
@@ -70,6 +70,11 @@ define(function (require, exports, module) {
      * The currently open quick open dialog.
      */
     var _curDialog;
+    
+    /**
+     * The width of the window.
+     */
+    var _windowWidth;
 
     /**
      * Defines API for new QuickOpen plug-ins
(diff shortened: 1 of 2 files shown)


busykai:
It sounds like this PR should be re-worked on top of #7227 (branch `pflynn/better-quicksearch-field`) which [unfortunately] is not yet merged. This one, however, could be closed since #7227 changes the approach quite radically (and should be landed at some point).

@ishanatmuz, do you have time to submit a new PR as @peterflynn [suggests](https://github.com/adobe/brackets/pull/6325#issuecomment-43277001)?

ficristo:
@ishanatmuz are you willing to setup a new PR with this fix?

ishanatmuz:
@ficristo I was caught up with other things in my life that I completely forgot about this. Let me see what I can do.

PR #6409, opened 2014-01-08
Switch language / syntax mode of current document

peterflynn:
This is an update of @JakeStoeffler's PR #4276.  It's merged up with the latest master and addresses the main code review comments I made there.  I also added some simple unit tests.

I still have two concerns about merging this immediately, though:

**1)** The dropdown UI doesn't look great. It uses a native `<select>`, which is just hard to style... and the way the statusbar is laid out makes it look misaligned in this case:
language picker

It would look nicer if we used an HTML popup like code hints, context menus, and the Recent Projects dropdown... but it would definitely add a bunch more work here.  So maybe we can skate by with the current `<select>`-based UI for now.  @larz0, any thoughts?

**2)** This foregrounds all the bugs in #2911 a lot more.  I added a checklist of issues there.  Are there any that are prominent enough to act as blockers here?  (JS code hints seems like it might be the most glaring issue, but unfortunately that's one of the harder ones to fix too).


jamesharrington:
Hey what is the status on this i was just thinking about making a extension to auto save the untitled doc to my temp folder so i could use highlighting. If this is coming soon i wont waste my time. Thanks for all the awesome work you guys are doing!

sprintr:
This has been merged a while ago. Secondly it doesn't deal with untitled files.

PR #6854, opened 2014-02-12
Add browser zoom commands

jasonsanjose:
Requires https://github.com/adobe/brackets-shell/pull/423

Allows the full brackets UI to be zoomed with Alt+ and Alt- and restored with Alt-0.


JeffryBooher:
@jasonsanjose do you want to close this or move forward with it?

jasonsanjose:
Ah, I can take this on now that the new CEF landed. Will update soon.

jasonsanjose:
From @JeffryBooher 

>I think we should add this global zoom to the global prefs and remember it between instances. We should also remember the editor font-size. Visual Studio remembers this setting by filename since each tab is a new instance of the document object. But it is remembered for that file between instances and is a project specific pref.

Good idea. Will do.

JeffryBooher:
@jasonsanjose sometimes the zoom gets stuck and i can't make it go any larger (even though it can be made larger) but I can make it smaller and only get back to a certain spot.  The same occurs when going the other way as well.  I haven't been able to nail it down yet.

JeffryBooher:
also, if you zoom all the way out or all the way in then try to keep going -- say you invoke the zoom out until it's reached its maximum zoom and then invoke the command 3 more times and it doesn't do anything.  You will then have to invoke the zoom in command 3 times before it starts to zoom in.

PR #6888, opened 2014-02-16
Added smart indent and auto close tags as preferences

TomMalbran:
More Editor preferences to configure using the new Preference System only.

### DIFF
--- a/src/editor/Editor.js
+++ b/src/editor/Editor.js
@@ -76,16 +76,19 @@ define(function (require, exports, module) {
         _                  = require("thirdparty/lodash");
     
     /** Editor preferences */
-    var USE_TAB_CHAR      = "useTabChar",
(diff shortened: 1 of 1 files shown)


berriganator:
I cannot seem to get the .json syntax correct to turn the auto indentation off. It's driving me nuts. Can you please write a better guide for those of us unaccustomed to editing .json files to control these features? The samples given in the Wiki are not straightforward to newbies. Thanks.

redmunds:
@berriganator There's an extension called "JSONLint" that will tell you if you have any syntax errors in a .json file. Works great!

There's also a "Brackets Preferences" extension that provides a UI for preferences, but it's very early in development, so I'm not sure how well it works.

berriganator:
Thanks! I don't know what changed since I copied and pasted the exact same text, but somehow it is now working. *mysteries*

PR #6911, opened 2014-02-18
Add proxy support for extension installation.

dangoor:
Configurable via a "proxy" preference (which could be used for other parts of
Brackets that should use HTTP proxy).

This fixes #6241, #5958 and [this backlog item](https://trello.com/c/SEV95fdK/902-support-for-a-proxy-in-extensions-installation)


brunowego:
:+1:

PR #6924, opened 2014-02-19
Update strings.js

wpt:
fix ru version


foolip:
@ingorichter, this is a bit unusual, but I wonder if you might know who @wpt is and how to get in touch? See https://github.com/wpt/myvimrc/issues/1 for some context.

PR #6975, opened 2014-02-22
add node.js 0.11 to travis-ci

Ramoonus:


ghostoy:
@dangoor Now Node.js 0.12 is out. Should you add support for it?

dangoor:
@ghostoy I haven't seen any discussions yet about plans for upgrading to node 0.12 or io.js. That's definitely something to think about.

prksingh:
There is a trello card for this: 

https://trello.com/c/DJwcfxyh

Also as mentioned by @peterflynn in the card, it would be good to knowthe potential benefits for upgrading as it would require significant work and could lead to multiple new issues

Issue #6904, opened 2014-02-18
Pressing [Enter] in Extension Manager Search box closes Extension Manager


ashmind:
**Steps to reproduce:**

1. Open Extension Manager
2. Place focus in the Search box
3. (Optional) Type any text
4. Press Enter

**Expected result:**
Nothing happens. Since a lot of search boxes in web and desktop only search on Enter, it is natural to press Enter after typing. If the search is immediate, Enter should do nothing.

**Actual result:**
Extension Manager closes without doing anything, so I have to reopen it and re-type my search.

**Environment:**

* Windows 7
* Brackets Sprint 36 experimental build 0.36.0-11506 (release 942505c3a)

lostlime:
+1

FezVrasta:
Any news on this?

abose:
Tagging @rroshan1 for extension manager improvements.

niklasHagner:
+1

Denisov21:
+1

petetnt:
Fixed in #12839

PR #7337, opened 2014-03-27
Fix for #6904 - Pressing [Enter] in Extension Manager Search box closes Extension Manager

ingorichter:
This will prevent that the Extension Manager Dialog will be closed when the `Enter` key is hit inside the search field.


busykai:
FWIW, we also have `File -> Project Settings` and `View -> Themes`. In the `Project Settings` case, closing on enter seems natural -- it only has one field. In case of `Themes` (which wasn't there at the time of the discussion above), it does not feel right to close the dialog when you focus on, for example, `Font Size` and hit enter. So it sounds like a special "marker" class which would make Dialog to ignore Enter would be a nice solution (if I understand it right, it is what @TomMalbran suggests). Extensions with complex configuration could benefit from it, e.g. Brackets git.

Another question is what's the right behavior should be: should it just stay on the input field and do nothing, should it switch the focus to the default button, should it act as tab? Should there be different classes for different behaviors?..

rroshan1:
It would be good if we could reach some consensus on the fix for this issue. This issue is experienced by many and it has been open for quite long now.

ficristo:
Superceded by #12839

PR #8658, opened 2014-08-05
Restore 'browser: false' to Brackets core's JSLint setup

peterflynn:
Fix Brackets project settings so `browser: true` is not set (which is how we had it before 834c4e9 in Sprint 37). Fix a few JSLint errors that snuck in because of this, plus some additional errors in ExtensionLoader that seem to have come from the Themes work in 0.42

### DIFF
--- a/src/extensions/default/InlineTimingFunctionEditor/TimingFunctionUtils.js
+++ b/src/extensions/default/InlineTimingFunctionEditor/TimingFunctionUtils.js
@@ -329,7 +329,7 @@ define(function (require, exports, module) {
             if (match && _validateCubicBezierParams(match)) {
                 return _tagMatch(match, BEZIER);
             } else { // this should not happen!
-                window.console.log("brackets-cubic-bezier: TimingFunctionUtils._getValidBezierParams created invalid code");
+                console.log("brackets-cubic-bezier: TimingFunctionUtils._getValidBezierParams created invalid code");
             }
         }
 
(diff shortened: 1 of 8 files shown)


nethip:
@peterflynn LGTM. Is `"predef": ["window", "setTimeout", "clearTimeout"]` the only change that is pending for this PR to go to master?

src/extensions/default/InlineTimingFunctionEditor/TimingFunctionUtils.js
```diff
@@ -329,7 +329,7 @@ define(function (require, exports, module) {
             if (match && _validateCubicBezierParams(match)) {
                 return _tagMatch(match, BEZIER);
             } else { // this should not happen!
-                window.console.log("brackets-cubic-bezier: TimingFunctionUtils._getValidBezierParams created invalid code");
+                console.log("brackets-cubic-bezier: TimingFunctionUtils._getValidBezierParams created invalid code");
```
MiguelCastillo: Why are we removing `window` from this one?

src/extensions/default/InlineTimingFunctionEditor/TimingFunctionUtils.js
```diff
@@ -398,7 +398,7 @@ define(function (require, exports, module) {
             if (match && _validateStepsParams(match)) {
                 return _tagMatch(match, STEP);
             } else { // this should not happen!
-                window.console.log("brackets-steps: TimingFunctionUtils._getValidStepsParams created invalid code");
+                console.log("brackets-steps: TimingFunctionUtils._getValidStepsParams created invalid code");
```
MiguelCastillo: And... Why are we removing `window` from this one?

MiguelCastillo:
@peterflynn I really like this idea... It will remove most the jslint definitions in source files, which I personally think isn't the best place to add those settings.

Why haven't we merged this in yet?

ficristo:
In #11693 I am converting Brackets core code to ESLint.
Want to land this patch before it?

ficristo:
Fixed in https://github.com/adobe/brackets/pull/12661

PR #8846, opened 2014-08-24
Update Italian (it) strings.js

chrispecoraro:
This is a more clear explaination confirmed by a native speaker.
For "REPLACE_WITHOUT_UNDO_WARNING", if the {0} placeholder is used, then the grammar is too difficult.
Example:  
Since more than (one) file NEEDS to be modified.
Since more than (two) file NEED to be modified.

To resolve this I used Since more than one file needs to be modified, since it satifies 1 up to n files.


JeffryBooher:
Hi @chrispecoraro I think maybe this got lost in all the holiday hustle. Is there any way you can get this fixed up in the next week or so? We'd like to get your work into the next release. Thanks!

chrispecoraro:
Are there new strings to add? Should I do a fresh "diff" with the English/source file?

chrispecoraro:
The changes in this pull request have already been added to master

PR #8929, opened 2014-09-01
Toggle Line Comment - removing leading whitespace breaks Jade preprocessor #8012

mccricardo:
Fixing issue #8012 concerning toggle line comment for Jade files. 

Jade comments should be placed on the beginning and not on the beginning of the line.

Two tests were written to test comment and uncomment.


src/editor/EditorCommandHandlers.js
```diff
@@ -237,6 +241,27 @@ define(function (require, exports, module) {
...
+    /**
+     * @private
+     * Given a line to be commented it will determine where the comment prefix position is.
+     * 
+     * @param {string} language id
+     * @param {string} line to be commented
+     * @return {integer}
+     */
+    function _getPrefixPos(language, line) {                        
+        if (language === "jade") {
```
peterflynn: EditorCommandHandlers shouldn't have hardcoded knowledge of specific languages -- this should be a general flag, e.g. `Language.indentLineComments`.  I'd imagine this same issue affects braceless/classic SASS too, for example.

src/editor/EditorCommandHandlers.js
```diff
@@ -207,7 +209,9 @@ define(function (require, exports, module) {
         if (containsNotLineComment) {
             // Comment out - prepend the first prefix to each line
             for (i = startLine; i <= endLine; i++) {
-                editGroup.push({text: prefixes[0], start: {line: i, ch: 0}});
+                // For each line we need to determine the position where de comment prefix should be inserted
+                prefixPos = _getPrefixPos(language, doc.getLine(i));
```
peterflynn: The `language` var was removed from this function more recently, so you'll need to re-add it

src/editor/EditorCommandHandlers.js
```diff
@@ -207,7 +209,9 @@ define(function (require, exports, module) {
...
+                editGroup.push({text: prefixes[0], start: {line: i, ch: prefixPos}});
             }
 
             // Make sure tracked selections include the prefix that was added at start of range
```
peterflynn: It looks like this code that deals with `trackedSel` contains assumptions about the comment being inserted at `ch: 0`.  You may need to update it to ensure it works as intended in the new case too.

mccricardo:
Closing due to inactivity from my end.

PR #8950, opened 2014-09-03
Updates the same URLs in all the other locales of the Getting Started project.

wormeyman:
In Pull Request #8606 I updated all the URL's in the root getting started, @peterflynn asked that I do the same for the other languages.

A few things I am thinking about:
- There is not a consistent coding style across the getting started files. If you would like i could use [Beautify](https://github.com/drewhamlett/brackets-beautify) & [whitespace-normalizer](https://github.com/dsbonev/whitespace-normalizer)
  or something else to make the HTML and CSS consistent.
- I can only read latin languages so on some i guessed where to place the IRC url.
- I mostly did a find and replace so there is a good chance i missed something.


peterflynn:
@wormeyman Sorry for the prolonged delay.  I just rechecked this and there are a few more changes we'd need:

* Please revert the changes to 'ru_utf8.html' and 'ru_utf8_wBOM.html' -- these are for unit tests and need to remain unaltered.
* Please revert the whitespace changes on blank lines -- we want the whitespace in these HTML files to match what happens when you edit the code in Brackets.
* Please squash into a single commit (14 is a lot for a change like this).
* _Potentially_ - It looks like a few locales are missing the https://brackets-registry.aboutweb.com link.  for consistency I wonder if we should add those even if the link titles are untranslated for now.  Anyone strongly object?

wormeyman:
Hey @peterflynn I'm also sorry for the long delay when you replied we had just had our second child, i really don't know if I will ever be able to get around to making those changes, especially because my brackets fork is completely messed up and out of date. These changes might be better broken up into separate issues tagged with starter bug I would close this myself but if you want to re assign it feel free otherwise i think we should close this pull request.

PR #8977, opened 2014-09-05
[HOLD] Sort project list alphabetically

brunnopleffken:
Really useful and missing feature if you are used to switch between multiple projects in a matter of minutes (like me) and have a lot of them. It sounds silly, but may save you some precious time and patience.


le717:
@brunnopleffken I would like to see this in core, but the remaining issue (needs to be a pref) still needs to be addressed. Can you make that change soon?

prafulVaishnav:
@brunnopleffken Hii.. Just wanted to know if you got the time to add the preference for it, so that it can be merged..

Issue #6822, opened 2014-02-10
Sidebar can't be resized once filling whole window


FezVrasta:
If you expand the sidebar to take the entire width of the brackets' window, you can't resize it again to the original size...

![image](https://f.cloud.github.com/assets/5382443/2124175/5c50921a-9239-11e3-8743-1c4f85ed040b.png)

solugebefola:
Has this issue been resolved?

demonofthemist:
As of today this issue is still there, I just tried it on my computer :)

PR #9122, opened 2014-09-16
Bug #6822. Provided maxSize property for the sidebar and added a check in resizer.js to convert maxSize into pixels if provided in percentage.

prateekagr98:
Provide maxSize to be 45% of window width if element is side bar and maxSize is not specified

This is for #6822.


ficristo:
@redmunds do you know if this is still relevant?

redmunds:
@ficristo Issue #6822 is still open, so I think this is still relevant. The contributor seems to have addressed all comments (although didn't add a "ready for another review" comment), but I have not verified the fix.

One change I would like to see is to have the max be larger (something like 80%). It might not seem useful to only show 20% of window for editor, but you might want to temporarily expand to find a particular file.

src/htmlContent/main-view.html
```diff
@@ -40,7 +40,7 @@
 
     <!-- Main UI: horizontal set of sidebar, main content vertical stack, and vertical toolbar -->
     <div class="main-view">
-        <div id="sidebar" class="sidebar panel quiet-scrollbars horz-resizable right-resizer collapsible" data-minsize="0" data-forceleft=".content">
+        <div id="sidebar" class="sidebar panel quiet-scrollbars horz-resizable right-resizer collapsible" data-minsize="0" data-maxsize="45%" data-forceleft=".content">
```
redmunds: I think max should be large. Maybe 80% ?

ficristo:
Thank you @redmunds
@prateekagr98 are you still interested? Could you update with the suggestion of @redmunds?

ficristo:

PR #9396, opened 2014-10-01
rename update-release-number grunt task to set-release

ingorichter:
Use the same shorter task name that we use on brackets-shell.


ficristo:
The task `set-release` already exists which is a combination of `update-release-number, write-config`.
Is this still necessary?

ficristo:
Closing in favour of #13101 issue

PR #9622, opened 2014-10-21
Remove deprecated file system APIs

le717:
Towards #8751.

Removes APIs listed in #6127. As you can see in that issue, there are a number of extensions that access it, but it's nearly a year since they were deprecated. It's time to remove them.


le717:
This PR has been coming up on a year open and two years since the old file system was deprecated. Any extensions that are not updated by now will probably never be updated. Could someone perhaps look at this and see about getting it merged?

zaggino:
Agreed @le717 , lets clear this. Merging into master, will take effect from 1.6 (1.5 is now on release branch)

Issue #9654, opened 2014-10-23
Untitled documents aren't added to MRU until visited a 2nd time


peterflynn:
1. Open a few files in the working set (let's call them A, B, and C)
2. File > New
3. Click back to the last file you'd been in (C).
4. Press Ctrl-Tab

Result: switches to file B

Expected; switches to the Untitled document, since that's the last document you were in.
(It looks like the Untitled doc is added to the _end_ of the MRU list instead of the start).

This is a regression from the split view work, but I'm not sure if it's that important to fix for 1.0.  That said, I've found it  relatively easy to hit since I often want to switch back to a previous file to cut/copy text from it, then switch back to the new Untitled file & paste -- and I you'll this every time with that workflow.

JeffryBooher:
@ingorichter please review

JeffryBooher:
FBNC @peterflynn

PR #9661, opened 2014-10-23
Fix makeFileMostRecent API

JeffryBooher:
Existing File Objects are opened using `CMD_OPEN_AND_ADD_TO_WORKING_SET` which opens the file then adds it to the working set. The act of adding it to the working set would move it to the front of the MRU list if it was the currently viewed file (which happens in step 1 of the `OPEN_AND_ADD_TO_WORKING_SET` command)

We create an in-memory document to edit an untitled document and call `MainViewManger._edit()` on the document object. This adds the file to the working set _before_ it has been opened which, because it wasn't the current file, means that its entry got added to the end of the MRU list.

We then call `_makeFileMostRecent()` which should have made the untitled document the most recent but there was were 2 flaws:

1) It was originally intended to only reorder the working set for items added to the working set (more on this later)
2) It was getting `ACTIVE_PANE` for untitled documents which needed to be resolved to an actual `paneId`.  This is why untitled documents were MRU ordered last.  

Because `_makeFileMostRecent` was not intended to work for Temporary Files but it probably should have been otherwise you could open a temporary view in the left pane and another temporary view in the right pane and never tab between them.  You would have been scurried off to another document somewhere which would close the temporary view.  Therefore, `_makeFileMostRecent` now works on temporary views and a new event is added `Pane.viewDestroy` to trap when temporary views are destroyed so they can be removed from the MRU list.
#9654

@peterflynn


JeffryBooher:
closing in lieu of #10327

Issue #9345, opened 2014-09-27
Edge Inspect extension still uses old preferences system


le717:
```
getPreferenceStorage is called with client ID 'C:/Users/le717/AppData/Roaming/Brackets/extensions/user/edge-inspect-extension/inspect.js,' use PreferencesManager.definePreference instead.
    at Object.getPreferenceStorage (file:///C:/Program%20Files%20(x86)/Brackets/dev/src/preferences/PreferencesManager.js:127:32)
    at Object.<anonymous> (file:///C:/Users/le717/AppData/Roaming/Brackets/extensions/user/edge-inspect-extension/inspect.js:62:36)
    at Object.context.execCb (file:///C:/Program%20Files%20(x86) DeprecationWarning.js:85
getValue is called to get preference 'hasShownGettingStarted,' use PreferencesManager.get instead.
    at PreferenceStorage.getValue (file:///C:/Program%20Files%20(x86)/Brackets/dev/src/preferences/PreferenceStorage.js:128:28)
    at Object.<anonymous> (file:///C:/Users/le717/AppData/Roaming/Brackets/extensions/user/edge-inspect-extension/inspect.js:63:26)
    at Object.context.execCb (file:///C:/Program%20Files%20(x86) 
```

I was not aware the extension is open source, so I reported it here.

Also, whenever this is fixed, if the hidden .git folder and `.DS_Store` files  can be removed from the archive, it'd download a lot quicker. :smile:

ficristo:
@madanbn @swmitra can we close this?

swmitra:
Yes @ficristo. We can close this issue as there are work items as part of Brackets core.

PR #9752, opened 2014-10-31
Remove old preferences system

le717:
For #8751.

This also removes `PreferencesManager.setValueAndSave()`, deprecated by 08e8208e2c2bd229be4e0f11700853890a981c5f in Sprint 37. In order to prevent a repeat of Quick Open, I removed it in a separate commit in case it needs to be reverted and set my search utility to scan more thoroughly. No known extension usage was discovered.

/cc @dangoor


busykai:
I guess it's pretty safe to remove the old preference implementation at this point. This PR is not mergeable, but before @le717 actualizes it, we need to decide.

CC @redmunds, @peterflynn

le717:
@busykai A problem is `convertPreferences()`, removed in this PR, was never marked as deprecated despite the underlying system being so, and it possibly should have been emitting a warning all this time, as discussed in #10412. Depending on that outcome, this PR may not land until as late as 1.4 (meaning a warning should be added and doesn't get in until 1.3) or could go straight in right now.

That said, a number of extensions still use the old system or `convertPreferences()` (or in the case of #9345, the changes were never pushed to the registry) (see https://github.com/adobe/brackets/issues/8751#issuecomment-61132680). But as @MarcelGerber said somewhere, "People tend to ignore the issues posted on their extension repos", so I'm not sure if it is worth waiting any longer.

le717:
Also, as we are late into v1.2, it may be risky to merge it now for fear of suddenly breaking extensions right before release. If it shouldn't be merged now, I'd go for it just as soon as the version is changed to 1.3 (with respect to a decision on adding a warning).

ficristo:
Closing in favour of https://github.com/adobe/brackets/pull/12720 since is more up-to-date.

PR #9784, opened 2014-11-03
Ability to specify that SSL errors should be ignored ...

zaggino:
... when downloading extensions, useful when behind a corporate proxy/firewall that is doing SSL decryption as part of the company security policy. Turned on by adding the following line to the preferences file:

```
"proxy.ignoreSSL": true
```


busykai:
@zaggino, @ingorichter I'd say we should go for supporting the entire set of `agentOptions`. except for the `Buffer`s should be replaced with corresponding the file names.

PR #9971, opened 2014-11-19
Issue  #9934 - error on zip file opening

hzahoori:
Hi @JeffryBooher  Please merge, my update to the bug, Now when you drag a file from a zip file it will return an error. 
I would love to get your advice on best practices for next bug.Thanks


nethip:
@peterflynn @hzahoori I had a quick look at `ClientHandler::OnDragEnter` and there is no code to check for supporting the right file type. I think the change should be do check for all the supported formats.

PR #10108, opened 2014-12-08
Moved padding-bottom to container so lists can't be horizontally scrolled

stowball:
If the sidebar is narrower than the list of open files, the user could click and drag right on the padding-bottom beneath the working set to scroll the file names over. The user could not click and drag back though.

This fix puts the padding on the container instead, so the overflow has no effect.

### DIFF
--- a/src/styles/brackets.less
+++ b/src/styles/brackets.less
@@ -706,7 +706,7 @@ a, img {
 
 #working-set-list-container {
     background: @bc-sidebar-bg;
-    padding-bottom: 21px; // Adds working set bottom padding to the last UL.
+    padding-bottom: 21px; // Adds bottom padding to the working set.
 }
 
 .working-set-header {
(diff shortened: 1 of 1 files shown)


redmunds:
@stowball I can't seem to reproduce the problem that this is fixing. Are you still seeing it? Can you add a screenshot or video?

Got any other details? Does it also require a vertical scrollbar? Which OS/version? Latest version of Brackets?

stowball:
I'm having trouble recording a screencast of it, but I've just noticed that my fix breaks the `.scroller-shadow bottom` position as that's absolutely positioned

stowball:
Managed to screencast the issue where you can drag the padding and scroll the file list out of view brackets-narrow-files-list-scroll-bug

abose:
seems useful to scroll ;) Reminds me of this.
image

stowball:
Do you follow me on Twitter?! https://twitter.com/stowball/status/652247896432074752?s=09

abose:
Coincidence!

Issue #9933, opened 2014-11-14
No substitution, author term when the language changes.


datamweb:
Hello.
I am developing a extensions for Brackets And Farsi are also supported.
In this case, writing packegs file:

```
{
    "name": "p_parsa_dadashi.talesh",
    "title": "test",
    "description": "test description ",
    "homepage": "https://github.com/datamweb",
    "author": "pooya parsa dadashi",
    "version": "0.0.1",
    "license": "MIT",
    "engines": {
        "brackets": ">=0.24.0"
    },
    "i18n": [ "en","fa-ir"],
    "package-i18n": {
        "fa-ir": {
                "description": "توضیحات متن تست داتام وب",
                "title": "داتام وب افزونه",
                "author": "پویا پارسا داداشی"


        }
    }
}
```

.
The author will not be translated word.
Please be overcome.

PlanetVaster:
So should the PR be pulled by now?

PR #10128, opened 2014-12-09
Now is possible to translate the extension author's name

nicolo-ribaudo:
Now is possible to translate the extension author.

Can also be translated contributors, even if they aren't yet shown in the ui.

(issue #9933, @datamweb)


ingorichter:
@nicolo-ribaudo Thanks for the contribution. Please add a couple of testcases for the new functionality. This way we can test some edge cases and make sure we find out if something breaks.

src/extensibility/node/package-validator.js
```diff
@@ -232,6 +232,29 @@ function validatePackageJSON(path, packageJSON, options, callback) {
                 }
             }
             
+            // normalize translated author and contributors
+            if (metadata["package-i18n"]) {
+                Object.keys(metadata["package-i18n"]).forEach(function (lang) {
+                    var packageLang = metadata["package-i18n"][lang];
+
+                    if (packageLang.author) {
```
peterflynn: This block (down to line 254) is duplicated with the code above, so please factor it into a reusable function.

src/extensibility/ExtensionManagerView.js
```diff
@@ -250,6 +250,12 @@ define(function (require, exports, module) {
                             info.metadata[prop] = info.metadata["package-i18n"][locale][prop];
                         }
                     });
+                    if (info.metadata["package-i18n"][locale].hasOwnProperty("author")) {
+                        info.metadata.author = info.metadata["package-i18n"][locale].author;
+                    }
+                    if (info.metadata["package-i18n"][locale].hasOwnProperty("contributors")) {
+                        info.metadata.contributors = info.metadata["package-i18n"][locale].contributors;
+                    }
```
peterflynn: I'm not sure I understand why these need special treatment, instead of just adding `"author"` and `"contributors"` to the array above.

peterflynn: Btw, we may want to add a comment here explaining that `"author"` is safe for localization because it's just a "friendly name" and the GitHub user id is stored in a separate `"owner"` field (which will never be localized).

nicolo-ribaudo:
@peterflynn I've cleaned up the code.

@ingorichter Are there already some translation test that I can extend?

ficristo:
@nicolo-ribaudo can you solve the merge conflicts?
I think this still miss the comment requested from @peterflynn

nicolo-ribaudo:
I'm sorry but I don't use Adobe Brackets anymore. If anyone is interested in cherry-picking my commits and make the needed changes, feel free to do that.

PR #10141, opened 2014-12-10
#10071 bash detection for extensionless file

gklo:
Hi, this is probably not the best solution. Please review my code and let me know :)

### DIFF
--- a/src/document/Document.js
+++ b/src/document/Document.js
@@ -680,7 +680,10 @@ define(function (require, exports, module) {
      */
     Document.prototype._updateLanguage = function (rawText) {
         var oldLanguage = this.language;
-        this.language = LanguageManager.getLanguageForPath(this.file.fullPath, null, rawText);
+        this.language = LanguageManager.getLanguageForPath(this.file.fullPath);
+        if (this.language === LanguageManager.getLanguage("unknown")) {
+            this.language = LanguageManager.getLanguageForContent(rawText);
+        }
         if (oldLanguage && oldLanguage !== this.language) {
             this.trigger("languageChanged", oldLanguage, this.language);
         }
--- a/src/document/Document.js
+++ b/src/document/Document.js
@@ -77,7 +77,7 @@ define(function (require, exports, module) {
      */
     function Document(file, initialTimestamp, rawText) {
         this.file = file;
-        this._updateLanguage();
+        this._updateLanguage(rawText);
         this.refreshText(rawText, initialTimestamp, true);
     }
     
@@ -675,10 +675,12 @@ define(function (require, exports, module) {
     
     /**
      * Updates the language to match the current mapping given by LanguageManager
+     * 
+     * @param {!string} rawText  Text content of the file.
      */
-    Document.prototype._updateLanguage = function () {
+    Document.prototype._updateLanguage = function (rawText) {
         var oldLanguage = this.language;
-        this.language = LanguageManager.getLanguageForPath(this.file.fullPath);
+        this.language = LanguageManager.getLanguageForPath(this.file.fullPath, null, rawText);
         if (oldLanguage && oldLanguage !== this.language) {
             this.trigger("languageChanged", oldLanguage, this.language);
         }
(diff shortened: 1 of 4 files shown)


gklo:
Since entensionless files are likely to be script, so I re-implement it as general shebang detection instead of bash specifed. It might have better performance as well because no need to go through each language object to test regexp.  Thanks to @apla, I have a clear idea of using regexp now.

src/document/Document.js
```diff
@@ -675,10 +675,12 @@ define(function (require, exports, module) {
     
     /**
      * Updates the language to match the current mapping given by LanguageManager
+     * 
+     * @param {!string} rawText  Text content of the file.
      */
-    Document.prototype._updateLanguage = function () {
+    Document.prototype._updateLanguage = function (rawText) {
```
peterflynn: This is called from other places that won't pass `rawText` (e.g. `_notifyFilePathChanged()`, `DocumentManager._handleLanguageAdded()`, etc.).  Will the behavior be correct then?  It safer & simpler to remove the arg and just invoke `this.getText()` instead.

peterflynn: But also... since this is called in multiple cases, various actions will trigger a "re-check" of the document's text.  If the user edits the document to add/remove the shebang, at what point would they expect the document to switch to/from Bash mode?  We should make sure that feels predictable.

src/language/LanguageManager.js
```diff
@@ -243,6 +243,28 @@ define(function (require, exports, module) {
...
+
+    /**
+     * Resolves a file content to a Language object.
+     * @param {!string} rawText  Text content of the file.
+     *
+     * @return {Language} The language detected based on file content or the fallback language
+     */
+    function getLanguageForContent(rawText) {
+        // Shebang detection for script languages 
+        if (_languages && rawText) {
```
peterflynn: I don't think `_languages` can ever be null

src/language/LanguageManager.js
```diff
@@ -263,7 +285,7 @@ define(function (require, exports, module) {
      *
      * @return {Language} The language for the provided file type or the fallback language
      */
-    function getLanguageForPath(path, ignoreOverride) {
+    function getLanguageForPath(path, ignoreOverride, rawText) {
```
peterflynn: Are we changing the API to _require_ passing the text content?  That's a breaking API change, which has to be done carefully (and there are many callers in the Brackets core code alone that will need to be updated).  OTOH, if it's optional we're inviting inconsistent behavior: some features in Bracekts will pass only the path (treating Bash files as plain text) while others will pass the full text and treat the file as Bash.  Is there any reasonable, performant way around that?

Might be worth thinking through use cases for `getLanguageForPath()` (and/or surveying current callers) to think through what kinds of problems might arise when different parts of Brackets treat the same file as different types.

src/document/Document.js
gklo: I believe `this.getText()` won't exist cause the functions are called by `Document` contructor.  Therefore,  using `rawText` is probably the best way to pass the contents to `LanguageManager`.  So far, the shebang detection happens only when user opens a file, I am not sure when it triggers during editing.  Could you or anybody help me for that?

src/language/LanguageManager.js
gklo: `_languages` seemed to be null once for some reason when I was debugging.  I will try testing it again and see.

src/document/Document.js
gklo: ... and yes rawText can be null, so `_updateLanguage()` can be called without `rawText`

src/language/LanguageManager.js
gklo: yea that's true ... my bad.  It is not really appropriate to pass content through `getLanguageForPath()`.  I think `getLanguageForContent()` should be kept for contents and `getLanguageForPath()` for path only, just like other `getLanguageFor*()` methods.  Do only one thing in a method.

src/language/LanguageManager.js
gklo: you are right, thanks for telling me :)

nextgenthemes:
I really would appreciate this or a extension, I am tired of manually switching to bash. Well I will probably rename all my scripts to .sh now even if I not really like it.

PR #10224, opened 2014-12-19
Restore the original key binding

### LINKED ISSUE #10157: Deleting a key binding to null in the user key map does not restore the original key binding
1. Open user key map.
2. Add a key binding to null (e.g. `"Cmd-O": null` on mac).
3. Save the file and check the File menu. File > Open does not have any shortcut.
4. Delete the key binding added in step 2.
4. Save the file and check the File menu again.

Result: Cmd-O shortcut is not restored to File > Open.

Workaround: Quit and Restart or Reload.

RaymondLim:
This fixes #10157.


redmunds:
Merging.

PR #10275, opened 2014-12-24
Be more specific about filesystem watch error

ingorichter:
We shouldn't hide the fact that the user hit an unsupported use case. Let's make it crystal clear and hope that some _phantom_ issues with file watching could be resolved this way.


nethip:
Thanks @ingorichter

PR #10285, opened 2014-12-27
Live Preview MultiBrowser improvements (#10206, #10239)

### LINKED ISSUE #10206: LiveDevelopmentMultiBrowser unit test issues
**1)** <del>The Jasmine test-runner window always logs an exception when it is first loaded, before any tests are run:</del> [console messages hidden]

<del>(because bootstrap-twipsy-mod is not loaded in the Jasmine window)</del> Fixed in #10285

**2)** Several LiveDevelopmentMultiBrowser tests fail for me with "timeout: timed out after 5000 msec waiting for livedevelopment.done.opened" (usually 2-3 tests fail per run).

**3)** The console is littered with error messages after running this test suite:

```
 [testWindow]  [node-error 10:14:11 AM] nodeSocketTransport: Socket closed, but couldn't locate client SpecRunnerUtils.js:543
runs.forEach._testWindow.console.(anonymous function) SpecRunnerUtils.js:543
handleLogEvent src/extensions/default/DebugCommands/NodeDebugUtils.js:114
trigger EventDispatcher.js:222
triggerWithArray EventDispatcher.js:260
NodeConnection._receive NodeConnection.js:479
```
(twice)

```
[testWindow]  Exception in 'base:log' listener on NodeConnection {domains: Object, _registeredModules: Array[0], _pendingInterfaceRefreshDeferreds: Array[0], _pendingCommandDeferreds: Array[0], _autoReconnect: true…}
TypeError: Cannot read property 'console' of null
    at Console.runs.forEach._testWindow.console.(anonymous function) [as log] (file:///C:/code/Brackets/brackets-app/brackets/test/spec/SpecRunnerUtils.js:544:53)
    at handleLogEvent (file:///C:/code/Brackets/brackets-app/brackets/src/extensions/default/DebugCommands/NodeDebugUtils.js:117:21)
    at NodeConnection.trigger (file:///C:/code/Brackets/brackets-app/brackets/src/utils/EventDispatcher.js:222:40)
    at Object.triggerWithArray (file:///C:/code/Brackets/brackets-app/brackets/src/utils/EventDispatcher.js:260:28)
    at NodeConnection._receive (file:///C:/code/Brackets/brackets-app/brackets/src/utils/NodeConnection.js:479:29)
```
(13 times - various listeners all with the same TypeError)

Even if these are caused by the timeouts above, it suggests the code does not fail gracefully in the event of a connection timeout.

**4)** One time this test suite seemed to get stuck with the Node process hanging onto file handles in the brackets/test/temp folder, which caused many other unit tests to fail since they were unable to clear that folder.  Restating the Node process released the file handles.  Depending on how easy it is to hit this failure case, we may want to do more to increase robustness...

### LINKED ISSUE #10239: Provide a toggle menu item to switch between Live Preview implementations
This is a follow-up to #10226. Manual editing of the Brackets configuration file may be cumbersome and error-prone. A much easier way to switch between the implementations is to provide a toggle menu item.

busykai:
Fix #10206 (point 1):

PreferencesManager.on("change") was executing before AppInit.appReady().
This was causing various inconsistencies, including setting up the UI before
LiveDevelopment is actually initialized. Handling of pref change is now moved
to AppInit.appReady and the implementation is set only after the modules are
actually initialized.

Fix #10239:

Toggle menu item to enable/disable multi-browser live preview. Under `File -> Enable Experimental Live Preview`. Multiple fixes to implementation life-cycle to support correct state transitions on switch.

Minor nits included: pref definition improved, suite name made shorter, spare lines removed, documentation improvements.

CC: @peterflynn, @sebaslv, @redmunds

### DIFF
--- a/src/LiveDevelopment/LiveDevMultiBrowser.js
+++ b/src/LiveDevelopment/LiveDevMultiBrowser.js
@@ -467,10 +467,13 @@ define(function (require, exports, module) {
     }
 
     /**
-     * Close all active connections
+     * Closes all active connections.
+     * Returns a resolved promise for API compatibility.
+     * @return {$.Promise} A resolved promise
      */
     function close() {
-        return _close(true);
+        _close(true);
+        return new $.Deferred().resolve().promise();
     }
     
     /**
--- a/src/LiveDevelopment/LiveDevelopment.js
+++ b/src/LiveDevelopment/LiveDevelopment.js
@@ -52,6 +52,7 @@
  *      2: Loading agents
  *      3: Active
  *      4: Out of sync
+ *      5: Sync error
  *
  * The reason codes are:
  * - null (Unknown reason)
--- a/src/LiveDevelopment/main.js
+++ b/src/LiveDevelopment/main.js
@@ -40,6 +40,7 @@ define(function main(require, exports, module) {
         Commands            = require("command/Commands"),
         AppInit             = require("utils/AppInit"),
         LiveDevelopment     = require("LiveDevelopment/LiveDevelopment"),
+        MultiBrowserLiveDev = require("LiveDevelopment/LiveDevMultiBrowser"),
         Inspector           = require("LiveDevelopment/Inspector/Inspector"),
         CommandManager      = require("command/CommandManager"),
         PreferencesManager  = require("preferences/PreferencesManager"),
@@ -50,9 +51,6 @@ define(function main(require, exports, module) {
         ExtensionUtils      = require("utils/ExtensionUtils"),
         StringUtils         = require("utils/StringUtils");
     
-    // expermiental multi-browser implementation
-    var MultiBrowserLiveDev = require("LiveDevelopment/LiveDevMultiBrowser");
-
     var params = new UrlParams();
     var config = {
         experimental: false, // enable experimental features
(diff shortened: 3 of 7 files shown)


busykai:
@redmunds, this is ready for a review.

redmunds:
@busykai I found a weird case when testing this. Let me know if I should file this:

1. Start Bracket s on a system with a local server setup and running
2. Switch to a site with files on the local server
3. I originally hit this by setting base url to be `http:/localhost/`, but this is *not* required
4. Switch to LiveDev MultiBrowser
5. Start Live Dev

Results
Live Dev never completely connects. It stays in Connecting state.

I don't think it's related to this change. Let me know.

busykai:
@redmunds, thanks for the use case! we will take a look.

CC @sebaslv

sebaslv:
@redmunds, I'm trying to reproduce this case but still couldn't do it. This is what I'm doing: 

1. Start a local server which serves let say index.html file
2. Start Brackets
3. Open the folder that contains the index.html being served
4. Open http://localhost/index.html in a browser
5. File -> Enable Experimental Live Preview
6. Start Live Dev

By following these steps I got LiveDev working. Please, let me know if those are representative of the steps you have described or if I misunderstood something. Thanks.

redmunds:
@sebaslv I did not open in browser first (step 4), so not sure if that's making a difference.

FYI, I'm seeing this on Win7 Firefox (latest) with WampServer 2.2 .

sebaslv:
@redmunds, Thanks for the info. I was able to reproduce it now by configuring local server, the step I missed before I think :). 

It seems that it never connects beacuse the page is not instrumented so the ```connect``` message never comes back from the browser after launching it. It looks like in this scenario, ```LiveDevMultiBrowser``` is getting an instance of ```UserServer``` since it has been  registered by ```LiveDevelopment``` and both implementations share ```LiveDevServerManager```.

Currently we're not managing that error. I'll create a fix for that independently of this case since it's pending. The thing is that it would close the connection which is something different to what the current implementation does in this scenario. I see that it launches the page and moves the icon to a connected status but keep live editing disabled. I could figure out how to identify the case and handle that in this particular way but not sure if that is the right UX we would like for this.  

What do you guys suggest?

redmunds:
@sebaslv I agree that it should be a separate issue and not hold up this PR. I'll open new issue.

Seems like doing a better job of cleaning up when changing implementations should fix that.

redmunds:
@sebaslv I filed issue #10374 . I'm not sure I understand all of the subtle points, so feel free to update it.

sebaslv:
Sounds good. I'll create a PR for that isuue. Thanks.

src/extensions/default/DebugCommands/main.js
```diff
@@ -230,9 +230,9 @@ define(function (require, exports, module) {
         });
     }
     
-    function toggleErrorNotification(bool) {
+    function togglePref(key, bool) {
         var val,
```
redmunds: These changes to DebugCommands/main.js are not necessary. I only refactored this code from `toggleErrorNotification()` into `togglePref()` to reduce duplication with new function, but it was moved to `LiveDevelopment/main.js` so there's no need for any changes here.

redmunds:
@busykai Done with review. This looks good. Just need to backout changes to DebugCommands/main.js and squash commits.

busykai:
@redmunds, done!

redmunds:
Merging.

PR #10324, opened 2015-01-07
Add menu item for LiveDev MultiBrowser mode

redmunds:
This is for #10239 

cc @busykai

### DIFF
--- a/src/extensions/default/DebugCommands/main.js
+++ b/src/extensions/default/DebugCommands/main.js
@@ -69,7 +69,10 @@ define(function (require, exports, module) {
         DEBUG_LOG_NODE_STATE            = "debug.logNodeState",
         DEBUG_RESTART_NODE              = "debug.restartNode",
         DEBUG_SHOW_ERRORS_IN_STATUS_BAR = "debug.showErrorsInStatusBar",
+        DEBUG_TOGGLE_LIVEDEV_MB_MODE    = "debug.toggleLiveDevMultiBrowser",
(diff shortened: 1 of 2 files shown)


busykai:
@redmunds, thanks for taking this on. I was waiting for #10285 to get attention since it defines the same preference. Anyway, I'll review and land this and then re-base the other one (quite important, by the way).

busykai:
@redmunds, I think the pref definition should be in `LiveDevelopment/main.js` as well as its handling. Is there any specific reason why we would want it in `Debug` instead of `File` (next to 'Live Preview')? I understand this is experimental still, but the pref and its command handling really belongs to the subsystem itself.

sbruchmann:
We could label the menu item as ‘Enable Live Preview MultiBrowser (Experimental)’.

redmunds:
@busykai I thought this would just be a temporary solution, so that's why I hard-coded the pref id and put it in the Debug menu.

busykai:
@redmunds, i sincerely hoped to land #10285 for 1.2. what if i merge this branch into that PR and re-work. we still don't know for how long both implementations would hang around together (probably for quite a while), so having the option exposed next to launch command in `File` makes more sense to me. what do you think?

redmunds:
@busykai 

> what if i merge this branch into that PR

Sounds good. I'll just close this one.

redmunds:
Closing.

@busykai Let me know when you're done with the **randy/issue-10239** branch so I can delete it.

busykai:
@redmunds, merged and deleted the branch, but #10285 is not yet ready.

PR #10327, opened 2015-01-07
Fix makeFileMostRecent API

### LINKED ISSUE #9654: Untitled documents aren't added to MRU until visited a 2nd time

JeffryBooher:
Fixes #9654 

This is a squashed version of #9661

Existing File Objects are opened using CMD_OPEN_AND_ADD_TO_WORKING_SET which opens the file then adds it to the working set. The act of adding it to the working set would move it to the front of the MRU list if it was the currently viewed file (which happens in step 1 of the OPEN_AND_ADD_TO_WORKING_SET command)

We create an in-memory document to edit an untitled document and call MainViewManger._edit() on the document object. This adds the file to the working set before it has been opened which, because it wasn't the current file, means that its entry got added to the end of the MRU list.

We then call _makeFileMostRecent() which should have made the untitled document the most recent but there was were 2 flaws:

1) It was originally intended to only reorder the working set for items added to the working set (more on this later)
2) It was getting ACTIVE_PANE for untitled documents which needed to be resolved to an actual paneId. This is why untitled documents were MRU ordered last.

Because _makeFileMostRecent was not intended to work for Temporary Files but it probably should have been otherwise you could open a temporary view in the left pane and another temporary view in the right pane and never tab between them. You would have been scurried off to another document somewhere which would close the temporary view. Therefore, _makeFileMostRecent now works on temporary views and a new event is added Pane.viewDestroy to trap when temporary views are destroyed so they can be removed from the MRU list.

### DIFF
--- a/src/view/MainViewManager.js
+++ b/src/view/MainViewManager.js
@@ -253,6 +253,19 @@ define(function (require, exports, module) {
     function _makeMRUListEntry(file, paneId) {
         return {file: file, paneId: paneId};
     }
+
+    /**
+     * Locates the first  MRU entry of a file
+     * @param {!File} File - the file
+     * @return {{file:File, paneId:string}}
+     * @private
+     */
+    function _findFileInMRUList(file) {
+        return _.findIndex(_mruList, function (record) {
+            return (record.file.fullPath === file.fullPath);
+        });
+    }
+
     
     /**
      * Retrieves the currently active Pane Id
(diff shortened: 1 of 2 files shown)


JeffryBooher:
@ingorichter ready for you to retest, review and merge.  Thanks!

ingorichter:
Looks much better now. I'm going to use it for today and merge it by the end of the day.

JeffryBooher:
@ingorichter thanks for the review!

Issue #10362, opened 2015-01-12
Quickview min image size too large


EJanuszewski:
Only a small change, seems there's no need for a min-width on the container either.

I have tested with various image sizes and all seem to behave as expected, I will send the pull request shortly.

prafulVaishnav:
@EJanuszewski @peterflynn I was trying to know the pros and cons for keeping min-width for preview. I think, minimum width of 100px will be useful for preview of small image size. If its large, we can reduce the min width size.

EJanuszewski:
@prafulVaishnav @peterflynn Shall I hold back on both pull requests then until a final decision?

prafulVaishnav:
@EJanuszewski Fixes to issue  #5960 and #10180 which is present in Brackets.less are considerably good and can be merged. So, as discussed you could create Pull Request for same.

EJanuszewski:
@prafulVaishnav I'm not too sure how to do a pull request for just one file, could you give me some assistance?

redmunds:
@EJanuszewski You should be creating a separate branch in your local fork of Brackets, pushing that branch to Brackets repo, and then creating a pull request for your branch. It's the same process whether it's 1 file or many. Take a look at the [How to Hack on Brackets](https://github.com/adobe/brackets/wiki/How-to-Hack-on-Brackets) article, especially the [Submitting a Pull Request section](https://github.com/adobe/brackets/wiki/How-to-Hack-on-Brackets#submitting-a-pull-request).

baherramzy:
Friendly ping to close this issue if it has been fixed.

Issue #10180, opened 2014-12-15
[CEF 2171] Image views don't scale when the height of the view changes


jasonsanjose:
1. Open the find in files panel
2. Resize it tall enough to cause scrolling when an image is viewed
3. Select an image from the project tree

Actual: Image is clipped on the bottom edge
Expected: Scale to fit or provide scroll bars

![image](https://cloud.githubusercontent.com/assets/1148713/5444415/31484716-845f-11e4-9162-7fb943a1665d.png)

peterflynn:
Note: I updated the repro Gist to an even 500px height for simplicity.

Some related info:
http://stackoverflow.com/questions/1948249/how-to-set-maximum-height-for-img-but-keep-proportions
http://stackoverflow.com/questions/17329040/img-max-height-100-causes-img-to-exceed-bounds

EJanuszewski:
I believe whilst fixing the meta data I may have also fixed this, see 356/357 https://github.com/EJanuszewski/brackets/commit/30cfd3833df69b6df1ad64836eebf9dad3e207b6#diff-d170b001614b4677146b7f46339e2d88R356

EJanuszewski:
@prksingh See my comment above, from what I can tell it seems to be working fine after those changes.

EJanuszewski:
Is there something wrong with the fix I did? Just curious why it's not been merged as it's tagged for the 1.2 release :)

prafulVaishnav:
@EJanuszewski Can you create the pull request for this fix? I have added a note at  EJanuszewski@30cfd38#diff-d170b001614b4677146b7f46339e2d88R356.

EJanuszewski:
@prafulVaishnav Already did just here https://github.com/adobe/brackets/pull/10364, let me know if there's something wrong with that pull and i'll make the changes

prafulVaishnav:
@EJanuszewski This pull contain fixes for different issues. Can you please create different pull request for different files? Brackets.less has fix for issue #5960 and #10180. QuickView.less has fix for issue #10362..

EJanuszewski:
@prafulVaishnav Sure, I didn't know different files needed different pull requests especially when they are related.

To clarify I need to make another pull request for https://github.com/adobe/brackets/issues/10362 and somehow remove https://github.com/EJanuszewski/brackets/commit/45a74c50c1de5300a0a04dfff8010565650ca26d from the current pull request?

prafulVaishnav:
[...]

PR #10364, opened 2015-01-12
Remove min width on image quickview

EJanuszewski:
Only a small change, seems there's no need for a min-width on the container either.

I have tested with various image sizes and all seem to behave as expected, I will send the pull request shortly.

### DIFF
--- a/src/styles/brackets.less
+++ b/src/styles/brackets.less
@@ -353,9 +353,8 @@ a, img {
             .image-centering {
                 display: inline-block;
                 vertical-align: middle;
-                margin-top: -50px; /* Offset as vertical align takes image metadata into account */
-                max-width: 90%;
-                max-height: 90%;
+                width: 90%; //As long as the width isn't set on the image itself theres no need for max-width/height
+                height: 90%;
             }
             .image-data {
                 font-weight: @font-weight-semibold;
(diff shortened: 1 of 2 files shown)


JeffryBooher:
@EJanuszewski you can tag the issues that a pull request addresses by adding #(issue-number), #(issue-number), etc....  That links the issues to the pull request so that it's easy to follow.

#10180 
#5960

EJanuszewski:
@JeffryBooher Ok thanks, hopefully any newer things I do i'll get it right first time, I'm new to referring pulls and issues onto someone elses project.

JeffryBooher:
@EJanuszewski no worries :)

src/styles/brackets.less
```diff
@@ -404,6 +403,7 @@ a, img {
 
         .image {
             position: relative;
+            height: calc(~'100% - 53px');
```
peterflynn: Where does the hardcoded value 53px come from?

peterflynn:
Are the brackets.less changes intended to be part of this PR?  The description at top only mentions #10362...

EJanuszewski:
Sorry I should have put a comment in, 53 is the height of the meta data + its margin(38+15).

Fixes for both are just in the brackets.less, it just happened that to fix the chopping off issue also seemed to fix the resizing, hopefully that helps, I'm on my phone so can't type loads.

PR #10369, opened 2015-01-13
Allow non-autoClose dialogs to trigger "buttonClick" more than once

peterflynn:
Allow non-`autoClose` dialogs to trigger "buttonClick" more than once (until the dialog is actually closed). This lets clients use `autoClose=false` to do simple validation.

Presumably the `.one()` was originally intended to prevent double-clicking a button (etc.) from triggering `"buttonClick"` twice, since the dialog remains semi-visible for a beat while animating closed after the first click. But
the keydown handler had no such protections. Now anyone calling `_processButton()` is protected from calls while the dialog is already closing.

Also fixes a JSLint warning introduced back in 34b1a186.

### DIFF
--- a/src/widgets/Dialogs.js
+++ b/src/widgets/Dialogs.js
@@ -71,7 +71,7 @@ define(function (require, exports, module) {
      */
     function _dismissDialog($dlg, buttonId) {
         $dlg.data("buttonId", buttonId);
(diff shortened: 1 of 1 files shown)


peterflynn:
@TomMalbran are you interested in reviewing, since you've been in the Dialogs code a bunch?

PR #10372, opened 2015-01-14
update extension manager deps

ingorichter:
This PR will update all the node dependencies of the extension manager. This will hopefully resolve some of the installation issues that we have seen in the past.
All the unit test passed successfully.


ingorichter:
@MarcelGerber once this got merged, we should update the registry to use a more recent version of the package-validator.

lucasecardoso:
This is not a bug, so I guess it should be closed?

ficristo:
This is a PR, and still valid.
Maybe we can rethink how it is done or update it, but the fact the dependencies should be update is still true.

ficristo:
According to the conversation in https://github.com/adobe/brackets/issues/12898 this seems important.

zaggino:
Closed this in favor of https://github.com/adobe/brackets/pull/12933

PR #10389, opened 2015-01-15
Highlight matched hints

sprintr:
Highlights matched hints in CSS

### DIFF
--- a/src/extensions/default/CSSCodeHints/styles/brackets-css-hints.css
+++ b/src/extensions/default/CSSCodeHints/styles/brackets-css-hints.css
@@ -24,3 +24,7 @@
 .brackets-css-hints .matched-hint {
     font-weight: 500;
 }
+
+.dark .brackets-css-hints .matched-hint {
+    color: #ccc;
+}


src/extensions/default/CSSCodeHints/styles/brackets-css-hints.css
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014 Adobe Systems Incorporated. All rights reserved.
+ * Copyright (c) 2015 Adobe Systems Incorporated. All rights reserved.
```
redmunds: We never update the copyright date on files. This is the date the file was added to code base.

sprintr:
@redmunds Done!

redmunds:
Good catch. It looks like same change needs to be made to **/src/extensions/default/JavaScriptCodeHints/styles/brackets-js-hints.css**.

sprintr:
JavaScript hints uses more than one rules for hints, depending on priority. It will be a nice idea to have @larz0's opinion on the choice of colors.

redmunds:
@sprintr 

> JavaScript hints uses more than one rules for hints, depending on priority.

Good point. I'll just merge this fix for now.

Issue #9880, opened 2014-11-08
Brackets doesn't detect corrupt state.json file


SirCheeseyMelon:
The program works fine an all, but every time I open it, it opens with the "getting started" folder

peterflynn:
We should consider showing an error if we're unable to write to it also, since that seems to have cropped up as a problem on Linux for several users

redmunds:
Updated title. Bracket should detect when state.json cannot be read or written and try to delete it so a new file can be created (i.e. start from scratch). If it cannot be deleted, then an informational message should be displayed explaining that file needs to be manually deleted.

nethip:
Moving this to 1.3

rroshan1:
Issue fixed with PRs #10390 and #10951.

Issue #9520, opened 2014-10-10
No error recovery or logging for invalid state.json files


sarciszewski:
On Windows 8.1...

1. Install Brackets
2. Open a folder
3. Install an extension (e.g. brackets-git)
4. Restart to install the extension

You end up back at the "Getting Started" folder. The extension is still installed, but on every startup it displays the "we just installed" prompt.

sarciszewski:
I've recreated the issue. Apparently it has to do with Windows 8 file corruption when a sudden power outage occurs. I would recommend notifying users when state.json is corrupt and giving them the option to reset it within Brackets.

redmunds:
@sarciszewski This sounds like the same underlying issue as #9880. I submitted pull request #10390 yesterday as a proposed fix. Marking this as Fix in Progress.

rroshan1:
Issue fixed with PRs #10390 and #10951.

sarciszewski:
:+1: Awesome, thanks ;)

PR #10390, opened 2015-01-15
Handle unreadable and unparsable state.JSON pref file #9880

redmunds:
This is for #9880 . Also #9520.

Delete unreadable pref file so it gets recreated. Reset contents of unparsable pref file.

@peterflynn This is a first stab at handling bad pref files. I put some questions in the diff.

### DIFF
--- a/src/preferences/PreferencesBase.js
+++ b/src/preferences/PreferencesBase.js
@@ -148,10 +148,12 @@ define(function (require, exports, module) {
      * @param {boolean} createIfNew True if the file should be created if it doesn't exist.
      *                              If this is not true, an exception will be thrown if the
      *                              file does not exist.
+     * @param {boolean} isStateJson True if the specified preferences file is state.json 
      */
-    function FileStorage(path, createIfNew) {
+    function FileStorage(path, createIfNew, isStateJson) {
         this.path = path;
         this.createIfNew = createIfNew;
+        this.isStateJson = isStateJson;
         this._lineEndings = FileUtils.getPlatformLineEndings();
     }
     
@@ -167,6 +169,7 @@ define(function (require, exports, module) {
             var result = new $.Deferred();
             var path = this.path;
             var createIfNew = this.createIfNew;
+            var isStateJson = this.isStateJson;
             var self = this;
             
             if (path) {
@@ -175,10 +178,12 @@ define(function (require, exports, module) {
                     if (err) {
                         if (createIfNew) {
                             // Unreadable file is also unwritable -- delete so get recreated
-                            if (err === FileSystemError.NOT_READABLE || err === FileSystemError.UNSUPPORTED_ENCODING) {
+                            if (isStateJson && (err === FileSystemError.NOT_READABLE || err === FileSystemError.UNSUPPORTED_ENCODING)) {
                                 appshell.fs.moveToTrash(path, function (err) {
                                     if (err) {
-                                        console.log("Cannot move unreadable preferences file to trash: " + path);
+                                        console.log("Cannot move unreadable preferences file " + path + " to trash!!");
+                                    } else {
+                                        console.log("Note: Old Preferences file " + path + " has been moved to trash and recreated. You may refer to the deleted file in trash in case you want to recreate it from the former one!!");
                                     }
                                 }.bind(this));
                             }
@@ -198,7 +203,7 @@ define(function (require, exports, module) {
                         try {
                             result.resolve(JSON.parse(text));
                         } catch (e) {
-                            if (createIfNew) {
+                            if (isStateJson && createIfNew) {
                                 // JSON parsing error -- start from scratch
                                 console.log("Invalid JSON settings at " + path + "(" + e.toString() + ") -- contents reset");
                                 result.resolve({});
--- a/src/preferences/PreferencesBase.js
+++ b/src/preferences/PreferencesBase.js
@@ -145,14 +145,14 @@ define(function (require, exports, module) {
      * 
      * @constructor
      * @param {string} path Path to the preferences file
-     * @param {boolean} createIfNew True if the file should be created if it doesn't exist.
+     * @param {boolean} createIfMissing True if the file should be created if it doesn't exist.
      *                              If this is not true, an exception will be thrown if the
      *                              file does not exist.
      * @param {boolean} isStateJson True if the specified preferences file is state.json 
      */
-    function FileStorage(path, createIfNew, isStateJson) {
+    function FileStorage(path, createIfMissing, isStateJson) {
         this.path = path;
-        this.createIfNew = createIfNew;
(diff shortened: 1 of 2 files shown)


src/preferences/PreferencesBase.js
```diff
@@ -173,6 +174,14 @@ define(function (require, exports, module) {
                 prefFile.read({}, function (err, text) {
                     if (err) {
                         if (createIfNew) {
+                            // Unreadable file is also unwritable -- delete so get recreated
+                            if (err === FileSystemError.NOT_READABLE || err === FileSystemError.UNSUPPORTED_ENCODING) {
+                                appshell.fs.moveToTrash(path, function (err) {
```
redmunds: It doesn't seem right introducing `FileSystemError` and `appshell.fs` here. Maybe a new `File.prototype.moveToTrash()` method?

Are these the right errors to look for? WHat about `FileSystemError.NOT_SUPPORTED` ?

src/preferences/PreferencesBase.js
```diff
@@ -189,7 +198,13 @@ define(function (require, exports, module) {
                         try {
                             result.resolve(JSON.parse(text));
                         } catch (e) {
-                            result.reject(new ParsingError("Invalid JSON settings at " + path + "(" + e.toString() + ")"));
+                            if (createIfNew) {
+                                // JSON parsing error -- start from scratch
+                                console.log("Invalid JSON settings at " + path + "(" + e.toString() + ") -- contents reset");
+                                result.resolve({});
```
redmunds: This blows away all data. We could also try to move file to trash here in case user wants to try to retrieve some data.

src/preferences/PreferencesBase.js
```diff
@@ -173,6 +174,14 @@ define(function (require, exports, module) {
                 prefFile.read({}, function (err, text) {
                     if (err) {
                         if (createIfNew) {
```
redmunds: Is the `createIfNew` flag safe to use? Should it check explicitly for `state.json`? Seems like we'd want to handle `brackets.json` the same way. What about project pref files?

src/preferences/PreferencesBase.js
```diff
@@ -173,6 +174,14 @@ define(function (require, exports, module) {
...
+                            // Unreadable file is also unwritable -- delete so get recreated
+                            if (err === FileSystemError.NOT_READABLE || err === FileSystemError.UNSUPPORTED_ENCODING) {
+                                appshell.fs.moveToTrash(path, function (err) {
+                                    if (err) {
+                                        console.log("Cannot move unreadable preferences file to trash: " + path);
```
redmunds: I just realized that we talked about making this a modal dialog -- let me know if that's true or writing to console is ok.

src/preferences/PreferencesBase.js
rroshan1: How about creating a backup file "state.json.backup" in the same directory and then deleting all text from the "state.json" file. We can follow it up with a modal dialog box with message- "Contents of corrupted "state.json" file has been cleared. You can recreate the contents manually referring the backup file "state.json.backup" created in the same directory."

redmunds: @rroshan1 

> How about creating a backup file "state.json.backup" in the same directory and then deleting all text from the "state.json" file.

That's essentially what my suggestion does. With my suggestion, the backup file is in the Trash/Recycle so you don't need to worry about creating unique names for duplicate items or with the user needing to know that files are building up and to delete them. Also, Brackets automatically creates a new "state.json", so no need to manipulate the contents of that file.

src/preferences/PreferencesBase.js
rroshan1: I think we should have "move to trash" for state.json only which is not generally intended to be user-editable. 
We may not want silly errors (like a missing comma) in the brackets.json/.brackets.json file causing it to be moved to trash. Being served the corrupted brackets.json file for editing would be a better solution as it is currently.

src/preferences/PreferencesBase.js
rroshan1: @redmunds I think that makes good sense!!

src/preferences/PreferencesBase.js
```diff
@@ -189,7 +203,13 @@ define(function (require, exports, module) {
                         try {
                             result.resolve(JSON.parse(text));
                         } catch (e) {
-                            result.reject(new ParsingError("Invalid JSON settings at " + path + "(" + e.toString() + ")"));
+                            if (isStateJson && createIfNew) {
+                                // JSON parsing error -- start from scratch
+                                console.log("Invalid JSON settings at " + path + "(" + e.toString() + ") -- contents reset");
```
prksingh: Will the file get deleted in this case, when there is no error in reading the file but in parsing.

src/preferences/PreferencesBase.js
```diff
@@ -189,7 +203,13 @@ define(function (require, exports, module) {
                         try {
                             result.resolve(JSON.parse(text));
                         } catch (e) {
-                            result.reject(new ParsingError("Invalid JSON settings at " + path + "(" + e.toString() + ")"));
+                            if (isStateJson && createIfNew) {
```
prksingh: Where do we create this file when `createIfNew` is true. I am unclear what is the meaning of this flag createIfNew. From its name it appears that we will create a new file after checking if one does not already exist. I can prefs file is getting updated in `save` but `createIfNew` is not being checked there

rroshan1: We check this flag before calling 'result.resolve({})' which eventually (after multiple calls) saves the new 'state.json' file.
I agree the flag name is not very clear. Renaming it to a more appropriate 'createIfMissing'.

src/preferences/PreferencesBase.js
rroshan1: No, in this case the file is not deleted just its content.

prksingh:
Merging.

src/preferences/PreferencesBase.js
```diff
@@ -144,13 +145,15 @@ define(function (require, exports, module) {
...
      *                              If this is not true, an exception will be thrown if the
      *                              file does not exist.
+     * @param {boolean} isStateJson True if the specified preferences file is state.json 
      */
-    function FileStorage(path, createIfNew) {
+    function FileStorage(path, createIfMissing, isStateJson) {
         this.path = path;
-        this.createIfNew = createIfNew;
+        this.createIfMissing = createIfMissing;
+        this.isStateJson = isStateJson;
```
nethip: @prksingh @rroshan1  Please rename this variable to something more generic, not `isStateJson`. If we have a similar requirement in the future we could use this flag, and in that case `isStateJson` does not make sense.

rroshan1: Agree with you. Will make it more generic!!

PR #10403, opened 2015-01-17
SVG code hints cleanup

sprintr:
- Attributes are now unique.
- Added `fill` values to `animateTransform`, `animateMotion`, `animateColor` and `set` tags.
- Named Color Support

### DIFF
--- a/src/extensions/default/SVGCodeHints/SVGAttributes.json
+++ b/src/extensions/default/SVGCodeHints/SVGAttributes.json
@@ -11,6 +11,15 @@
     "animate/fill": {
         "attribOptions": ["freeze", "remove"]
     },
+    "animateColor/fill": {
+        "attribOptions": ["freeze", "remove"]
+    },
+    "animateMotion/fill": {
+        "attribOptions": ["freeze", "remove"]
+    },
+    "animateTransform/fill": {
+        "attribOptions": ["freeze", "remove"]
+    },
     "animateTransform/type": {
         "attribOptions": ["rotate", "scale", "skewX", "skewY", "translate"]
     },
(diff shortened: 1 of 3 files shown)


src/extensions/default/SVGCodeHints/SVGAttributes.json
```diff
@@ -11,6 +11,15 @@
     "animate/fill": {
         "attribOptions": ["freeze", "remove"]
     },
+    "animateColor/fill": {
+        "attribOptions": ["freeze", "remove"]
+    },
```
redmunds: Why are there separate entries for each `[tag]/fill` case when the options are the same (i.e. `["freeze", "remove"]`) ?

sprintr: Earlier it was using alias in such situations, that was removed for readability. I should use `type` property for each of these attributes.

redmunds: I'm not sure what you mean. Locally, I replaced all of the `*/fill` entries with:

```json
    "fill": {
        "attribOptions": ["freeze", "remove"]
    },
```

and it seems to work correctly.

sprintr: I was saying that `animate*/fill` should only hint `freeze` and `remove`, so we can use `fill` for color names.

> I'm not sure what you mean.

```xml
"animate*/fill": {
    "attribOptions": ["freeze", "remove"]
},
"fill": {
    "type": "color"
},
```
Then we can use its type property for hinting color names.

redmunds: OK, that makes sense. Do you want to add color name support in this PR so that it's clear why it was done this way?

sprintr: Yes, color name support will be added in this PR.

sprintr:
@redmunds Added color names support.

redmunds:
Looks good. Merging.

PR #10413, opened 2015-01-19
Highlight scrollbar tickmark for the current Find match

peterflynn:
Highlight scrollbar tickmark for the current Find match differently from the other matches (similar to the text highlighting within the editor: the current match is orange and the other matches are yellow).

Also, fix a bug where the wrong text highlight color was seen when a file had many search results (> `FIND_HIGHLIGHT_MAX`) or was too big to search for all matches at once (> `FIND_MAX_FILE_SIZE`): ensure `.CodeMirror-searching` is always applied to the current-match highlight, even when other matches aren't highlighted.

I thought this would be a nice enhancement to get into 1.2 :-)

### DIFF
--- a/src/search/FindReplace.js
+++ b/src/search/FindReplace.js
@@ -163,6 +163,7 @@ define(function (require, exports, module) {
                 }
             }
         
+            console.assert(state.matchIndex !== -1);
             if (state.matchIndex !== -1) {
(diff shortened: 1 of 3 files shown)


prksingh:
@peterflynn Changes look good to me. This would be really useful too :) Merging for 1.2

Issue #10370, opened 2015-01-13
New Untitled doesn't respect project tab settings


cfjedimaster:
I've got a project with the following settings:
{
    "sbruchmann.staticpreview.basepath": "/Users/ray/Dropbox/Writing/Cordova book/",
	"spaceUnits":2,
	"sortDirectoriesFirst":true,
	"useTabChar":false
}

Basically spaces instead of tabs and size == 2. I did CTRL+N to make a new blank document and it did not pick up on this. Instead it defaults to Tabs and size 4.

When I edited it in the Untitled doc and opened *another* one, it seemed to remember the right settings though so I couldn't recreate it.

redmunds:
I'm seeing something similar with New Untitled docs and the `wordWrap` setting. I have this global preference:

```json
    "language": {
        "unknown": {
            "wordWrap": false
        }
    }
```

Problem is that trying to toggle on View > Word Wrap does nothing.

peterflynn:
The issue is probably PreferencesManager not using the project scope because the Untitled documents don't have a well-defined location on disk yet (so they'll just use your Brackets global settings instead).  We probably shouldn't pass any path for Untitled documents, so that they default to the current project's scope.

PR #10422, opened 2015-01-20
Adding fix for Issue#10370 ...Changing PrefManager to use project default path for new docs

prksingh:
#10370  Changed _buildContext in PreferenceManager to use the project path in case of untitled documents

### DIFF
--- a/src/preferences/PreferencesManager.js
+++ b/src/preferences/PreferencesManager.js
@@ -421,12 +421,7 @@ define(function (require, exports, module) {
     function _buildContext(filename, languageId) {
         var ctx = {};
         if (filename) {
-            ctx.path = filename;
-            // In case of untitled documents we want to use the project root instead of
-            // the random generated file path. This is to resolve #10370
-            if (!ProjectManager.isWithinProject(filename)) {
-                ctx.path = ProjectManager.getProjectRoot().fullPath;
-            }
+            ctx.path = ProjectManager.getPreferenceContextPath(filename);
         } else {
             ctx.path = currentFilename;
         }
--- a/src/preferences/PreferencesManager.js
+++ b/src/preferences/PreferencesManager.js
@@ -35,6 +35,7 @@ define(function (require, exports, module) {
     var OldPreferenceStorage    = require("preferences/PreferenceStorage").PreferenceStorage,
         AppInit                 = require("utils/AppInit"),
         Commands                = require("command/Commands"),
+        ProjectManager          = require("project/ProjectManager"),
         CommandManager          = require("command/CommandManager"),
         DeprecationWarning      = require("utils/DeprecationWarning"),
         FileUtils               = require("file/FileUtils"),
@@ -421,6 +422,9 @@ define(function (require, exports, module) {
         var ctx = {};
         if (filename) {
             ctx.path = filename;
+            if (!ProjectManager.isWithinProject(filename)) {
+                ctx.path = ProjectManager.getProjectRoot().fullPath;
+            }
         } else {
             ctx.path = currentFilename;
         }
--- a/src/preferences/PreferencesManager.js
+++ b/src/preferences/PreferencesManager.js
@@ -422,6 +422,8 @@ define(function (require, exports, module) {
         var ctx = {};
         if (filename) {
             ctx.path = filename;
+            // In case of untitled documents we want to use the project root instead of
+            // the random generated file path. This is to resolve #10370
             if (!ProjectManager.isWithinProject(filename)) {
                 ctx.path = ProjectManager.getProjectRoot().fullPath;
             }
--- a/src/document/DocumentManager.js
+++ b/src/document/DocumentManager.js
@@ -192,6 +192,22 @@ define(function (require, exports, module) {
         }
         return result;
     }
+    /**
+     * Returns whether the given filePath is of an untitled File
+     * @param {!string} filePath
+     * @return {boolean} true if given filePath is of an untitled File, false otherwise
+     */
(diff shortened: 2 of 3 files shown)


src/preferences/PreferencesManager.js
```diff
@@ -35,6 +35,7 @@ define(function (require, exports, module) {
     var OldPreferenceStorage    = require("preferences/PreferenceStorage").PreferenceStorage,
         AppInit                 = require("utils/AppInit"),
         Commands                = require("command/Commands"),
+        ProjectManager          = require("project/ProjectManager"),
```
dangoor: I'm not sure you want to get an import loop going here... I guess it's okay because nothing is using `ProjectManager` at the root of the file, but these kinds of cross-module dependencies tend to increase over time and that has led to tricky to figure out bugs.

dangoor:
We had some discussions early with this prefs system about handling of files that aren't in the project. Applying project configuration to a file that's not in the project is likely to cause confusion because of differences in indentation, linting settings, etc.

I totally agree with the premise: untitled files should use the project prefs, but I think you'd want to use a mechanism that restricts this change to just untitled files.

peterflynn:
@dangoor I think it depends on the feature...  For features that aren't really file-specific, it could be very confusing to lose the project-level settings when you happen to have an external file active.  E.g. if you run Find in Files you may not get the right filters.  I think we decided that the right fix there is to have features like that explicitly request the root project scope when fetching prefs, instead of the default current-file-driven scope... but that was never implemented afaik.

For Untitled files, it could be weird if you ultimately save to a path outside the project -- it would use the project's settings while untitled but then switch to default settings once saved, which seems a bit unexpected.  Maybe that's still the lesser of two evils, but I'm not 100% convinced...

dangoor:
@peterflynn Requesting project prefs should work with `PreferencesManager.get("key", PreferencesManager.CURRENT_PROJECT)`.

One thing I forgot to mention in my first response was that for files outside of the project it would ultimately be good to search for .brackets.json (or .editorconfig) above the file that's outside of project scope. This adds a fair bit of complexity, though, which is why we didn't implement this initially.

Regarding Untitled files, it seems like you only have two options:

1. Use only the user-level prefs (the current behavior?)
2. Use the project settings

The user experience is off in option 1 if the user does save in their project, and it's off in option 2 if they save outside of the project. I'm guessing that creating a new file within the project is likely to be the more common case, which would make option 2 better.

peterflynn:
The experience wouldn't be off in option 2 _if_ files outside the project still use the project's settings rather than the global default.  You're right that the ideal is to look up the folder tree to try to find some _other_ project settings file for external files... I think the sticking point there was that we didn't want to make prefs fetches async.  (Maybe we could solve that by giving DocumentManager some side channel to tell PrefsManager to pre-cache that info before the Document-creation promise resolves... but that's getting pretty far afield of this PR).

Anyway, to avoid controversy it seems best for this PR to only apply to Untitled documents for now.  @prksingh one way we could do that is to add a `DocumentManager.isUntitledPath()` API, so that PreferencesManager could check for specifically untitled files without having to know what the special dummy paths look like.  (It could also be "semi-private," with `_` prefix, if we don't think anyone other than PrefsManager should have a valid reason to call it).

dangoor:
Yes, you're right that if files outside the project scope use the project settings then there would be no disconnect.

Not to get too far off of this PR, but I wanted to note that since prefs have events for changes, it is possible to asynchronously search for preferences files up the file tree, as editorconfig does, while still having synchronous prefs lookups. Whenever a new prefs file is encountered, change events would be fired for the prefs present in the file, causing everything to update as needed.

It seems reasonable to ask on brackets-dev what people think about using project prefs for files outside of the project. The least intrusive way to fix the issue discussed here is what you suggest at the end: just apply to Untitled documents for now.

peterflynn:
@dangoor Btw, `PreferencesManager.get("key", PreferencesManager.CURRENT_PROJECT)` doesn't work -- `PreferencesManager._getContext()` -> `_buildContext()` blows away the scope order and replaces it with one for the current file.  Bug?

The inline editor uses some [special code](https://github.com/adobe/brackets/blob/master/src/editor/MultiRangeInlineEditor.js#L67-L70) to get at project prefs -- but that's a different case where it's dealing with view state (which has no project scope, but has a project layer), and it wants to force writing to project-specific prefs even if there's no current project-specific value set.  In that case a seemingly _different_ prefs bug API bug was at play that I think we talked about over email last fall:

> it seems like if you call:
`PreferencesManager.setViewState(“foo”, value, { location : { scope: "user", layer: "project" } })`
…the preference is silently dropped on the floor.  We get to ProjectLayer.set(), and getPreferenceLocation() returns undefined if the preference isn’t already set for the current project.  You have to specify layerID explicitly or it won’t work.
 
> One thing that was confusing to me is that the ProjectLayer.set() docs seem to say layerID is optional: “If no layerID is provided, then the current project path is used.”

For simple retrieval, I think maybe you just need to make an object where `scopeOrder` matches `PreferencesManager.CURRENT_PROJECT` but there's an added `path` field set to the project root folder.

dangoor:
@peterflynn Yes, I think you're right that the behavior of `PreferencesManager.get("key", PreferencesManager.CURRENT_PROJECT)` should be considered a bug. I think the `contextBuilder` is relatively new.

prksingh:
I have added `DocumentManager.isUntitledPath` API but calling it directly from PrefManager is causing a cyclic dependency error (DocumentManager has a dependency on PreferenceManager) . Since ProjectManager already has a DocumentManager reference, I have moved the untitled check there.

@peterflynn Can you review this?

PR #10456, opened 2015-01-24
Resize the Sidebar to fill the whole window now possible (#6822)

lucaslouca:
This change was made to avoid issue #6822. 

The user can now resize the Sidebar to fill the whole window. 

When the user resizes the Window an upper bound width (maxsize) is set for the sidebar. This upper bound is computed by taking into consideration the main-toolbar width (on the right) and also the fact of leaving enough space for the Sidebar resizer icon to appear when the user hovers over the sidebar edges.

Also important is the scenario when the Sidebar is at its maximum size (filling the whole window) and the user decreases the Window's width. In that case the Sidebar's width should also be "dragged-along" and its width should be decreased to avoid the issue #6822 all over again. The content's left position must also be adjusted in that case.

### DIFF
--- a/src/view/WorkspaceManager.js
+++ b/src/view/WorkspaceManager.js
@@ -70,6 +70,7 @@ define(function (require, exports, module) {
     function calcAvailableHeight() {
         var availableHt = $windowContent.height();
 
+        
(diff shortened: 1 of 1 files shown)


ficristo:

Issue #10228, opened 2014-12-19
Mac: Cmd+Delete should trigger Don't Save in Save Changes dialog


jeremiahlee:
This is the Mac OS X system default for save dialog boxes.

JeffryBooher:
http://support.apple.com/en-us/HT201236

pthiess:
@jeremiahlee As a workaround you may use tab to go to the Don't Save button.

peterflynn:
Closing as fixed

PR #10459, opened 2015-01-24
[#10228] Cmd+Delete now triggers Don't Save in Save Changes dialog

lucaslouca:
Cmd+Delete (Mac) now triggers Don't Save in Save Changes dialog. This is the system's default behaviour (see http://support.apple.com/en-us/HT201236).

(Issue #10228)

### DIFF
--- a/src/widgets/Dialogs.js
+++ b/src/widgets/Dialogs.js
@@ -179,8 +179,8 @@ define(function (require, exports, module) {
                 $focusedElement.click();
             }
         } else if (brackets.platform === "mac") {
-            // CMD+D Don't Save
-            if (e.metaKey && (which === "D")) {
+            // CMD+Backspace Don't Save
+            if (e.metaKey && (e.which === KeyEvent.DOM_VK_BACK_SPACE)) {
                 if (_hasButton(this, DIALOG_BTN_DONTSAVE)) {
                     buttonId = DIALOG_BTN_DONTSAVE;
                 }


src/widgets/Dialogs.js
```diff
@@ -180,7 +180,7 @@ define(function (require, exports, module) {
             }
         } else if (brackets.platform === "mac") {
             // CMD+D Don't Save
```
MarcelGerber: You should rather call this CMD+Delete, right?
Also, I think CMD+D still should trigger Don't Save, shouldn't it?

lucaslouca: Hello Marcel,

according to Apple CMD-Backspace is the default for Don't Save. Or did I got it wrong :/

Anyhow I have updated the comment. Thanks for the tip.


Regards,
Lucas L.

MarcelGerber: From what I can tell (not that I have a Mac...), it's Cmd+Delete:
image

lucaslouca: Hello,

Apple keyboards sometimes have only one key for deletions (that is, one key that acts as a backspace and delete key). This key is located on the upper right corner of the keyboard. For instance on my MacBook Pro its labeled 'Delete' and on my wireless Apple keyboard it is just labeled with a left pointing arrow ('<----'). So hitting CMD+Backspace/Delete triggers the Don't Save action.

CMD+D is also bind on the "Duplicate" action in Brackets and doesn't seem to trigger Don't Save at all under the Mac. 

I have tested the Don't Save behaviour on Apple's standard apps (like TextEdit, Preview) and CMD+D doesn't seem to trigger Don't Save either.
Whereas CMD+Backspace does.



Best Regards,
Lucas

MarcelGerber: Ah ok, didn't know that. As stated above, I don't own a Macbook.

lucaslouca: Yeah, at least with the keyboards that I own from Apple that is the case. If I'm not mistaken, on older versions of Mac OS X 
the Don't Save action was bound to CMD+D... but not anymore :\

peterflynn:
@lucaslouca It sounds like some Mac apps use Cmd-D also (e.g. #283 mentions this), so this should probably be `||`ed with the original check.

Also, could you squash this down to one commit?  3-4 commits for a single-line change makes the git history a little sloppy.

Looks great otherwise though!

peterflynn:
Ah, I forgot that the existing Cmd-D code has been broken for a long time, due to conflicting with a menubar shortcut (see #4135 / [scoped keybindings story](https://trello.com/c/I07x1pO0/1002-more-flexible-context-sensitive-scoped-key-event-handling)).  So this actually looks fine as is -- if you can just squash the commits, I think we can go ahead & merge.

lucaslouca:
Hello,

done :)

peterflynn:
Looks good -- thanks!

PR #10468, opened 2015-01-26
apply theme to .force-syntax-highlighting

mat-mcloughlin:
I've added changed where the theme gets applied so it also applies to any element with the `.force-syntax-highlighting` class

### DIFF
--- a/src/view/ThemeManager.js
+++ b/src/view/ThemeManager.js
@@ -167,7 +167,7 @@ define(function (require, exports, module) {
             filename: fixPath(theme.file._path)
         });
 
-        parser.parse("#editor-holder {" + content + "\n}", function (err, tree) {
+        parser.parse("#editor-holder, .force-syntax-highlighting {" + content + "\n}", function (err, tree) {
             if (err) {
                 deferred.reject(err);
             } else {


MiguelCastillo:
@mat-mcloughlin I like the idea.  Initially the rules were not restricted to just the editor, then we restricted them to the editor because we didn't want to allow theme authors to mess with the rest of the application because the DOM could change and then themes could potentially be broken.

But I like the opt-in approach.  @peterflynn , do you have any objections to this change?

mat-mcloughlin:
You'll be making me happy if you do :)

image

MarcelGerber:
Btw, this was discussed in https://groups.google.com/forum/#!topic/brackets-dev/6_TFeUSUf80.

MiguelCastillo:
Oh thanks for the link!

@mat-mcloughlin I think the use case for this class goes beyond to the scope of syntax highlighting, we should come up with a more generic name.  Meaning that adding this class will allow somebody to style the tree, for example.  If this is only for syntax highlighting, why do you need to rely on themes though?  Couldn't you have your extension load its own css/less and add styling that way?  Check out how I do it in [ternific](https://github.com/MiguelCastillo/Brackets-Ternific) - [here](https://github.com/MiguelCastillo/Brackets-Ternific/blob/master/views/styles/ternific.css).  You can load up your css file like [this](https://github.com/MiguelCastillo/Brackets-Ternific/blob/master/main.js#L27), for example.

Let me know what you think.

MiguelCastillo:
Thinking about it a bit more, this needs to be thought out a bit more.  This actually falls in the theming for the whole tool, which we have a trello card for.  This quick one off patch is most likely going to be obsoleted.  I don't think we should add this - I don't know how whole editor theming is going to play out.

@mat-mcloughlin  I really suggest adding your own style as I described in my previous post. Let me know if I can help you there.

MarcelGerber:
I still think it's a good idea. He tries to match the appearence of his code hints popup with how it'll look like after being inserted into the editor.
Also, I wouldn't call it whole-tool themes/schemes/WhateverYouWannaCallIt. In this case, the extension explicitly asks to be styled using themes, while whole-tool themes would do it the other way round, they'd actively apply their styles to the whole Brackets UI.

mat-mcloughlin:
Hey, I'll take a look but the point of this is that it would apply the users chosen theme to the code hint popup.

MiguelCastillo:
@mat-mcloughlin did you end up resolving your issue?

mat-mcloughlin:
No I ended up leaving it. Its a nice to have and didn't want to implement a complicated solution until I knew what you were planning regrading implementing this in to core

MiguelCastillo:
Fair enough.  I agree there is value in having this non-intrusive solution that most likely will not interfere with any further work on theming.

src/view/ThemeManager.js
```diff
@@ -167,7 +167,7 @@ define(function (require, exports, module) {
             filename: fixPath(theme.file._path)
         });
 
-        parser.parse("#editor-holder {" + content + "\n}", function (err, tree) {
+        parser.parse("#editor-holder, .force-syntax-highlighting {" + content + "\n}", function (err, tree) {
```
MiguelCastillo: Let's change `.force-syntax-highlighting` to something more generic because we are not enforcing that css rule to only apply to syntax highlighting, which make the name misleading.  Let's call it something more along the lines of `.theme-override`.

MiguelCastillo:
@dangoor @redmunds  are you guys ok with this minor tweak to enable folks to accomplish their goals more easily?  I think this adds value.

The caveat of course is that any future DOM changes outside of `#editor-holder` have a chance of breaking extensions, so we would want to add a h1-bolded disclaimer on this.  This is the only reason I have been a bit hesitant about it.

dangoor:
Since this is opt-in, it seems reasonable to me.

mat-mcloughlin:
so should I change it to `theme-override`?

MiguelCastillo:
`theme-override` or another name that is generic...  If there are no options to chose from, then we can just go with `theme-override`.

redmunds:
Looks good to me.

> The caveat of course is that any future DOM changes outside of `#editor-holder` have a chance of breaking extensions

I don't see how this change has any effect on future DOM changes breaking extensions. Let me know if I'm missing something.

redmunds:
How about `.apply-editor-coloring` or `.apply-editor-syntax-coloring` ?

MiguelCastillo:
@redmunds if someone writes an extension that for example styles some part of the file tree using this new `class`... And the file tree goes through a change in which a class is renamed/removed or a div is added, then such extension could break because they are writing css rules that target something that changed.

Also, the class name could be more generic.  Using that class will enable someone to theme the file tree.  At which point `.apply-editor-coloring` is no longer accurate.

redmunds:
@MiguelCastillo 

> if someone writes an extension that for example styles some part of the file tree using this new class... And the file tree goes through a change in which a class is renamed/removed or a div is added, then such extension could break because they are writing css rules that target something that changed.

That's the same as it's always been. I guess the point is that by providing this mechanism, we're endorsing these types of extensions so it will be more likely. As you said, we should have a disclaimer that we don't guarantee that this will never change and extension authors should watch for API changes.

MiguelCastillo:
@redmunds yeah exactly :)   `endorsing` is the key.

mat-mcloughlin:
@redmunds @MiguelCastillo is that a yea or nay :)

MiguelCastillo:
@mat-mcloughlin its a yea :)   just brining everyone up to speed on the implications.

abose:
Tagging @swmitra

PR #10469, opened 2015-01-26
Fix #15 Added HTMLHinter extension as submodule to brackets

puzzledShark:
Added the htmlhinter extension to brackets as a submodule, and also added it so it runs by default in extensionloader


redmunds:
@KeniMorri I don't think you intended to submit this to Brackets repo, right? Probably meant for your fork of Brackets.

KeniMorri:
@redmunds Yeah, my sincerest apologies, I screwed up.

PR #10477, opened 2015-01-28
Fix Korean translation of split-view menu items

mixed:
The CMD_SPLITVIEW_VERTICAL is up/down split but The (left/right) is understandable.
For example, Japanese is not a up/down but a left/right.

```
//current
"CMD_SPLITVIEW_VERTICAL"    : "위/아래로 분할",
"CMD_SPLITVIEW_HORIZONTAL"  : "왼쪽/오른쪽으로 분할",

//propose
"CMD_SPLITVIEW_VERTICAL"    : "왼쪽/오른쪽으로 분할",
"CMD_SPLITVIEW_HORIZONTAL"  : "위/아래로 분할",

//japanese==propose
"CMD_SPLITVIEW_VERTICAL": "左右分割",
"CMD_SPLITVIEW_HORIZONTAL": "上下分割",
```

### DIFF
--- a/src/nls/ko/strings.js
+++ b/src/nls/ko/strings.js
@@ -242,9 +242,9 @@ define({
     "RIGHT"             : "오른쪽",
 
     "CMD_SPLITVIEW_NONE"        : "나누지 않음",
-    "CMD_SPLITVIEW_VERTICAL"    : "위/아래로 분할",
-    "CMD_SPLITVIEW_HORIZONTAL"  : "왼쪽/오른쪽으로 분할",
-    "SPLITVIEW_MENU_TOOLTIP"    : "에디터를 위/아래 또는 왼쪽/오른쪽으로 분할합니다",
+    "CMD_SPLITVIEW_VERTICAL"    : "왼쪽/오른쪽으로 분할",
+    "CMD_SPLITVIEW_HORIZONTAL"  : "위/아래로 분할",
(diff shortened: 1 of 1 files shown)


nethip:
@mixed you mentioned above that 

```
//japanese==propose
"CMD_SPLITVIEW_VERTICAL": "左右分割",
"CMD_SPLITVIEW_HORIZONTAL": "上下分割"
```

is Japanese. When I punched this into google translate, google is saying detected lanugage is Chinese. Am I misunderstanding something here.

And also could you have someone do a linguistic review for this change?

peterflynn:
@mixed So if I understand correctly, you're proposing to just swap the strings for `CMD_SPLITVIEW_VERTICAL` and `CMD_SPLITVIEW_HORIZONTAL`?  The tricky part is that both strings are actually ambiguous: does "up and down split" mean the _splitter line_ is vertical (creating left & right panels), or the _layout_ is vertical (top & bottom panels)?  Either meaning could be misinterpreted by some people.  The icons in the UI hopefully make it more clear what each string means.

I'm also a little worried it could be confusing to reverse the meaning of the two strings months after the original meaning shipped in a build...

Would a native speaker like @revi  or @taggon or @ds5apn like to weigh in?

taggon:
@peterflynn the both strings should be swapped. @mixed is right.

revi:
@mixed is correct, +1 as well.

peterflynn:
Thanks for the confirmation!

Issue #9785, opened 2014-11-03
[UX] Inactive panel gets dimmed in split view in release v1.0 


MikeSpock:
This isn't necessarily a bug, but definitely a usability issue. Split view is often used when you want to read one code and write the other. The inactive panel gets dimmed which makes the content hard to read.
This is especially bad when you want to use Brackets in a strongly lit enviroment.

peterflynn:
We discussed this again in review today and here's what we want to do:

* Editor content does not dim anymore when a pane is inactive.  (If people want this effect, they can install the "Zen Panes" extension).
* Pane header _does_ still dim when pane is inactive.
* If users report that this makes it too hard to tell the pane is focused (and the dimming/"Zen Panes" effect isn't a good enough fix), then we can revisit this and consider other treatments like a pane border or a brighter highlight in the pane header.

le717:
@peterflynn +1 I like that idea. I'd rather install an extension dim the pane rather than undim it.

prksingh:
@peterflynn Please review

peterflynn:
Closing -- this will be fixed in 1.2 per my [comment above](https://github.com/adobe/brackets/issues/9785#issuecomment-69669669).  Please let us know if you have any more feedback.

PR #10495, opened 2015-01-30
Fixing UX issue of inactive panel getting too dim to read text for dark ...

prksingh:
...themes.

Bug: #9785 [UX] Inactive panel gets dimmed in split view in release v1.0

### DIFF
--- a/src/styles/brackets.less
+++ b/src/styles/brackets.less
@@ -327,7 +327,6 @@ a, img {
     /* Placeholder shown when there is no editor open */
 
     .view-pane {
-        opacity: 0.7; // quieten inactive pane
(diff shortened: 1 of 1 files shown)


peterflynn:
Looks good to me!  @larz0 do you think the headers need any tweaks to make it more obvious which one is active now?

split panes

I think the active line highlight + blinking cursor make it pretty clear already... but I figure it's worth getting a second opinion :-)

larz0:
@peterflynn looks good to me too 👍 like you said, it's pretty clear already. Thanks @prksingh.

peterflynn:
All good to merge then!

Issue #10492, opened 2015-01-30
Uploading/installing extension fails is package.json has BOM (byte order marker)


peterflynn:
`JSON.parse()` fails when passed a string starting with a UTF byte order marker.

Unfortunately, when package validation fails this yields a confusing error message -- "Unexpected token" with no visible sign of what token or where.  Here's a screenshot from scriptin/brackets-indent-softwraps#2:

![screenshot](https://cloud.githubusercontent.com/assets/3739277/5967148/221b1010-a7bf-11e4-96fc-f0fb60625df8.png)

Extension registry upload presumably gives a similar error.

It's debatable whether a .json file starting with a BOM is considered valid, but we should either
(a) be lenient and accept it (i.e. strip out the BOM manually before trying to parse it); or
(b) give a clearer error message in Extension Manager & the extension registry.

If we decide to strip out the BOM something like [this code](https://github.com/nategood/httpful/blob/fca7e7b3eff49e40c7c93d79e928f8ecc6388f88/src/Httpful/Handlers/MimeHandlerAdapter.php#L44-L53) should do the trick.

peterflynn:
Note: the reason the error message is missing the token is that the UTF-16 BOM (which a UTF-8 BOM turns into when read by Brackets? - presumably since JS native strings are UTF-16ish) also happens to be the code for the character "zero-width nonbreaking space" so the browser renders it as nothing.

le717:
I vote for stripping the BOM as well, and I might be able to submit a patch for this in the coming week if no one else wants to take it before then.

Out of off-topic curiosity, because proper multiple-encoding file support is still on the backlog, couldn't something a tad similar to the code you linked be used to support editing of non-UTF 8 files (as in, detect, store, and strip BOM on file read, then rewrite BOM + content on save)?

PR #10502, opened 2015-01-31
Remove BOM and nonbreaking space from package.json

le717:
For #10492.

@peterflynn Surprisingly, the fix for the issue was not what what you first suggested. As you see in the first commit, I tried that and it did not work.

Also ping @scriptin as this issue was discovered through their extension.


le717:
FYI, I will flatten this before merging. I kept the first attempt commit for the time being in case it does work and I Did It Wrong.&trade;

ficristo:
Instead of this approach, wouldn't be better use `require` to read the `package.json` instead of using `fs.existsSync` & `fs.readFile`?
If it still will fail to read the file would be a problem of nodejs and not of Brackets.

BTW, if you think your are approach is better, in the second commit it seems to me you aren't stripping the BOM anymore: could you rename the function to reflect more what it does?

ficristo:
@le717 do you have time to look at this?

ficristo:
@zaggino reminded me, in another PR, that node require does caching so I'm not sure if we can use here.
The `stripBOM` function is still unclear to me: it doesn't seem to strip the bom and I'm not sure sure if we should replace the special char with a space instead.
@le717 can you verify?

src/extensibility/node/package-validator.js
```diff
@@ -173,6 +173,15 @@ function findCommonPrefix(extractDir, callback) {
 }
 
 /**
+ * Strip the BOM from the file if it has one.
+ * @param {String} data The raw file contents.
+ * @return {String} The same data, sans the BOM.
+ */
+function stripBOM(data) {
+    return data.substr(data.indexOf("{")).replace(/\u0000/g, "");
```
zaggino: this looks a bit weird, I believe this implementation is better:
https://github.com/sindresorhus/strip-bom/blob/master/index.js

zaggino (changes requested):

src/extensibility/node/package-validator.js
le717: Would you prefer I reimplement that library or would it be better to import and `require` it? Just thinking in case it would have use in other places.

zaggino: We could probably have it as an npm dependency, nothing wrong with that. But also pulling it out into some utils would be fine with me too.

PR #10503, opened 2015-02-01
Czech translation for v1.0

dvorapa:
- forked from kvarel/patch-14 branch
- updated link to czech translation of Adobe Third party software notices website
- updated sha of the forked root commit

### DIFF
--- a/src/nls/cs/strings.js
+++ b/src/nls/cs/strings.js
@@ -34,6 +34,7 @@ define({
     "GENERIC_ERROR"                     : "(chyba {0})",
     "NOT_FOUND_ERR"                     : "Soubor nenalezen.",
     "NOT_READABLE_ERR"                  : "Soubor nelze číst.",
+    "EXCEEDS_MAX_FILE_SIZE"             : "Soubor větší než {0} MB nemůže být otevřen v aplikaci {APP_NAME}.",
     "NO_MODIFICATION_ALLOWED_ERR"       : "Cílová složka nemůže být změněna.",
     "NO_MODIFICATION_ALLOWED_ERR_FILE"  : "Oprávnění neumožní provádět změny.",
     "CONTENTS_MODIFIED_ERR"             : "Soubor byl změněn mimo aplikaci {APP_NAME}.",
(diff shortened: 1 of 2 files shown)


src/nls/cs/strings.js
```diff
@@ -410,9 +422,11 @@ define({
     "CMD_TWITTER"                         : "{TWITTER_NAME} - Twitter",
     "CMD_ABOUT"                           : "O aplikaci {APP_TITLE}",
     "CMD_OPEN_PREFERENCES"                : "Otevřít soubor s preferencemi",
+    "CMD_OPEN_KEYMAP"                     : "Otevřít uživatelské klávesové zkratky",
 
     // Řetězce pro main-view.html
     "EXPERIMENTAL_BUILD"                   : "experimentální verze",
+    "RELEASE_BUILD"                        : "verze",
```
abose: Google says - "Verze" translates to version and stavět translates to 'build'. Is this intended or should we use stavět.

dvorapa: "Stavět" is a verb meaning "to construct" or "to build" (building, structure). There are two possibilities, how to translate "build" in computer development domain. The first is "verze" and the second is "sestavení". I don't think, that the second possibility fits better. Trust me, I'm from the Czech republic. And btw, Google Translator is not good translator if you want to check czech language. You should use Lingea if you want to check translations

abose: Thanks! just wanted to confirm before merging.

dvorapa: no problem

peterflynn: @dvorapa Just to explain the process... Ideally we'd like another native speaker to review translation work.  If no one is able to review, then we use Google Translate to check for typos and other obvious errors the best we can.

In this case since most of these translations came from @kvarel, I think _you_ basically count as the reviewer anyway (assuming you looked over all these new strings) -- so we maybe could have skipped that step.

dvorapa: I haven't known about this process before. Is there any plan to check it also with a grammar check tool? In my opinion it would be maybe better than Google Translator. Because as you can see in https://github.com/dvorapa/brackets/commit/692fb164c937595623c24794b170611d010c2297 commit from #10563 pull request, there were some typos and gramatical mistakes in the translation.

PR #10514, opened 2015-02-03
Fix for ImageView bug 5960 & 10180

EJanuszewski:
@prafulVaishnav Hopefully i've done it correct this time :)


prafulVaishnav:
Thanks @EJanuszewski ..

Issue #10507, opened 2015-02-02
Startup flicker


Code-DJ:
Is it possible to reduce or remove the startup flicker. On the Mac I have Brackets 1.1.0-15558 with the Dark theme. When the app is opened, it opens with dark gray background, then switches to white background, finally loads the editor UI. All of this happens within a second but it's still a little unpleasant.

Even leaving the background dark gray would be a much nicer UX.

codymaines:
I started looking into this. On Windows there's a slightly different, but equally frustrating start of white background before the .dark class gets applied to the body of the app. Changing my Windows them to 'High Contrast' did not change this.

My initial thought is to add a rule for the body background color to start the same as the top nav & side bar.

PR #10534, opened 2015-02-06
#10507 - to reduce dark theme flicker, set .body & .not-editor backgroun...

codymaines:
...ds to @project-panel-base-color

### DIFF
--- a/src/styles/brackets.less
+++ b/src/styles/brackets.less
@@ -47,6 +47,7 @@ html, body {
     color: @bc-text;
     height: 100%;
     overflow: hidden;
+    background-color: @project-panel-base-color;
 
     /* Turn off selection for UI elements */
     .user-select(none);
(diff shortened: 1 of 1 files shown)


nethip:
With this change this is how the initial screen looks

capture. 

I believe it is better to have this change as an extension rather than core change as it does not fit with other components of Brackets default theme. Thanks for taking effort in trying to fix the flickering issue.

peterflynn:
I don't think this fixes the flicker -- it just reverses it so that people using the light theme see it flicker dark first.  I think the real fix would be to make sure the theme preferences are available and ThemeManager has initialized before we render the HTML UI.  This would be tricky for third-party themes since we don't load extensions until slightly later -- but as a workaround, ThemeManager could easily cache (in viewstate prefs) the light/dark flag from the last-used theme so that metadata is available immediately during startup.

PR #10541, opened 2015-02-06
Update the selected file if a renamed folder is a prefix of the path

### LINKED ISSUE #10527: Renaming folder breaks file selection
1. Create a new project
2. Create a file "file.js" at the top level
3. Create a folder "folder" at the top level
4. Create a file "file-in-folder.js" inside "folder"
5. Select "file-in-folder.js" in the file tree (not the working set)
6. Rename "folder" to something else
7. Try to select "file.js" in the file tree

Result: The filename turns blue but does not get the background highlight. Clicking back and forth between file-and-folder.js and file.js in the file tree causes inconsistent state (the correct document doesn't open in the editor and sometimes the filename goes into rename mode).

njx:
This seems to fix #10527, but I don't know the code well enough to know if there would be other side-effects - I just made this fix by local inspection of the code. @dangoor could you take a look?

This fix exposes another issue, though: if you follow the steps in #10527, but rename the folder so it clashes with an existing filename, you get an error dialog, but now the selection is messed up again. The reason is that we update `this._selections.selected` before we know whether the rename was successful or not. (I think you would get a similar bug without this fix if you were just renaming an ordinary file.) So there's probably more work that would need to be done to make sure that gets updated only after we know the rename succeeded, but that looks like a more complex fix and someone who knows the code better should probably get into it :)

### DIFF
--- a/test/spec/ProjectModel-test.js
+++ b/test/spec/ProjectModel-test.js
@@ -732,13 +732,41 @@ define(function (require, exports, module) {
                 });
                 
                 it("adjusts the selection if the renamed file was selected", function () {
+                    spyOn(model, "_renameItem").andReturn(new $.Deferred().resolve().promise());
                     model.setSelected("/foo/afile.js");
                     model.startRename("/foo/afile.js");
                     model.setRenameValue("something.js");
                     model.performRename();
                     expect(model._selections.selected).toBe("/foo/something.js");
                 });
+                
+                it("does not adjust the selection if renaming it fails", function () {
+                    spyOn(model, "_renameItem").andReturn(new $.Deferred().reject().promise());
+                    model.setSelected("/foo/afile.js");
+                    model.startRename("/foo/afile.js");
+                    model.setRenameValue("something.js");
+                    model.performRename();
+                    expect(model._selections.selected).toBe("/foo/afile.js");
+                });
 
+                it("adjusts the selection if a parent folder is renamed", function () {
+                    spyOn(model, "_renameItem").andReturn(new $.Deferred().resolve().promise());
+                    model.setSelected("/foo/afile.js");
+                    model.startRename("/foo");
+                    model.setRenameValue("bar");
+                    model.performRename();
+                    expect(model._selections.selected).toBe("/bar/afile.js");
+                });
+                
+                it("does not adjust the selection if renaming a parent folder fails", function () {
+                    spyOn(model, "_renameItem").andReturn(new $.Deferred().reject().promise());
+                    model.setSelected("/foo/afile.js");
+                    model.startRename("/foo");
+                    model.setRenameValue("bar");
+                    model.performRename();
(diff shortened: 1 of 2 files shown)


njx:
Oops, the original commit had a problem. I amended and force-pushed it, so if you looked at this branch in the last five minutes please reset :)

njx:
Also - this probably needs unit tests, but I don't have time right at the moment - if someone else wants to take a stab that would be great; otherwise I could look at writing some later (but am not sure what the timeline is to get this into 1.2).

njx:
OK, I improved the fix to not adjust the selection if the rename fails, and added tests for these cases.

I wasn't entirely sure if the selection adjustment needed to happen in the `FILE_CREATING` case - logically it should if the new item remains selected after the rename, but it seems like in the actual UI that's not the case (the selection switches to the working set), and there are no existing tests around the file tree selection after a create. However, it seemed safer to do the adjustment in both cases (since it was doing it in both cases before my change).

@dangoor would be great if you could do a quick review. If we do take this fix, it would also be good for people to do some testing around both creating files/folders and renaming them.

test/spec/ProjectModel-test.js
```diff
@@ -732,13 +732,41 @@ define(function (require, exports, module) {
                 });
                 
                 it("adjusts the selection if the renamed file was selected", function () {
+                    spyOn(model, "_renameItem").andReturn(new $.Deferred().resolve().promise());
```
njx: This wasn't necessary before because the selection update would happen even if the rename failed (since the files don't actually exist). In general, it would probably be a good idea to either add a mock filesystem to these tests, or always stub out `_renameItem` and any other functions that access the real filesystem.

dangoor:
I haven't tested this, but it looks good to me.

njx:
@madanbn @ryanstewart I feel pretty good about this fix if you want to try to get it in for 1.2 - but would suggest you have folks do some focused testing around creating and renaming files/folders in the file tree; I did some testing around basic scenarios but didn't go deep.2

peterflynn:
Looks like there's additionally a view/pane management bug here.  If I follow the #10527 steps and then select the nested "file-in-folder.js" again afterward, I get this error in the console:

```
/..<snip>../folderRenamed/file-in-folder.js found in pane working set but pane.addView() has not been called for the view created for it
```

Same thing happens any time I select that file in the future (or select it in the working set).

peterflynn:
@njx Otherwise this looks great and the tests are good too.  Is it worth fixing the Pane issue before merging this?  If the bug repros all the way back to 0.44, merging it doesn't seem urgent for 1.2, but OTOH the Pane issue does appear to be a second bug in different code...

peterflynn:
Yeah, looking at it more, it seems separable enough that this is worth merging now.  I'll put up a separate PR for the Pane issue.

PR #10546, opened 2015-02-07
Fixed typos in MainViewFactory docs

sprintr:
[See](https://github.com/adobe/brackets/blob/master/src/editor/ImageViewer.js#L465)


peterflynn:
Good catch, thanks!

PR #10554, opened 2015-02-09
Add extension changelogs to the extension manager

zaggino:
As discussed here: https://groups.google.com/forum/#!topic/brackets-dev/MgoqM-nFJR0
and mentioned here: https://trello.com/c/86DGrlbJ/794-extension-update-notifications

The extension adds a "See changelog" button to the current extension manager and on demand downloads information about an extension from GitHub. Result looks something like this:

image

Sadly, code heavily depends on GitHub not changing anything - it parses both `CHANGELOG.md` if available and first commit page of an extension which puts latest 35 commits to the generated changelog. I've tried using GitHub API but it has limit of only 60 requests per hour without authorizations which basically disables this feature after opening a few logs.

I'd appreciate some comments onto this before putting any more time into it. Thanks :)

### DIFF
--- a/src/extensions/default/ExtensionUpdateNotifications/ChangelogDialog.js
+++ b/src/extensions/default/ExtensionUpdateNotifications/ChangelogDialog.js
@@ -1,15 +1,28 @@
-/*global define*/
+/*global brackets, define, Mustache*/
 
 define(function (require, exports, module) {
     "use strict";
 
-    var ChangelogDownloader = require("./ChangelogDownloader");
+    var Dialogs                 = brackets.getModule("widgets/Dialogs");
+    var Strings                 = brackets.getModule("strings");
+    var ChangelogDownloader     = require("./ChangelogDownloader");
+    var changelogDialogTemplate = require("text!ChangelogDialogTemplate.html");
 
     function show(extensionId) {
         ChangelogDownloader.downloadChangelog(extensionId)
             .then(function (changelog) {
+
                 console.log("changelog for " + extensionId);
                 console.log(JSON.stringify(changelog, null, 4));
+
+                var compiledTemplate = Mustache.render(changelogDialogTemplate, {
+                    title: extensionId,
+                    changelog: changelog,
+                    Strings: Strings
+                });
+
+                Dialogs.showModalDialogUsingTemplate(compiledTemplate);
+
             })
             .catch(function (e) {
                 console.error(e);
--- a/src/extensions/default/ExtensionUpdateNotifications/ChangelogDialog.js
+++ b/src/extensions/default/ExtensionUpdateNotifications/ChangelogDialog.js
@@ -1,0 +1,21 @@
+/*global define*/
+
+define(function (require, exports, module) {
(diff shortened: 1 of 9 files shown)


mackenza:
@zaggino this is great.  I tried it out with my extensions and it now works great with the http://keepachangelog.com/ Markdown I am using.

One thing I noticed is that when it cannot get/use/parse a changelog it doesn't do anything.  Would it be possible to display some sort of indication that the button was pressed but no changelog will be shown?

le717:
@zaggino Just pulled the latest changes, and it is still awesome. :D If we could get this and the quick open rewrite in v1.3, that'd be nearly all I want for that release. 

I did catch the following URL conversion error with https://github.com/mackenza/Brackets-PHP-SmartHints, not sure why that one breaks and the others don't:

changelog-url-break

mackenza:
The anomoly with my SmartHints repo is that I renamed it and so it appears as a new extension to the Registry.  I am not sure that is why it might not be parsing right but it is a difference between that repo and others that work.

mackenza:
Would love to see this merged

nethip:
@zaggino Great work!
@mackenza Sure! I am CCing our product manager to check if we can get this into the product.

CC @ryanstewart

ficristo:
I must admit I didn't look at the rest of the code when I saw we are scraping github. Sincerly I'm against this PR simply because of that.
In the wiki there was some discussion to ask extension authors to have a changelog file in their extension that follow a particular changelog format.
I think a better plan to this would be to change the registry to require a changelog file and show that.
I would be awesome to be able to show only the part of changelog since the last install.
So if I have brackets-git v1.0.0 and there is an update to v1.2.0 I would like to see only what changed between those versions.

mackenza:
seeing as this PR is over a year and a half old... it would seem better to just close it as "won't merge" than suggest changes.

zaggino:
feel free to close this and delete the code, i'm not planning on putting any more work into this

ficristo:
@mackenza maybe you already know but I think is worth clarifying a couple of things:
- I don't work for Adobe
- I'm a committer only since this august
- I looked in the past weeks to even older PR

I don't like the current situation in which Adobe is focusing mostly on Dreamweaver but I'm trying to do my best to still improve Brackets. This means, for me, looking even to old PR and suggest changes.
Because I think they deserve at least an answer.
For this particular PR, I hope to have understood it correctly, and yes, as is, I'm more on the side "won't merge". But personally I think is worth putting a plan to have something like this.

mackenza:
@ficristo  yes, I am aware you are a community core member not an Adobe employee and thanks for stepping up and trying to tackle the PRs/issues in the face of limited assistance from our project stewards. The community appreciates it.

My comment about this PR was not meant as a negative. I just think that when a PR is this old and is from another core contributor (like @zaggino ), then it's probably past the time where it's going to be worked on and if not mergable as-is, then it's probably worth just closing it with a comment like "good idea but PR is too old to merge" or something like that?

Just a thought.

PR #10558, opened 2015-02-10
Make LiveDevMultiBrowser's Launcher instance configurable via public set Launcher() method

humphd:
We're working on using the new LiveDevMultiBrowser code to allow an in-browser Brackets to use an iframe-based browser and postMessage-based protocol, see https://github.com/humphd/brackets/issues/27.  One of the snags we've hit is that we need a way to override how the default browser is launched, and the `Launcher` is currently not configurable.

This patch follows the same pattern currently in place for swapping the protocol's transport (which we also make use of in order to use postMessage to/from the iframe), and adds a `setLauncher()` method.  If you'd consider this API change, we'd be able to get our own iframe-based launcher hooked into live dev.

cc @busykai, @sebaslv, @sedge

### DIFF
--- a/src/LiveDevelopment/LiveDevMultiBrowser.js
+++ b/src/LiveDevelopment/LiveDevMultiBrowser.js
@@ -87,7 +87,7 @@ define(function (require, exports, module) {
         LiveDevServerManager = require("LiveDevelopment/LiveDevServerManager"),
         NodeSocketTransport  = require("LiveDevelopment/MultiBrowserImpl/transports/NodeSocketTransport"),
         LiveDevProtocol      = require("LiveDevelopment/MultiBrowserImpl/protocol/LiveDevProtocol"),
-        Launcher             = require("LiveDevelopment/MultiBrowserImpl/launchers/Launcher");
+        DefaultLauncher      = require("LiveDevelopment/MultiBrowserImpl/launchers/Launcher");
     
     // Documents
     var LiveCSSDocument      = require("LiveDevelopment/MultiBrowserImpl/documents/LiveCSSDocument"),
@@ -113,6 +113,12 @@ define(function (require, exports, module) {
      * Protocol handler that provides the actual live development API on top of the current transport.
      */
     var _protocol = LiveDevProtocol;
+
+    /**
+     * @private
+     * Current browser launcher for preview.
+     */
+    var _launcher;
     
     /**
      * @private
@@ -525,7 +531,7 @@ define(function (require, exports, module) {
         // open default browser
         // TODO: fail?
         // 
-        Launcher.launch(url);
+        _launcher.launch(url);
     }
     
     /**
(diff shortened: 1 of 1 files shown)


sebaslv:
@humphd, looks good to me having this mechanism to dynamically set a different ```Launcher``` instance. We have a unique launcher instance now to make it compatible with CDT-based live dev UI but, in the near future we would like to decouple launch mechanism from live dev session to better support launching multiple clients per session.

Just curious, have you consider some way of static configuration also for your use case? I mean, something like having a pref that allows setting the ```Launcher``` instance to be loaded during initialization  rather than always loading DefaultLauncher and then dynamically swapping it?

humphd:
OK, great.  Using this patch (which I landed on our brackets fork), we've just today been able to get Live Dev working in the browser with an iframe and postMessage--it's awesome :)  It would be great if you'd take this patch upstream so we don't have to modify brackets at all to make it work.

Regarding your other question, I've been wondering about that too.  I don't love the idea of doing it with prefs, since that still forces changes in brackets proper, and isn't something you can do from an extension.  I like the model you have with the live dev server manager, where you can register a server with a priority.  In an ideal world, it would be great if an extension could somehow provide the transport, launcher, and maybe a few other internal implementations for Live Dev, and have Live Dev's `init` use these instead of starting with the defaults and then being forced to switch them.  It's a bit of a pain to switch them, because you have to get the timing right, or your transport/launcher get overwritten by the defaults again.

I had the same issue with swapping out my in-browser filesystem (https://github.com/filerjs/filer) for the appshell's impementation, which requires me to change the brackets source internally.  It would be great if there was a way for extensions to override more of the internal code, but that's a larger issue, and I don't want to block this bug on figuring that out if possible.

busykai:
Sorry I haven't been able to get to this for a little while. I'll review and merge this code. We'd want to have as little diffs as possible with the mainstream Brackets. As for static configuration, we'll consider it as the next step (this one may actually help it).

src/LiveDevelopment/LiveDevMultiBrowser.js
```diff
@@ -800,6 +806,22 @@ define(function (require, exports, module) {
...
+     * (e.g., default browser, iframe-based browser, etc.)
+     * The launcher must provide the following method:
+     *
+     * - launch(url): Launch the given URL in the appropriate browser.
+     *
+     * @param {{launch: function(string)}} launcher
+     */
+    function setLauncher(launcher) {
+        if (!(launcher && launcher.launch)) {
+            throw new Error("LiveDevMultiBrowser.setLauncher(): launcher must have `launch` method");
```
busykai: We'd normally log an error instead of throwing it. However, it may still make sense to have an error to be able to trace back the origin of the call. Like so (note the commas, not concatenation):

```javascript
console.log("Invalid launcher object: ", launcher, new Error("LiveDevMultiBrowser.setLauncher()");
```

busykai: Well, and `return` from the `if`, of course...

busykai:
@humphd, looks good, only one minor comment.

humphd:
Thanks, fixed.

src/LiveDevelopment/LiveDevMultiBrowser.js
```diff
@@ -800,6 +806,23 @@ define(function (require, exports, module) {
...
+     * (e.g., default browser, iframe-based browser, etc.)
+     * The launcher must provide the following method:
+     *
+     * - launch(url): Launch the given URL in the appropriate browser.
+     *
+     * @param {{launch: function(string)}} launcher
+     */
+    function setLauncher(launcher) {
+        if (!(launcher && launcher.launch)) {
+            console.log("Invalid launcher object: ", launcher, new Error("LiveDevMultiBrowser.setLauncher()");
```
busykai: This has a syntax error. :)

humphd:
Ugh. Fixed.

humphd:
@busykai ping?

busykai:
Looks good! Merging. Sorry for the delay.

humphd:
Great, thanks!

PR #10567, opened 2015-02-10
Fix unit test that broke due to Find highlighting fix in PR #10413

peterflynn:
Fix unit test that broke due to a side fix in PR #10413, and also:
- Enhance whole FindReplace integration suite to verify that search tickmarks are rendered.
- Bump up wait time in WorkingSetView-test that was failing sporadically. The test had zero margin for error on the wait, which may be making it fragile.

### DIFF
--- a/test/spec/FindReplace-test.js
+++ b/test/spec/FindReplace-test.js
@@ -352,6 +352,10 @@ define(function (require, exports, module) {
                 });
             }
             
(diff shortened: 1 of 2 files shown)


peterflynn:
@prksingh, could you review?

prksingh:
Looks good. Merging

PR #10572, opened 2015-02-11
FR and JP translation fix for Getting Started index.html file

pantkowiak:


redmunds:
@nethip It looks like this was merged *after* you split off the `release` branch. If you want this in 1.2 it will need to be pulled in.

nethip:
@redmunds Thanks for pointing this out. I will merge this to release branch.

peterflynn:
Luckily there's nothing else on master that's not already on release, so this can be a straightforward merge from master->release without fear of bringing in any other unwanted commits.  If other stuff (not stable enough for 1.2) had landed on master in the meantime, we'd have to `git cherry-pick` this commit instead.

nethip:
@peterflynn I just merged these changes back to release. So this PR should be merged now. Thanks!

PR #10579, opened 2015-02-12
Consolidate code handling file-rename path fixups

### LINKED ISSUE #10527: Renaming folder breaks file selection

peterflynn:
Consolidate code among places that handle file-rename path fixups, which tends to be bug-prone:
- Find in Files rename handling - had a subtle bug with superset filenames similar to #9904
- File tree rename handling - see recent fix #10527, and has same #9904-like bug
- Pane rename handling - did not work for any folder rename (discovered while testing #10527)

Added unit tests for Pane and for the new shared FileUtils code.

Fix a bunch of misleading var names & docs related to the arguments
passed to rename event handlers. And clarify some other docs in Pane.

_Not_ needed for 1.2 -- all of the edge-case bugs fixed here have been around for months already.  This is more about improving maintainability going forward.

### DIFF
--- a/src/file/FileUtils.js
+++ b/src/file/FileUtils.js
@@ -41,6 +41,11 @@ define(function (require, exports, module) {
         StringUtils         = require("utils/StringUtils"),
         Dialogs;            // This will be loaded asynchronously
 
+    // Asynchronously loading Dialogs to avoid the circular dependency
+    require(["widgets/Dialogs"], function (dialogsModule) {
+        Dialogs = dialogsModule;
+    });
+
     
     /**
      * @const {Number} Maximium file size (in megabytes)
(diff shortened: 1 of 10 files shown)


src/file/FileUtils.js
```diff
@@ -41,6 +41,11 @@ define(function (require, exports, module) {
         StringUtils         = require("utils/StringUtils"),
         Dialogs;            // This will be loaded asynchronously
 
+    // Asynchronously loading Dialogs to avoid the circular dependency
+    require(["widgets/Dialogs"], function (dialogsModule) {
+        Dialogs = dialogsModule;
+    });
+
```
peterflynn: I moved this to the top just for clarity.  This change isn't related to any of the rename stuff, and doesn't affect the timing of when this module is loaded.

PR #10585, opened 2015-02-12
Updated The Dutch Translation

ghost:
@redmunds, hope I have done it the good way this time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "Brackets",
-    "version": "1.2.0-0",
-    "apiVersion": "1.2.0",
+    "version": "1.3.0-0",
+    "apiVersion": "1.3.0",
     "homepage": "http://brackets.io",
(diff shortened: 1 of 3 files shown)


prksingh:
@svenvNL Thanks for translation. I can see some other commits were pulled in from Master. Specifically one which changes version to 1.3 

Can you start a fresh pull request maybe, with only the translations?

svenvNL:
@prksingh I had this before, but some things broke because I did some stupid things whit Gitub. As you can see, only the translation will be merged. I changed the version back from 1.3 tot 1.2 in the pr.

prksingh:
@svenvNL Got it. It may still be a better idea to a start a new pull request with just the translation since it is one file. Also, the release branch has been merged with master so rather than merging this request to release and then again to master, it would be cleaner if you can start a fresh PR with the translation to master. Thanks!

PR #10588, opened 2015-02-12
Added fix for JS Hints, fixes Issue #10281

### LINKED ISSUE #10281: JS Code Hints parameter tips hidden beneath non-native menu bar
Please fix this.

noragrossman:
Should fix [Issue #1028](https://github.com/adobe/brackets/issues/10281)

### DIFF
--- a/src/extensions/default/JavaScriptCodeHints/ParameterHintManager.js
+++ b/src/extensions/default/JavaScriptCodeHints/ParameterHintManager.js
@@ -108,7 +108,11 @@ define(function (require, exports, module) {
         left = Math.max(left, editorLeft);
         left = Math.min(left, editorLeft + $editorHolder.width() - hintWidth);
 
-        if (top < 0) {
+        // Get the top of the editable document
+        var cm = session.editor._codeMirror;
+        var localTop = cm.charCoords({line: 0, ch: 0}).top;
+
+        if (top < localTop) {
             $hintContainer.removeClass("preview-bubble-above");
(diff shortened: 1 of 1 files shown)


src/extensions/default/JavaScriptCodeHints/ParameterHintManager.js
```diff
@@ -108,7 +108,11 @@ define(function (require, exports, module) {
         left = Math.max(left, editorLeft);
         left = Math.min(left, editorLeft + $editorHolder.width() - hintWidth);
 
-        if (top < 0) {
+        // Get the top of the editable document
+        var cm = session.editor._codeMirror;
+        var localTop = cm.charCoords({line: 0, ch: 0}).top;
```
peterflynn: This won't be correct if the document has been scrolled.

What you want is to compare `top` to the top bounds of the editor, which is probably something like `$editorHolder.offset().top`.  That ignores split views and looks at the bounds of the entire "main view" content area, but the z-ordering probably works out such that it's ok to overlap a neighboring split-view pane... worth double checking with testing though.

zaggino:
Not replicating in master anymore, closing without merge.

image

MarcelGerber:
Did you mean to reopen this PR?

zaggino:
It was meant to fix problem on Linux as related issue states, need to boot up linux and check there. I tested windows...

zaggino:
Confirmed fixed in latest Linux too.

PR #10592, opened 2015-02-13
Temporary Fix for weird project tree scrollbar on Mac

prafulVaishnav:
This is for #10175

### DIFF
--- a/src/styles/brackets_scrollbars.less
+++ b/src/styles/brackets_scrollbars.less
@@ -21,11 +21,15 @@
  *
  */
 
- /* Temporary fix for issue#10150 for weird project tree scrollbar behaviour. Scrollbar not appearin on hover. Forcing the project tree div to render displays the scrollbar. So setting transparent background color on hover.
+ /* Temporary fix for issue#10150 for weird project tree scrollbar behaviour on mac. Scrollbar not appearing on hover. Forcing the project tree div to render displays the scrollbar. So setting transparent background color of div on hover.
 */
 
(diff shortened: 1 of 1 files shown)


nethip:
@redmunds @peterflynn Could you guys take a look at this PR?

redmunds:
This fix also needs to be applied to the working set. Try selector `.platform-mac .working-set-view:hover`.

I found 1 subtle case on Yosemite (10.10) where the fix doesn't work:
1. Make Brackets window non-full-screen, and non-maximized so there is space to the left of window
2. Place mouse over project tree so scrollbar is displayed
3. Move mouse *slowly* to the left until you see the horizontal resize cursor (e.g <->)
4. Move mouse left again until it's outside Brackets window

Results:
Mouse is outside Brackets window, but scrollbar is still displayed. If this doesn't repro, then try moving *slower* in step 3.

nethip:
@redmunds Sure! we will extend this to working set views as well. Thanks for unit testing this in a detailed way!

@prafulVaishnav Please look at this failure case and see how this can be fixed.

prafulVaishnav:
@redmunds Thanks a lot.. I was unable to reproduce the bug on working set so just did fix for project tree only. I will do the same for working set to be on safer side. I checked the fix on 10.9, will check it for 10.10 too..

prksingh:
Can still see the bug with moving out of the brackets window slowly on the left. But this fixes most other common cases. Merging

PR #10602, opened 2015-02-15
use npm to download extension dependencies after installing from registry

zaggino:
Seems to be working fine. Tests included, but currently they are disabled in Gruntfile so they need to be run manually, wasn't investigating what was the reason for disabling the tests.


mackenza:
Is this being considered for merging?  I think it's a great addition speaking as some who has published a node-based extension with dependencies.

zaggino:
@mackenza not sure if this will make its way into Brackets, but I've been playing with a concept of going around Brackets' extension registry altogether and download whole extensions from npm. It works conceptually, but needs more testing and polishing, see here: https://github.com/zaggino/brackets-npm-registry

mackenza:
ah, I see.  That would closely resemble what Atom does with APM wrapping
NPM, yes?

zaggino:
To be honest, I've no idea what Atom does or doesn't do, never had time to take a closer look at it.

mackenza:
Yeah, basically this is what they do.  The have the Atom Package Manager
which seems to add some Atom-specific meta data fields to Package.json and
they use NPM as the actual dependency manager.

Atom is a hot mess, honestly.  Truly an editor without an identity.

nethip:
@zaggino Thanks for putting up this PR! Great work and thinking. I like the idea of keeping the node extensions lean and downloading the dependencies using npm. 

Couple of things to consider for this PR.

1) What changes do we need to do for our Registry(like changing validation logic e.t.c)?
2) I have some concerns about versioning of the dependent npm modules. If an extension developer mentions the version of the dependent module to be something like `>0.1.3`, what happens if the latest version of that dependent module is buggy and is not compatible with that extension?
3) What can we do something similar which can help extension developers who do not use node modules?

Talking about the third point, how about downloading the extension itself from just a link? Something like, the registry will only have links (mostly to GitHub repositories) and we download the extension itself from Github repository/external link followed by installing the dependencies(which this PR is about), if any. 


CC @ryanstewart @peterflynn

mackenza:
@nethip I would think point 2 is a concern for any dependency manager and not unique to this PR's application of NPM.  If a package author doesn't want to "roll the dice" on having package updates break their extension, they should be explicit about the dependency.  i.e. `1.0.3` rather than `>=1.0.3`

On 3, NPM is quickly becoming the dependency manager to rule them all.  Certainly it's not restricted to Node modules anymore.  I can find most of the modules I use for front end JS in there, as well.  I think what @zaggino is suggesting is we use the current package registry to download the actual extension zip (as we do today) but as part of the installation, rather than just unzipping the files, we also run `npm install` to pick up the dependencies.  This sounds like the best of both worlds.

zaggino:
@ficristo @petetnt @abose @ingorichter this would be worth a second look, it's no magic and it'd help a lot for extensions so they wouldn't have to bundle everything in a .zip file

package.json
```diff
@@ -12,6 +12,9 @@
         "branch": "",
         "SHA": ""
     },
+    "dependencies": {
+        "npm": "2.15.10"
```
petetnt: Have you tested this out with `npm@3`. I didn't have much issues (if any) from moving to `2->3` when 3 started rolling out but I don't know if it has some troubles with how Brackets resolves the extension dependencies.

src/extensibility/node/ExtensionManagerDomain.js
```diff
@@ -97,17 +98,61 @@ function _performInstall(packagePath, installDirectory, validationResult, callba
...
                 // The status may have already been set previously (as in the
                 // DISABLED case.
                 if (!validationResult.installationStatus) {
                     validationResult.installationStatus = Statuses.INSTALLED;
                 }
                 callback(null, validationResult);
+            };
+
+            if (err) {
+                fail(err);
```
petetnt: Nit: I prefer `return fail(err)`;

package.json
zaggino: No, I could test/fix this to work with 3 but original PR was put together when 2 was latest and I don't think it's worth spending time on it if it won't be merged before npm 4 comes out.

src/extensibility/node/ExtensionManagerDomain.js
```diff
@@ -97,17 +98,61 @@ function _performInstall(packagePath, installDirectory, validationResult, callba
...
+            if (packageJson && packageJson.dependencies) {
+                npmDependencies = Object.keys(packageJson.dependencies).map(function (key) {
+                    return key + "@" + packageJson.dependencies[key];
+                });
+            }
+
+            if (npmDependencies && npmDependencies.length > 0) {
+                npm.load(function (err, npm) {
+                    if (err) {
+                        fail(err);
```
petetnt: Nit: I prefer `return fail(err)`;

src/extensibility/node/ExtensionManagerDomain.js
```diff
@@ -97,17 +98,61 @@ function _performInstall(packagePath, installDirectory, validationResult, callba
...
+                packageJson = null;
             }
+
+            if (packageJson && packageJson.dependencies) {
+                npmDependencies = Object.keys(packageJson.dependencies).map(function (key) {
+                    return key + "@" + packageJson.dependencies[key];
+                });
+            }
+
+            if (npmDependencies && npmDependencies.length > 0) {
```
petetnt: Nit: Unneccessary `> 0` I think, but it does bring the intention more clear.

src/extensibility/node/ExtensionManagerDomain.js
```diff
@@ -97,17 +98,61 @@ function _performInstall(packagePath, installDirectory, validationResult, callba
...
+                npm.load(function (err, npm) {
+                    if (err) {
+                        fail(err);
+                        return;
+                    }
+                    npm.commands.install(installDirectory, npmDependencies, function (err, result) {
+                        if (err) {
+                            fail(err);
```
petetnt: Nit: I prefer `return fail(err);`

petetnt:
Looks good to me with some minor nits about personal preferences.

package.json
petetnt: Fair enough. `@3` would bring some nice benefits, but I do see how testing for it might be unfeasible at this point. I'd like to see this land sooner than the later though, preferably with `3` 😼

zaggino: The review I asked for is more about the idea of this then the code (I'm happy to fix/update it), when I first put up this PR there wasn't enough will to merge this and potentially deal with problems like invalid dependencies in package.json and etc. But I still believe this would help a lot of extension authors to move forward.

zaggino: Oh, and `npm@3` doesn't play well with node `0.10` so we need to merge two other PRs first.

petetnt: Yeah, good catch on that one. I'd love to get rid of both of those ancient dependencies at the same time 👍

ficristo:
I like the idea but I defer the final decision to others, I'm not confident on how this will interact with the registry. A part the following few questions, LGTM.

I suppose we can start to using `dependencies` in the root `package.json` if things aren't extensions, right?
Can you use `require` instead of `JSON.parse(fs.readFileSync...`?
The API of npm are really supported? Looking around seems not. Can you use instead an `exec` command or similar and run `npm install --production`?
Your npm extension provide a reinstall button, don't we need it?
On [VSCode](https://github.com/Microsoft/vscode/blob/1ae99cf1fbb92dcdcd85b6fd45c39f3534fa7213/src/vs/workbench/parts/files/browser/files.contribution.ts) by default they are exluding `**/.git/object/**` instead of `.git/**`, aren't your git extension affected?

zaggino:
@petetnt @ficristo updated the PR per your suggestions.
Tested by installing `https://github.com/zaggino/brackets-coffeelint` from an URL.

src/extensibility/node/ExtensionManagerDomain.js
```diff
@@ -97,17 +97,80 @@ function _performInstall(packagePath, installDirectory, validationResult, callba
...
+            };
+
+            if (err) {
+                return fail(err);
             }
+
+            var packageJson;
+
+            try {
+                packageJson = JSON.parse(fs.readFileSync(path.join(installDirectory, "package.json"), {
```
ficristo: Cannot you use `require` instead of `JSON.parse(fs.readFileSync`?

src/extensibility/node/ExtensionManagerDomain.js
```diff
@@ -97,17 +97,80 @@ function _performInstall(packagePath, installDirectory, validationResult, callba
...
+                    encoding: "utf8"
+                }));
+            } catch (e) {
+                packageJson = null;
+            }
+
+            if (packageJson && packageJson.dependencies) {
```
ficristo: Could you invert this condition and call immediatly `return finish();`?
We could reindent the block in this way and it would immediately clear we are looking for dependencies only (without looking at the else branch)

src/extensibility/node/ExtensionManagerDomain.js
```diff
@@ -97,17 +97,80 @@ function _performInstall(packagePath, installDirectory, validationResult, callba
...
+                var spawn = require('child_process').spawn;
```
ficristo: Put the require at the top with the others?

ficristo:
I didn't expect to become complex like this, but I believe is better than using npm API. Thank you @zaggino.
From my side looks good.

zaggino:
Have a look again guys, I've ripped out the spawn stuff into a separate function so the condition is much more readable. Also used `fs.readJson` which is used in Brackets to read `package.json` contents on multiple places (require does caching, we don't want that here).

zaggino:
Since 1.9 is too far away, can this be reviewed / merged for 1.8?

package.json
```diff
@@ -15,7 +15,8 @@
     "dependencies": {
         "anymatch": "1.3.0",
         "chokidar": "1.6.0",
-        "lodash": "4.15.0"
+        "lodash": "4.15.0",
+        "npm": "2.15.10"
```
ficristo: Do you think is worth using npm3 now that we are on node6?

src/extensibility/node/ExtensionManagerDomain.js
```diff
@@ -80,6 +81,54 @@ function _removeFailedInstallation(installDirectory) {
...
+
+    var exitCode;
+    child.addListener("exit", function (code) {
+        exitCode = code;
+    });
+
+    child.addListener("close", function () {
+        stderr = Buffer.concat(stderr);
+        stdout = Buffer.concat(stdout);
+        if (exitCode > 0) {
```
ficristo: I assume `exit` is always called and `close` is always called after it.
Otherwise is better to initialize `exitCode`.

src/extensibility/node/ExtensionManagerDomain.js
```diff
@@ -98,17 +147,40 @@ function _performInstall(packagePath, installDirectory, validationResult, callba
...
+            };
+
+            var finish = function () {
                 // The status may have already been set previously (as in the
                 // DISABLED case.
                 if (!validationResult.installationStatus) {
                     validationResult.installationStatus = Statuses.INSTALLED;
                 }
                 callback(null, validationResult);
+            };
```
ficristo: Could you move `fail` and `finish` function outside this block?
I think starting the block with the `if (err)` would make think a little bit clearer.

src/extensibility/node/ExtensionManagerDomain.js
```diff
@@ -98,17 +147,40 @@ function _performInstall(packagePath, installDirectory, validationResult, callba
...
+            } catch (e) {
+                packageJson = null;
             }
+
+            if (!packageJson || !packageJson.dependencies) {
+                return finish();
+            }
+
+            _performNpmInstall(installDirectory, function (err) {
+                return err ? fail(err) : finish();
```
ficristo: I think something like this would make easier to add / review code if we ever need it.
```js
if (err) {
  return fail(err);
}

finish();
```

ficristo (changes requested):
If something fails we remove the entire extension folder, right? So is correct to assume a reinstall button is not needed?

zaggino:
sorry I missed this one... ready to re-review @ficristo

ficristo:
@zaggino I downloaded your extensions brackets-coffeelint from github and then installed it using drag&drop. And worked great.
Then I modifyed the `package.json` of my local copy to add a dependency that doesn't exists.
I used again the drag&drop of the new zip and it failed to install, as expected.
What I feel a bit odd is that it removed the installation of my previous working copy. While I understand this behaviour when installing a new extension is a bit disappointing when you update an existing one.
Can we done something in this case? Sorry to realize this behaviour only now.

You should update `npm-shrinkwrap.json`.

zaggino:
@ficristo updated shrinkwrap and fixed the issue you've mentioned, please re-review

ficristo (approved):

ficristo:
Nice!
@petetnt, do you want to take another look?
@zaggino I'm not sure if it is necessary, but please open a new issue to update brackets-registry.

zaggino:
@ficristo Open what issue? Update how? Only thing that pops to my mind would be some kind of UI message - this extension has npm dependencies? Or am I missing something, I'm not familiar with the registry code.

madanbn:
moving it out to 1.9 release as we are closer to the 1.8 release time frame.

ficristo:
@zaggino actually I'm not particular familiar either, but in the registry there is the deps folder which contains a zip of the extensibily folder of brackets. I don't have any idea how it is generated but seems to me that it has to be keep in sync with brackets code.

zaggino:
Right, had a look, I'll resolve this too.

zaggino:
Anything missing for this except brackets-registry PR ?

zaggino:
image

swmitra (approved):

swmitra:
@zaggino The last pic has forced me to look at this PR :smile. 
I will test the changes and merge the PR today.

src/extensibility/node/ExtensionManagerDomain.js
```diff
@@ -90,6 +90,20 @@ function _removeFailedInstallation(installDirectory) {
 function _performInstall(packagePath, installDirectory, validationResult, callback) {
     validationResult.installedTo = installDirectory;
 
+    var fail = function (err) {
```
MiguelCastillo: question [NIT]: can you refresh my memory if we tend to have function expressions like this over function definitions?

src/extensibility/node/npm-installer.js
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Checks package.json of the extracted extension for npm dependencies
+ * and runs npm install when required.
+ * @param {Object} validationResult return value of the validation procedure
+ * @param {Function} callback function to be called after the end of validation procedure
+ */
+function performNpmInstallIfRequired(validationResult, callback) {
+
+    function finish() {
```
MiguelCastillo: aha! it is inconsistent. you are using a function definition here.  Do you mind making them at least consistent?

MiguelCastillo:
this looks good to me. @swmitra - I wont have time to test this tonight... so I am really interested to hear how your testing goes.  if for some reason you can't do it, let me know as well and I will find cycle for tomorrow.

src/extensibility/node/npm-installer.js
zaggino: https://github.com/adobe/brackets/pull/10602/commits/93a5604575227c81ca13f29757e60c81321897d7

ficristo:
Today I was thinking if this will work if the user is behind a proxy: do we need to pass it to npm script?

zaggino:
I believe that it will pick up whatever is configured for npm on the machine. npm's config is stored in user's folder so if the domain is running under the same user, everything should be allright.

swmitra (commented):
@zaggino Can you please pull from master and resolve the conflict. We are good to go with the PR 👍

ficristo:
> is configured for npm on the machine

Don't we use the npm version embedded in this PR? If so don't we need to pass the proxy somehow to npm?

zaggino:
- rebased as requested by @swmitra 
- added proxy support requested by @ficristo 
- tested both with and without proxy by installing `https://github.com/zaggino/brackets-coffeelint` from an URL

note: original bundled `request` was failing with my proxy but upgrade to the latest version fixed that, so I've added this to my commit even though it was not directly related - request is now managed as npm dependency (it can be in root as it's node-only dependency) and old request committed to the repository was removed

ficristo:
IMHO if we change extensibility/node in this way we should wait for #12933 first.

zaggino:
#12933 depends on the access to the registry machine and I don't know what else, so it might take a year or two to fix that...

zaggino:
removed `request` update which @ficristo complained about, should be ready to merge

swmitra:
Great work as always @zaggino 👍 .
I guess this PR is good to go now and if we merge now, it will give us some additional time for testing as well.

swmitra:
Merging now ...

zaggino:
btw, fun fact: this PR took exactly 2 years to merge
image

swmitra:
I bet it's better late than never 👍 
I hope we will never see a PR raised today getting merged on 15th Feb 2019 😃

zaggino:
Tested this by publishing a copy of archive from npm into brackets registry (TypeScript 0.9.4) and it works great. Thanks a lot guys for getting this to the core.

image

Issue #10617, opened 2015-02-17
File drop icon only shown when outside editor


peterflynn:
This is a regression from the new editor drag & drop feature -- bug goes away when disabled.

1. View a folder in Finder
2. Drag a file over top of the Brackets window

Result: green "(+)" icon is only shown when file is dragged over non-editor parts of the UI

Expected: icon consistently shown across the whole UI

Note: if you release the button and _drop_ the file over top of the editor, it still does open correctly though.

peterflynn:
This might be Mac-only btw.  The drag/drop implementation is very different between Mac & Win.

peterflynn:
Hmm, I'm also seeing some weird behavior where files on some external drives on my Mac actually don't open at all when dropped over the editor -- that would be more serious.

peterflynn:
I think I have a pretty simple fix for this, just by finessing our file-drop event handler vs. CM's text-drop one...  Still need to test on Windows though.

peterflynn:
Closing as fixed -- PR has landed

PR #10622, opened 2015-02-18
Disabling drag and drop of CM by default as it was leading to some defec...

prksingh:
...ts.

The following are the defects that got introduced because of CM drag and drop on Mac.

https://github.com/adobe/brackets/issues/10590
https://github.com/adobe/brackets/issues/10617

User still has the ability to enable this features through the preference "dragDropText": true.


nethip:
Merging...

nethip:
@sprintr Thanks for pointing out. I just updated it.

peterflynn:
@MarcelGerber Fyi, not sure if you saw this change yesterday -- because we found those two late-breaking bugs, we decided it would be safer to start out with this preference disabled by default.  I think I have a fix for #10617 that should be easy to get into 1.3, and hopefully we'll find some workaround for the CM/Chrome issue in #10590, so at that point it seems like we could flip the default back to true if desired...

PR #10639, opened 2015-02-20
Remove chrome=1, #10632

### LINKED ISSUE #10632: 'chrome=1' can be removed from all the Getting Started pages
The Getting Started pages have the following:
meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1"

'chrome=1' was for Chrome Frame which has long since been discontinued and is no longer available, and was only ever needed for old versions of Internet Explorer (for Windows XP) which is also no longer supported by Brackets.

Denisov21:
Fix #10632


samples/pt-pt/Primeiros Passos/index.html
```diff
@@ -132,4 +132,4 @@
     [::::::::::::::     ::::::::::::::]
     [[[[[[[[[[[[[[[     ]]]]]]]]]]]]]]]
 
--->
\ No newline at end of file
+-->
```
le717: This line here is the only error I see.

peterflynn:
@Denisov21 I'm not too bothered about that trailing newline, but you do need to merge with master.  There's no real merge conflict here, but in the Finnish version there was a change on the line just below the line you modified -- and since git is extra paranoid it flags that as a conflict anyway, and you can't clear that without merging.

Denisov21:
CLOSED the PR, I had made ​​a mistake. Now I have built a new PR correct!

Issue #10548, opened 2015-02-08
user stylesheet — feature request


tobibeer:
* allow to define a user stylesheet
* doesn't even have to be exposed in the ui
* just some user.css that to make theme-independent style-adjustments

I'll gladly remove this quickshot after: https://github.com/tobibeer/brackets-custom-css

peterflynn:
Since this can already be done by writing a custom extension -- or even more simply by using the extension you wrote -- and it's not requested very often, this probably isn't a very high priority item.  But I agree it could be nice to have...

tobibeer:
I just brought it up because adding an extension like the above to the extension repository seems a little overkill, maybe not.

As suggested, just including a css file at some well defined location (if existing) would do the trick neatly, even without any ui to access or create it.

Some simple instructions that read "to overrule theme styles, create and save changes to **user.css** at folder foo". Seems like something having 3 loc.

As for wanting or needing it: something like changing the font-size in the sidebar should be a very frequent request.

FabianCernota:
I'm working on it

peterflynn:
@tobibeer Note that almost 100% of the font size requests we've gotten were from people using high DPI displays, which is really a different problem that should be solved without the user having to customize anything.  The upcoming 1.2 build fixes this on Windows, and hopefully a [future version of Chromium](https://code.google.com/p/chromium/issues/detail?id=143619) will enable us to do the same on Linux.

FabianCernota:
Created an Pull-Request #10642

tobibeer:
The DPI issue is indeed a separte one, so fixing it via custom styles is surely not the right way. However, a few style customization requirements seem independent of that, imho.

onnlinetool:
Downloaden Downloaden Downloaden Downloaden Downloaden Downloaden Downloaden 

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

Downloaden =========>>>>>>>>>> http://bit.ly/2oiWvqi

PR #10642, opened 2015-02-21
Adding feature for adding own CSS to brackets GUI (#10548)

FabianCernota:
You can now adding own CSS declaration without using a extension.

Put an custom.css on the same level like config.json or index.html and activate it over View -> Themes

Reload the UI with F5

### DIFF
--- a/src/brackets.js
+++ b/src/brackets.js
@@ -342,6 +342,10 @@ define(function (require, exports, module) {
         // Add the platform (mac or win) to the body tag so we can have platform-specific CSS rules
         $("body").addClass("platform-" + brackets.platform);
         
+        var themePrefs = PreferencesManager.getExtensionPrefs("themes");
+        if (themePrefs.get("customCSS") === true) {
+            ExtensionUtils.addLinkedStyleSheet("custom.css");
+        }
         // Browser-hosted version may also have different CSS (e.g. since '#titlebar' is shown)
         if (brackets.inBrowser) {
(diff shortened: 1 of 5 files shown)


peterflynn:
I'm not sure we need a checkbox in the UI -- why not just load the file any time it exists?  The keymap.json file works that way and it seems fine.

To match how preferences and keybinding overrides work though, I wonder if we should listen for user changes to this file and update the Brackets UI immediately, without requiring a reload.  It's not hard -- e.g. [here's how KeyBindingManager does it](https://github.com/adobe/brackets/blob/release-1.1/src/command/KeyBindingManager.js#L1382-L1386) -- and re-replacing the original style tag should be easy too.  The only question is whether messing with CSS is so error-prone that it shouldn't go into effect without an explicit reload gesture...

FabianCernota:
Okey i will change it so that the CSS File will load if exists.

redmunds:
@FabianCernota Not sure if you saw my comment in IRC, but I think `ExtensionUtils` methods `addLinkedStyleSheet()` and `addEmbeddedStyleSheet()` will need to be fixed to detect "custom.css" and insert any new stylesheets *before* that stylesheet. Otherwise, "custom.css" will not override all extensions and themes.

sprintr:
I think it will be nice to put `custom.css` in `extensions` directory, (where we already have `disabled` and `user` directories) so it can be accessible from Brackets UI.

PR #10643, opened 2015-02-21
Remove chrome=1, (Fix #10632)

### LINKED ISSUE #10632: 'chrome=1' can be removed from all the Getting Started pages

Denisov21:
Fix #10632

### DIFF
--- a/samples/da/Kom godt i gang/index.html
+++ b/samples/da/Kom godt i gang/index.html
@@ -3,7 +3,7 @@
 
     <head>
         <meta charset="utf-8">
-        <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
+        <meta http-equiv="X-UA-Compatible" content="IE=edge">
         <title>KOM GODT I GANG MED BRACKETS</title>
         <meta name="description" content="En interaktiv guide til at komme godt i gang med Brackets.">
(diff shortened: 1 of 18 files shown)


Denisov21:
"Changes made -- ready for another review"

peterflynn:
@Denisov21 The diff looks fine, but as a matter of git cleanliness this should be 1 commit, not 18.  It looks like you're using github.com to edit the files online?  When editing multiple files, you should use git on your local computer instead so you can create a single commit with all your changes.  See [this Stackoverflow answer](http://stackoverflow.com/questions/17815895/can-i-edit-two-files-then-make-one-commit-using-github-web-based-editor) or [our quick setup instructions](https://github.com/adobe/brackets/wiki/How-to-Hack-on-Brackets#quick-start) for more info.

Denisov21:
@peterflynn It is true I did with github so online with Google Chrome.I am new and I use very little git. However, the link you sent me are very interesting.

PR #10645, opened 2015-02-21
Update README.md

Denisov21:
I have change Updating Existing Branch and Pull Request


peterflynn:
This conflicts with PR #9028, which I should really update and get landed...

@Denisov21 Could you give some background on why the extra steps are needed?  (Is that option sometimes _not_ selected by default?  Are there other ways to dismiss this page other than clicking Commit?)  The shorter the instructions are, the more likely people will read & follow them, so personally I think we should keep them streamlined if possible :-)

Denisov21:
@peterflynn Close, becouse I think you're right!

Issue #10634, opened 2015-02-19
Live Preview : Duplicates elements on copy/paste of entire document


gideonthomas:
Hi,

It seems like when you copy and paste an entire html document into the editor (that already has content), DOM elements in the body are duplicated in the Live Preview. This happens regardless of whether the new body content you paste in is the same or different.

I am running the latest version of Brackets on Mac OS X Mavericks.

Here's a screencap of it:
![livepreviewfail](https://cloud.githubusercontent.com/assets/5816130/6272186/c6feb5da-b833-11e4-90d0-2f43084f2842.gif)

Thanks :)

gideonthomas:
I dug into the code a bit and I think I tracked it down to this [function](https://github.com/adobe/brackets/blob/master/src/language/HTMLDOMDiff.js#L538) 

This [statement](https://github.com/adobe/brackets/blob/master/src/language/HTMLDOMDiff.js#L575) results in `null` because the new `<html>` tag gets a new `tagID` which cannot be mapped to a corresponding `<html>` tag in the old DOM structure. 

So, it fails the `if(oldElement)` check and executes [this](https://github.com/adobe/brackets/blob/master/src/language/HTMLDOMDiff.js#L614). In the function `generateChildEdits`, it fails [this](https://github.com/adobe/brackets/blob/master/src/language/HTMLDOMDiff.js#L361) condition, not entering the loop where the delete/insert operation (delete the old `html` tag with the old `tagID` and insert the new one) happens. It instead only does the insert operation.

Seems like the problem comes from passing in `null` as the first parameter into `generateChildEdits`. But not sure how to fix it though.

peterflynn:
Possibly related to #5336

MohabAli:
solved ... just disable (StaticServer)
go to ... Extension manager >> Default >> disable (StaticServer)
and use an alternative (Autosave) extension then enable (auto save (400 ms))
it will be in (File) after installation
https://github.com/nmaxcom/brackets-autosaver
run (live preview) enjoy

PR #10653, opened 2015-02-25
for #10634 Element duplication when copying root node

xyin96:
for #10634
fixed element duplication when copy/pasting html tag.

The issue was sending a null as the oldElement in generateChildEdits, which implies an insertion edit.
The fix was to check if there was a root node in the oldNodeMap, and passing that node as the oldElement so that generateChildEdits had a node to compare to.

### DIFF
--- a/src/language/HTMLDOMDiff.js
+++ b/src/language/HTMLDOMDiff.js
@@ -604,7 +604,7 @@ define(function (require, exports, module) {
                 // because it isn't the child of any other node. The browser-side code doesn't
                 // care about parentage/positioning in this case, and will handle just setting the 
                 // ID on the existing implied HTML tag in the browser without actually creating it.
-                if (!newElement.parent ) {
+                if (!newElement.parent) {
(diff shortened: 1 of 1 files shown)


humphd:
Fantastic.  Glad to see a patch for this so quickly.  @xyin96++

AJDBenner:
Just tested this code and I'm getting an error in the console: "data-brackets-id=1 not found".

xyin96:
Yeah this patch generates that error. I don't know how to fix it, but from what I've noticed, the referenced id is the tagID of the old <html> tag, which I assume has been replaced and deleted when the user copies and pastes an html document into the editor.

busykai:
@xyin96, @humphd let me think about this and experiment a little bit. it's quite possible that this case (entire DOM re-instrumentation) should be handled by different means, not by editing existing DOM on the remote.

Denisov21:
https://github.com/drewhamlett/brackets-beautify/issues/49

PR #10672, opened 2015-03-02
Update README.md

coliff:
Bump version number to 1.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ back and forth between your real source code and the browser view.
 * **Do it yourself.** Because Brackets is open source, and built with HTML, CSS
 and JavaScript, you can [help build](https://github.com/adobe/brackets/blob/master/CONTRIBUTING.md) the best code editor for the web.
 
-Brackets is at 1.0 and we're not stopping there. We have many feature ideas on our
+Brackets is at 1.2 and we're not stopping there. We have many feature ideas on our
 [trello board](http://bit.ly/BracketsTrelloBoard) that we're anxious to add and other
 innovative web development workflows that we're planning to build into Brackets. 
 So take Brackets out for a spin and let us know how we can make it your favorite editor. 


README.md
```diff
@@ -15,7 +15,7 @@ back and forth between your real source code and the browser view.
 * **Do it yourself.** Because Brackets is open source, and built with HTML, CSS
 and JavaScript, you can [help build](https://github.com/adobe/brackets/blob/master/CONTRIBUTING.md) the best code editor for the web.
 
-Brackets is at 1.0 and we're not stopping there. We have many feature ideas on our
+Brackets is at 1.2 and we're not stopping there. We have many feature ideas on our
```
redmunds: I don't think we want to update this for every release. I think the intent is more like "Brackets is at version 1 and we're not stopping there."

le717: How about "Brackets may have reached version 1, but we're not stopping there."?

peterflynn: @le717 Nice, I like that wording

coliff: LGTM :+1:

nethip:
This not only looks good to but inspiring as well :smile:

peterflynn:
@nethip I think there was a misunderstanding -- in the thread above we were talking about using a _different_ wording from what's in this PR, so we don't have to update the README every release.  So we wanted to wait to merge this until the wording was updated...  @coliff can you put up a new PR with the new wording?

coliff:
@peterflynn Sure - are we going with:
'Brackets may have reached version 1, but we're not stopping there.'
Please confirm and I'll open a PR.

nethip:
@peterflynn @coliff My Bad I don't know why I was under the impression that the suggested wording was already in the PR. I will review the new PR going to be uploaded by @coliff

redmunds:
@coliff 

> are we going with: 'Brackets may have reached version 1, but we're not stopping there.' ?
> Please confirm and I'll open a PR

Yes

PR #10683, opened 2015-03-03
Fix src/xorigin.js to not use function statement that isn't top-level and break strict mode when minifying

humphd:
Working on getting Brackets to run well in-browser, I'm playing with various minificaiton strategies.  While doing this I hit an issue where xorigin.js was breaking strict mode.  This fixes it.

See http://whereswalden.com/2011/01/24/new-es5-strict-mode-requirement-function-statements-not-at-top-level-of-a-program-or-function-are-prohibited/ for details on why this is bad.

### DIFF
--- a/src/xorigin.js
+++ b/src/xorigin.js
@@ -55,7 +55,7 @@
     var previousErrorHandler = window.onerror;
 
     // Our error handler
-    function handleError(message, url, line) {
+    var handleError = function(message, url, line) {
         // Ignore this error if it does not look like the rather vague cross origin error in Chrome
         // Chrome will print it to the console anyway
         if (!testCrossOriginError(message, url, line)) {
@@ -70,7 +70,7 @@
 
         // Restore the original handler for later errors
         window.onerror = previousErrorHandler;
-    }
+    };
 
     // Install our error handler
     window.onerror = handleError;


peterflynn:
I'm confused...  This function _is_ at the top level of its scope (not nested inside some other block construct like an `if`), and this module has `"use strict"` on it, so if this were a violation of strict mode this code wouldn't even run at all.

Do you know what Uglify transforms this code to that causes it to break?  Maybe looking at the diff would make things more clear...

humphd:
The problem isn't what's there now, it's how what's there gets transformed by more aggressive minification.  By doing the change I have in this PR, you end up with this (I've expanded it):

```js
!function() {
    "use strict";
    var testCrossOriginError;
    if (-1 !== navigator.userAgent.search(" Chrome/") ? testCrossOriginError = function(message, url, line) {
        return "" === url && 0 === line && "Script error." === message
    } : "Opera/" === navigator.userAgent.slice(0, 6) && (testCrossOriginError = function(message) {
        return "Uncaught exception: DOMException: NETWORK_ERR" === message
    }), "undefined" == typeof brackets && "file://" === document.location.href.substr(0, 7) && testCrossOriginError) {
        var previousErrorHandler = window.onerror, handleError = function(message, url, line) {
            if (testCrossOriginError(message, url, line))
                alert("Oops! This application doesn't run in browsers yet.\n\nIt is built in HTML, but right now it runs as a desktop app so you can use it to edit local files. Please use the application shell in the following repo to run this application:\n\ngithub.com/adobe/brackets-shell"), window.onerror = previousErrorHandler;
            else if (previousErrorHandler)
                return previousErrorHandler(message, url, line)
        };
        window.onerror = handleError
    }
}();
```
So this change is innocuous in the case that Brackets' source is just used as is, but necessary when minifiying the code and having the AST transformed.

humphd:
For reference, what happens when I don't change it:

```js
!function() {
    "use strict";
    var testCrossOriginError;
    if (-1 !== navigator.userAgent.search(" Chrome/"))
        testCrossOriginError = function(message, url, line) {
            return "" === url && 0 === line && "Script error." === message
        };
    else if ("Opera/" === navigator.userAgent.slice(0, 6))
        testCrossOriginError = function(message) {
            return "Uncaught exception: DOMException: NETWORK_ERR" === message
        };
    if ("undefined" == typeof brackets && "file://" === document.location.href.substr(0, 7) && testCrossOriginError) {
        var previousErrorHandler = window.onerror;
        function handleError(message, url, line) {
            if (testCrossOriginError(message, url, line))
                alert("Oops! This application doesn't run in browsers yet.\n\nIt is built in HTML, but right now it runs as a desktop app so you can use it to edit local files. Please use the application shell in the following repo to run this application:\n\ngithub.com/adobe/brackets-shell"), window.onerror = previousErrorHandler;
            else if (previousErrorHandler)
                return previousErrorHandler(message, url, line)
        }
        window.onerror = handleError
    }
}();
```
It's because that final `if` case gets expanded to hold all the rest of the code.

peterflynn:
This seems like a bug in the minifier then -- if it can't reliably transform strict-complying code without sometimes breaking strict mode.  Since we use nested functions tons of places in the Brackets codebase, we could easily hit the same bug in other parts of the codebase without warning...

* Is there a bug filed on uglify already about this issue?  It'd be good to have an upstream bug reference with a TODO about restoring the code to its cleaner original format after that bug is fixed.
* Is there an uglify flag that can turn off this particular behavior?  That seems safer than fixing the one case that currently breaks and leaving it open to breaking again in other parts of our code anytime potentially innocuous changes are made.

humphd:
I looked at it a bit further, and it's not a bug in uglify.  It's happening because I'm explicitly turning off function hoisting in order to get the extensions to work properly (it seems that the text plugin loading breaks if that's on).  So it's whack a mole here :(

PR #10690, opened 2015-03-04
Update README.md

coliff:
'Brackets may have reached version 1, but we're not stopping there'.

Thanks @le717 for the change suggestion

RE: #10672


prksingh:
Looks good @coliff. Thanks. Merging.

PR #10702, opened 2015-03-05
Change "Find All and Select" shortcut to Ctrl-Alt-A

PlanetVaster:
Fixed bug #8243

### DIFF
--- a/src/base-config/keyboard.json
+++ b/src/base-config/keyboard.json
@@ -1,4 +1,4 @@
-﻿{
+{
     "file.newDoc":  [
         "Ctrl-N"
     ],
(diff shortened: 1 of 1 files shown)


PlanetVaster:
Hello?

peterflynn:
Changed to more descriptive title

peterflynn:
Per the thread in #8243, I think it would be a good idea to start a discussion on brackets-dev before changing this.  There seems to be little consensus yet.  By my reading of the thread there were 2 votes for Ctrl-F3, 1 vote for Ctrl-Alt-A, and 3 votes for Ctrl-Shift-A (plus 2 more votes as a 2nd choice after Ctrl-F3).

There's also still the issue that both other major desktop editors with this feature (Sublime and Atom) use the Alt-F4 shortcut.  People always complain whenever our shortcuts differ from the established standard...  It's especially delicate in this case since we've been shipping with the existing shortcut for almost a year.

PR #10708, opened 2015-03-06
Fix bug #10617 (File drop icon only shown when outside editor)

peterflynn:
Fix bug #10617 -- Capture file drag/drop events occurring over CodeMirror before its handlers touch them, to ensure we see them. For events not over CM, use a regular bubble listener so better-behaved UI (e.g. upload drop zones) _can_ preempt the default file-open behavior.

Also, move these drag event handlers into DragAndDrop to reduce clutter.

### DIFF
--- a/src/brackets.js
+++ b/src/brackets.js
@@ -370,36 +370,10 @@ define(function (require, exports, module) {
         
         // Update title
         $("title").text(brackets.config.app_title);
-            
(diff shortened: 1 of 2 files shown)


prksingh:
Verified on Windows and Mavericks

peterflynn:
Thanks!

PR #10716, opened 2015-03-07
Attempt to make the Recent Projects unit tests more reliable

peterflynn:
Attempt to make the Recent Projects unit tests more reliable on the build machine -- instead of waiting a fixed 100ms and then asserting dropdown is visible, just wait until dropdown is visible (test will fail after the default 1000ms timeout if it's still not visible by then).

Reviewing the code, it looks like the dropdown should _synchronously_ show as visible -- it's added to the DOM and afaict is `display: block` right away.  But if I add logging code to the test, `.is(":visible")` often (though not always) remains false until a short delay... and computed style shows `display: none` too!!  That only seems to happen in the test harness though - anybody's guess as to why or how...  Maybe some weird interaction between cross-window scripting and the CSS animation the popup uses?

This seems worth a shot to reduce false-positive failures, at least...


peterflynn:
Added a second commit here to remove a bunch of duplicated JS Code Hints tests that I noticed -- I diffed the two code snippets to confirm that they really were effectively identical.

_One_ of the JS Code Hints tests that's been failing recently is in this group, so perhaps reducing it down to only one copy will help the stability there too...

prafulVaishnav:
@peterflynn Merging the PR. I will try to find out the reason for the above problem.

Issue #10325, opened 2015-01-07
Suggestions - Drag and Drop text, line highlights, and drop target split view panes


OffLuffy:
I'm new to this program, so sorry if there's a way to do something already that I've suggested!

---
##### Drag & Drop Text:
The ability to drag and drop selected text to move it around easier

---
##### Active Line Highlighting:
The current line (lines if more than one row is selected) have a different background color. If another pane in split view is active, the active line on the other panes should remain highlighted to act as somewhat of a place marker if you're referring to the other pane frequently without making it the active pane.

---
##### Drop Target for Split View Panes
The ability to drag and drop files (or file names in the open files pane into panes) to open it in that pane

---
Great work so far! Winning my favor over N++ even without above suggestions, HTML files or not.

OffLuffy:
An a latent note, another suggestion (which I never see employed in any text editor, so maybe a reason that it isn't) is to allow the use of `Shift` + `Arrow keys` to modify the bounds of block-selections (`alt` + dragging to select text) instead of having it revert to a normal selection mode. This would feel more natural after making a block-selection.

peterflynn:
**Drag/drop text:** Please upvote [Drag and Drop to move selected text within editor](https://trello.com/c/io24Byhm/910-drag-and-drop-to-move-selected-text-within-editor) on our feature backlog.

**Active Line Highlight for inactive pane/window:** Makes sense. It should be possible to do this in a theme, but it seems reasonable to do it in core too.  @larz0, thoughts?  We'd probably want a dimmer active line highlight when inactive (that's what other editors that don't hide the highlight when inactive do).

**Drop target for split panes:** It should work this way already when dragging in a file from the OS folder view. Please let us know if not.  For files that are already open, you don't have to drag them all the way to the other pane -- you can just drag them between sections in the open files list in the upper-left.

**Keys to modify block selection:** Sounds like a good candidate for someone to write an extension -- so adding the 'extension idea' tag for that.

Note: In the future, it would be good to file unrelated things as separate bugs.  But also please check the backlog to see if it's already tracked there.

redmunds:
Drag/drop text is [actively being worked on](https://github.com/adobe/brackets/pull/9584) (and is close to completion), so no need to vote!

aringot:
Anyone working on the active line highlight ? I would like to do it

larz0:
@peterflynn we're using 0.7 opacity to dim the entire inactive pane so using the same active line color could be okay.

[...]

PR #10737, opened 2015-03-13
Fix highlighting 10325

aringot:
Fix for: https://github.com/adobe/brackets/issues/10325
Also discussed here : https://groups.google.com/forum/#!topic/brackets-dev/SPokrvCyh9Y

The goal of my fix is to add (actually keep) a highlight on the (now) inactive-pane focused line when you use several panes and have the "Highlight Active Line" option checked.

Here is how it looks like with the light theme :
image

With the dark theme :
image

Only the currently focused line has a line number colored differently to clearly see the difference between active and inactive pane. (and the highlight is a bit dimmer for the inactive pane)

As it is just about styles, I didn't modify any JavaScript, I only modified LESS files.
I tested that the modification was effective both on light and dark theme, for normal and inline editors.


abose:
@aringot I coudnt get this to work. I just fetched the pull request and reloaded brackets, but the active line hilights are not coming. should i do something more?

prksingh:
@abose View > Highlight Active Line

prksingh:
@aringot For inactive pane in general line number appears highlighted. But for *Nested* inactive pane, line numbers have no highlight. Was this intentional?

ficristo:
@aringot ping

aringot:
Sorry guys, I've been pretty busy...
@prksingh , given my commit messages I got highlight even in the Nested editor. I'll try to have a look at it but I changed my laptop and OS, so I'll need to set everything up again.

PR #10742, opened 2015-03-14
Change 'split selection into lines' shortcut on Linux

Projjol-zz:
Current combination is a system default for locking the machine/ standby. New combination follows convention used by other editors and should work without affecting any system defaults.

### DIFF
--- a/src/base-config/keyboard.json
+++ b/src/base-config/keyboard.json
@@ -71,7 +71,13 @@
         }
     ],
     "edit.splitSelIntoLines": [
-        "Ctrl-Alt-L"
+        {
+           "key": "Ctrl-Alt-L"
+        },
+        {
+           "key":"Ctrl-Shift-L",
+           "platform": "linux"
+        }
     ],
     "edit.addCursorToPrevLine": [
         {


peterflynn:
Changed to less ambiguous title

src/base-config/keyboard.json
```diff
@@ -71,7 +71,13 @@
         }
     ],
     "edit.splitSelIntoLines": [
-        "Ctrl-Alt-L"
+        {
+           "key": "Ctrl-Alt-L"
```
abose: Can we use Ctrl+shift+L in all platforms so that it works uniformly?

prafulVaishnav: @abose I think we should not change the shortcut key without any strong reasons because users are accustomed to the shortcut keys in Brackets and changing any will not be received gladly by users.

Projjol: @prafulVaishnav are you suggesting that it should not be changed for any platform or that it's changed for Linux while the other systems remain the same?

prksingh: @Projjol I think he meant change the shortcut only for linux, which sounds ok to me. I have seen different key bindings for the same command on different platforms in other cases as well (Moving line up / down). So this change should be fine

PR #10758, opened 2015-03-18
Increase timeout for deleting temp directory

mdewangan:
Test case - FindInFile was failing due to smaller timeout. This test case failed in some VMs.


nethip:
@mdewangan you need to sign brackets cla. The cla is at this place. http://dev.brackets.io/brackets-contributor-license-agreement.html

PR #10768, opened 2015-03-20
Removing confusing menu items 

### LINKED ISSUE #10754: Main menu bar Find > Find In / Replace In are confusing or useless
Win 8.1, Brackets 1.2. release.

Steps to reproduce:
1. Click a folder in File Tree
2. Folder collapses/uncollapses but isn't selected
3. Can't use "Find in Selected File/**Folder**" or "Replace in Selected File/**Folder**"

prksingh:
Removing confusing menu items "Find in Selected File/Folder" and "Replace in Selected File/Folder". This fixes #10754

### DIFF
--- a/src/command/Commands.js
+++ b/src/command/Commands.js
@@ -92,7 +92,6 @@ define(function (require, exports, module) {
     // FIND
     exports.CMD_FIND                    = "cmd.find";                   // FindReplace.js               _launchFind()
     exports.CMD_FIND_IN_FILES           = "cmd.findInFiles";            // FindInFilesUI.js             _showFindBar()
-    exports.CMD_FIND_IN_SELECTED        = "cmd.findInSelected";         // FindInFilesUI.js             _showFindBarForSubtree()
     exports.CMD_FIND_IN_SUBTREE         = "cmd.findInSubtree";          // FindInFilesUI.js             _showFindBarForSubtree()
     exports.CMD_FIND_NEXT               = "cmd.findNext";               // FindReplace.js               _findNext()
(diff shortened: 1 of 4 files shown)


prafulVaishnav:
@prksingh Please remove corresponding string entries too. Also, if we are removing the menu items, we should remove all corresponding functions handleing those menu items.

src/command/Commands.js
```diff
@@ -92,7 +92,6 @@ define(function (require, exports, module) {
     // FIND
     exports.CMD_FIND                    = "cmd.find";                   // FindReplace.js               _launchFind()
     exports.CMD_FIND_IN_FILES           = "cmd.findInFiles";            // FindInFilesUI.js             _showFindBar()
-    exports.CMD_FIND_IN_SELECTED        = "cmd.findInSelected";         // FindInFilesUI.js             _showFindBarForSubtree()
```
prafulVaishnav: should not we remove _showFindBarForSubtree() if it is used only through this menu item.

prksingh: `_showFindBarForSubtree` is also being used with Command `CMD_FIND_IN_SUBTREE`

prafulVaishnav: Okk.. This command must be registering somewhere so we should be removing it..

prksingh:
@prafulVaishnav Removed strings. Handler function are still needed as mentioned in the comment

src/command/Commands.js
prksingh: yes, I have made that change when I removed the strings. Check in `FindInFiles.js` and verify

prafulVaishnav:
@prksingh Thanks.. Merging..

PR #10792, opened 2015-03-25
integration of code-folding extension into brackets

thehogfather:
Integration of core code-folding extension files into brackets' default extension folder.


Florian-R:
Out of curiosity, any reason to not keep this as an extension?

prksingh:
@Florian-R There has been demand to have code folding in the default user experience for brackets. It made sense to reuse an already popular extension to provide this support in core

abose:
@thehogfather Thanks :+1: 
The Travis CL builds are failing mostly due to JSLint related errors. Could you fix the lint errors and update?

Florian-R:
@prksinghI had hoped that Brackets has kept some sort of Node philosophy, "small core, vibrant userland" but anyway thanks for your input.

thehogfather:
@abose sure thing.

ryanstewart:
@Florian-R in general, that's our philosophy, but code folding has been on our list as "core" for a very long time. Most major editors have it out of the box. It's also far and away one of the most popular extensions, so it was clear that users were missing it.

ryanstewart:
A couple of things we want to do as part of the merge:
- Double check the shorts that are used. I think they're probably fine but we should take a look at that.
- Move the preferences into our regular preferences system (brackets.json)
- Create an option to disable code folding as preference

Open question here: Do we allow separate code folding preferences per project? or just globally? or per file?

src/extensions/default/CodeFolding/DefaultSettings.js
```diff
@@ -0,0 +1,19 @@
...
+define(function (require, exports, module) {
+	"use strict";
+    
+	module.exports = {
+		minFoldSize: 2,
+		saveFoldStates: true,
+		alwaysUseIndentFold: true,
+		enableRegionFolding: true,
+		fadeFoldButtons: false,
+        maxFoldLevel: 2 // this value is only used when fold all is called
```
MiguelCastillo: NIT: indentation

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,72 @@
...
+    "use strict";
+    var PreferencesManager      = brackets.getModule("preferences/PreferencesManager"),
+        stateManager            = PreferencesManager.stateManager.getPrefixedSystem("code-folding"),
+		DefaultSettings			= require("DefaultSettings"),
+        store = {},
+		settings = {},
+        folds = "folds";
+    
+    function simplify(folds) {
+        if (!folds) { return folds; }
```
MiguelCastillo: Tweak this to 
```
if (!folds) {
    return folds;
}
```

src/extensions/default/CodeFolding/DefaultSettings.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * JSON values for default settings
+ * @author Patrick Oladimeji
+ * @date 8/23/14 15:45:35 PM
+ */
+/*jslint vars: true, plusplus: true, devel: true, nomen: true, indent: 4, maxerr: 50 */
+/*global define*/
+define(function (require, exports, module) {
```
MiguelCastillo: Files are generally mixed tabs/spaces. For consistency, we should stick to 4 spaces.

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,72 @@
...
+        var res = {}, range;
+        Object.keys(folds).map(function (line) {
```
MiguelCastillo: Why call `map` when you are not really transforming data and creating a new array form it?  A forEach would suffice.

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,72 @@
...
+            range = folds[line];
+            res[line] = Array.isArray(range) ? range : [[range.from.line, range.from.ch], [range.to.line, range.to.ch]];
+        });
+        return res;
+    }
+    
+    function inflate(folds) {
+        if (!folds) { return folds; }
```
MiguelCastillo: Expand this out to multiple lines

```
if (!folds) {
    return folds;
}
```

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,72 @@
...
+        Object.keys(folds).forEach(function (line) {
+            obj = folds[line];
+            ranges[line] = {from: {line: obj[0][0], ch: obj[0][1]}, to: {line: obj[1][0], ch: obj[1][1]}};
+        });
+        
+        return ranges;
+    }
+    
+    module.exports = {
+        get: function (id) {
```
MiguelCastillo: Common style is to make these named functions that you assign to the `module.exports`.  E.g.

``` javascript
function get(id) {
}

module.exports.get = get;
```

We probably also benefit from some jsdocs :)

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,72 @@
...
+            store = (stateManager.get(folds) || {});
+            return inflate(store[id]);
+        },
+        set: function (id, value) {
+            store[id] = simplify(value);
+            stateManager.set(folds, store);
+            stateManager.save();
+        },
+		getSetting: function (key) {
```
MiguelCastillo: Indentation here wont be a problem once they are moved from here to named functions.

src/extensions/default/CodeFolding/foldhelpers/foldcode.js
```diff
@@ -0,0 +1,217 @@
...
+    var CodeMirror = brackets.getModule("thirdparty/CodeMirror2/lib/codemirror"),
+		prefs = require("Prefs");
+
+    module.exports = function () {
+        function doFold(cm, pos, options, force) {
+            force = force || "fold";
+            if (typeof pos === "number") {
+                pos = CodeMirror.Pos(pos, 0);
+            }
+            var finder = (options && options.rangeFinder) || CodeMirror.fold.auto;
```
MiguelCastillo: instead of checking is options exists in each line that uses `options`, you can have give it a default value.

``` javascript
options = options || {};
```

MiguelCastillo:
@thehogfather  It is awesome that you are brining this into brackets!! High-5.

I did not realize how big your extension was!  But I did a quick pass and it seems that one glaring inconsistency is the mix of spaces/tabs and that lots of the function are defined right in the `module.exports`.  I am guessing it will be a couple of rounds of tweaks/refactorings to get your extension in line with the rest of brackets. :)

Really happy this is coming in.

src/extensions/default/CodeFolding/Prefs.js
thehogfather: Indeed that is a typo. Should be a forEach.

thehogfather:
@MiguelCastillo No problem. I'll work on these fixes and update. :)

MarcelGerber:
You should probably merge your translations (at least the English ones) into our main `strings.js` as other default extensions do the same.

src/extensions/default/CodeFolding/Prefs.js
le717: But how can you return `folds` if it not passed?

thehogfather: That's equivalent to returning undefined. I've updated for clarity :).

thehogfather:
I've tidied up and refactored a bit based on the suggestions (thanks all). I've moved preferences to brackets.json and merged the english strings into brackets main strings. There is also an option to disable the extension in the settings dialog (should this be on the menu instead - or as well?). 

I have translations for fi, fr, ja, nl, pt-br and ru but haven't merged yet. Shall I go ahead and merge those?

nethip:
@thehogfather Please go ahead and merge all the translations for this, barring fr and ja. The fr and ja will need to go through our translation process.

abose:
@thehogfather As i understand, the translations, the HTML dialogs and the menu entries are for the settings dialog for code folding. But as @ryanstewart pointed out in an earlier comment, Brackets doesn't set app preferences with dialogs; But customizable options are set in the preerences file and the default options are set in the 'brackets.json' file in the repo. As the code folding extension is directly integrated with brackets, i think it would be best if it confirms to brackets preferences settings. 
So i think we should
> delete all the preferences dialogs and related strings and move all configurable settings to brackets.json.
> Remove the menu entry for code folding settings
> Code folding by default will be enabled for all users. The option to disable code folding can be set as a preferences entry.
@peterflynn @ryanstewart

MarcelGerber:
It's probably easier to have a common prefix for all string names, like `CODE_FOLDING_*`.
Also, you can now remove your old nls files.

thehogfather:
@abose thanks for the clarification. I'd misunderstood @ryanstewart's reference to brackets.json. Shall update.

abose:
@thehogfather I see a console error message "Cannot assign Ctrl-Alt-X to e4b.TOGGLE_PANEL. It is already assigned to codefolding.expand"- when we run brackets+extract builds. Could we change the expand and collapse shortcuts to something like  "ctrl + shift + ["   and "ctrl + shift + ]"  (like in sublime).

thehogfather:
@abose we can certainly change the shortcut keys - although probably not to `Ctrl+Shift+[` since that clashes with `navigate.prevDoc`. May I suggest `Ctrl+Alt+[` and `Ctrl+Alt+]`?

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,127 @@
...
+        prefs.set(foldsKey, store);
+    }
+
+    /**
+        Get the code folding setting with the specified key from the store
+        @param {!string} key The key for the setting to retrieve
+        @returns {string} the setting with the specified key
+    */
+    function getSetting(key) {
+        return prefs.get(key, PreferencesManager.CURRENT_PROJECT);
```
thehogfather: @abose @ryanstewart I need some clarification on how to retrieve preferences from the correct scope. For the sake of argument, the above line is attempting to retrieve code-folding settings that have been entered at the project level (i.e., using .brackets.json in a project dir). But it appears that the values in that file are actually not being returned in this instance. In fact after a little debugging, the project scope does not appear in the scope order used in the _getContext function defined in PreferencesManager (perhaps because there is no path property in the context sent from PreferencesManager.CURRENT_PROJECT). What is the correct way to use this api?

MiguelCastillo: @thehogfather `prefs.get(key)` is supposed to return the value by first searching the project `.brackets.json`, and if the key isn't found there, it will search the use (global) `brackets.json`.  You don't need the second parameter.  But it seems like you are after something a bit different.  Can explain on what you are trying to achieve?

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,127 @@
...
+            return;
+        }
+        var res = {}, range;
+        Object.keys(folds).forEach(function (line) {
+            range = folds[line];
+            res[line] = Array.isArray(range) ? range : [[range.from.line, range.from.ch], [range.to.line, range.to.ch]];
+        });
+        return res;
+    }
+    /**
```
MiguelCastillo: add a courtesy new line right here :D

abose:
shortcuts Ctrl+Alt+[ and Ctrl+Alt+] :+1:

src/extensions/default/CodeFolding/Prefs.js
thehogfather: @MiguelCastillo thanks, that was how I thought it worked. All I am trying to do is to override any code-folding preferences set at the user or default level with one set at the project level. For instance placing `    "code-folding.enabled": false
` in .brackets.json file to disable code-folding. I couldn't get this to work consistently yesterday so I thought I'd 'force' it to look at the project scope using the second parameter [as suggested by the wiki](https://github.com/adobe/brackets/wiki/Preferences-System#looking-up-prefs). At the moment, when I put a breakpoint in developer console to debug, it works as expected. Without the breakpoint it doesn't seem to be reading the correct preference value.
Any insights or questions would be useful. I'll also have a closer look later today.
Cheers.

thehogfather: Actually all is working consistently and as expected now.

src/nls/ru/strings.js
```diff
@@ -549,21 +549,39 @@ define({
...
     // extensions/default/RecentProjects
     "CMD_TOGGLE_RECENT_PROJECTS"                : "Предыдущие проекты",
-    
+
     // extensions/default/WebPlatformDocs
-    "DOCS_MORE_LINK"                            : "Подробнее\u2026"
+    "DOCS_MORE_LINK"                            : "Подробнее\u2026",
+
+    //extensions/default/CodeFolding
+    "COLLAPSE_ALL"                  : "Свернуть все",
```
abose: As we are not using the preferences dialog, do we need these translations now? We could just use the root English string as it used only for setting the preferences variable.

src/nls/fi/strings.js
```diff
@@ -624,7 +624,25 @@ define({
     "CMD_TOGGLE_RECENT_PROJECTS"                : "Viimeisimmät projektit",
 
     // extensions/default/WebPlatformDocs
-    "DOCS_MORE_LINK"                            : "Lue lisää"
+    "DOCS_MORE_LINK"                            : "Lue lisää",
+
+    //extensions/default/CodeFolding
+    "COLLAPSE_ALL"                  : "Pienennä kaikki",
```
abose: As we are not using the preferences dialog, do we need these translations now? We could just use the root English string as it used only for setting the preferences variable.

abose:
@thehogfather How would it behave if the user had already installed the extension from the registry. I did a basic round of testing and coudnt find any issues in the case. But will this 'double loading' of the same extension source cause any problems? How do we handle this case?

abose:
The package.json file which has the details for the extension description,author,Contributors etc.. is missing. Could you put that file file back in? 
Also need to change the name to "name": "code-folding" from "name": "brackets-code-folding" - to differentiate between the default code folding extension and the code folding extension in the registry. Should the version number could be retained to get which version of the extension we integrated?

src/nls/fi/strings.js
thehogfather: @abose 5 of the translations are for the menu items under `View` i.e., 'collapse all', 'expand all', 'collapse current', 'expand current', 'collapse custom regions'. I think it is good idea to keep these menu items (as they also show the shortcut keys). If there are strong reasons why you think we should remove them then we can discuss them.

As for the other preference strings, you are absolutely right. I've only kept them to use in the options parameter (name and description) when defining preferences. So they can be taken out of the non root strings.

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,128 @@
...
+define(function (require, exports, module) {
+    "use strict";
+    var PreferencesManager      = brackets.getModule("preferences/PreferencesManager"),
+        prefs                   = PreferencesManager.getExtensionPrefs("code-folding"),
+        strings                 = brackets.getModule("strings"),
+        store                   = {},
+        foldsKey                = "folds";
+
+    //default preference values
+    prefs.definePreference("enabled", "boolean", true,
```
MarcelGerber: I don't think any other Brackets code uses the fourth parameter `options` just yet.

thehogfather:
@abose at the moment what happens is that the first version of the extension loaded takes precedence. However event handlers on the EditorManager, DocumentManager and ProjectManager may be handled twice. The only drawback I can see is  that persistence of folded states is done twice when closing the application or switching projects. In both cases the integrity of the fold states is preserved.

I can update the version on the registry so that it is only initialised if code-folding doesnt already exist (e.g. by inspecting properties on the CodeMirror object). For those who do not update the extension, the menu items under `View` will still show an entry for the `Settings Dialog`.

src/extensions/default/CodeFolding/Prefs.js
thehogfather: No I didn't think so either. But I thought if the API exists why not use it?

src/extensions/default/CodeFolding/main.js
```diff
@@ -0,0 +1,326 @@
...
+            cm.unfoldCode(cursor.line);
+        }
+    }
+
+    /**
+        Collapses all foldable regions in the current document. Folding is done up to a level 'n'
+        which is defined in the preferences. Levels refer to fold heirarchies e.g., for the following
+        code fragment, the function is level 1, the if statement is level 2 and the forEach is level 3
+
+            function sample() {
```
le717: If this (`sample()`) is dead code, can you remove it? :)

src/extensions/default/CodeFolding/main.js
```diff
@@ -0,0 +1,326 @@
...
+                    CodeMirror.commands.foldToLevel(cm, range.from.line, range.to.line);
+                }
+            } else {
+                cm.foldCode(line);
+            }
+        }
+    }
+
+    /**
+        Collapses all custom regions defined in the current editor
```
le717: I think there's a small indentation issue here.

le717:
One thing I'm noticing is lack of [JSDocs](https://github.com/adobe/brackets/wiki/Brackets-JSDocs-Guidelines). I find the FuncDocr and Reasonable Comments extensions help make that easier. :)

src/nls/fi/strings.js
abose: What does collapse custom regions do?
Also putting in all the code folding options in the top level view menu would increase the menu real estate. It would be ideal to move it into a top level 'code folding' entry in the view/edit menu(as in sublime ) and then put the folding options as submenus. The string freeze is by 6'th, so we would need to put in all strings into brackets before that.

abose:
@thehogfather Yes it would be good if the extension is aware that there could be code folding extension by default and do the loading changes accordingly. You could push in a later update to the original extension.

src/nls/fi/strings.js
thehogfather: @abose collapse custom regions closes all custom defined region blocks in the current editor. These are defined using comments of the current mode with the keywords `region & endregion` e.g.,:

    \\region
    var point = {x: 1, y: 2};
    \\endregion

With respect to menus,  as far as I know, brackets does not implement submenus yet. Hence I'm not sure how to go about creating the sublime-like sub-menu for code-folding. So did you mean a top level "code folding" entry (just like view/edit), whose menu items are the folding options?

And as for the strings, I have no more changes to make. Are you happy with the code-folding strings present in the root string? In addition, do let me know if you would like me to remove all code-folding related translations I have merged into non-root strings. These can certainly be sorted promptly.

abose: I failed to notice that brackets didn't have submenus; i guess this is the way for now.

abose:
@thehogfather Will be merging this by tomorrow :+1: Thanks for the help. Will have to write the unit tests and some other changes, but once its in master we could help with those.

thehogfather:
@abose no problem. Happy to help out anytime.

MiguelCastillo:
@thehogfather sorry i disappeared.  Let me take another pass at the code later this evening.

My take on the code folding extension in the registry is that once this is merged in there isn't really a need for the one in the registry, especially if it is **just duplicate functionality**.  In general, I am not too concerned if someone decides to install an extension for functionality that already exists natively in brackets.  However, we should be good citizens and make sure that the integrated extension does not use the same settings as the one that already exists in the registry - to avoid collision.  Currently both use `"code-folding"`, so we would probably pick another name for this integrated extension so that people with the extension already installed don't have conflicts.

Also, I am thinking that this will be announced as a feature for a future release, so lots of people will be aware that code folding is native in brackets :D

@abose default extensions don't need a package.json and the couple that have it are theme extensions because the `theme` field is needed by the extension manager.  So package.json file should *not* come in - to stay consistent with the other default extensions.  I don't believe there is a convention where default extensions need to have a name that starts with "brackets" - I don't think it is a requirement either.

abose:
@MiguelCastillo The name for the extension in the registry is 'brackets-code-folding' and the one that is integrated is 'code-folding' to prevent the conflicts as you pointed out. Since the extension is used by a large number of users, I was hoping that @thehogfather would update the extension in the registry to  consider the case where brackets already have the extension as default[detected using the extension name from package.json?]. If the extension is pulled out of the registry, wouldn't It affect the users who have the extension installed. That was why I thought to put in the package.json file in; apart from the dev attributions. Also most of the default extension (except themes) were developed inside the brackets source base itself; that may be why the package details are missing in them.

Also did you get a chance to look at the code, I was planning to integrate it today for kicking off the string translations and giving sufficient window for unit tests from our side as the 1.3 release is very close now.

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,114 @@
...
+ * This is to make it easy for the user who edits their preferences file to easily manage the potentially numerous lines of preferences generated by the persisting code-folding state.
+ * @author Patrick Oladimeji
+ * @date 3/22/14 20:39:53 PM
+ */
+/*jslint vars: true, plusplus: true, devel: true, nomen: true, indent: 4, maxerr: 50 */
+/*global define, brackets*/
+define(function (require, exports, module) {
+    "use strict";
+    var PreferencesManager      = brackets.getModule("preferences/PreferencesManager"),
+        prefs                   = PreferencesManager.getExtensionPrefs("code-folding"),
```
MiguelCastillo: Any objection to renaming `code-folding` to maybe `brackets-code-folding`? Simply to helps everyone keep settings separate.

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,114 @@
...
+    prefs.definePreference("enableRegionFolding", "boolean", true,
+                           {name: strings.ENABLE_REGION_FOLDING, description: strings.ENABLE_REGION_FOLDING});
+    prefs.definePreference("fadeFoldButtons", "boolean", false,
+                           {name: strings.FADE_FOLD_BUTTONS, description: strings.FADE_FOLD_BUTTONS_HELP});
+    prefs.definePreference("maxFoldLevel", "number", 2,
+                           {name: strings.MAX_FOLD_LEVEL, description: strings.MAX_FOLD_LEVEL_HELP});
+    prefs.definePreference("folds", "object", {});
+
+    /**
+        Simplifies the fold ranges into an array of pairs of numbers.
```
MiguelCastillo: @thehogfather - do you think we can get all the jsdocs to follow the conventions followed in Brackets?

```
/**
  *  Simplifies the fold ranges into an array of pairs of numbers.
  *
  * @param {!{number: {from: {ch, line}, to: {ch, line}} folds the raw fold ranges indexed by line numbers
  * @returns {number: Array<Array<number>>} an object whose keys are line numbers and the values are array
  *        of two 2-element arrays. First array contains [from.line, from.ch] and the second contains [to.line, to.ch]
  */
```

[This](https://github.com/adobe/brackets/blob/master/src/extensions/default/JavaScriptCodeHints/main.js#L122) is a example.

src/extensions/default/CodeFolding/Prefs.js
thehogfather: in principle no objection. But I would suggest that I change the prefs in the version in the registry to "brackets-code-folding" for consistency - to match the name in the package.json.

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,114 @@
...
+
+        return ranges;
+    }
+
+    /**
+        Gets the line folds saved for the specified path.
+        @param {string} path the document path
+        @returns {number: {from: {line, ch}, to: {line, ch}}} the line folds for the document at the specified path
+    */
+    function get(path) {
```
MiguelCastillo: Could we rename this `get` and `set` to `getFolds` and `setFolds`?  It improves intent and readability. :)

src/extensions/default/CodeFolding/Prefs.js
thehogfather: @MiguelCastillo not a problem. Thanks for the pointer.

src/extensions/default/CodeFolding/Prefs.js
abose: That would be great. It would also help existing users keep saved settings from the extension.

src/extensions/default/CodeFolding/Prefs.js
MiguelCastillo: I like the extra meta data in there :)

src/extensions/default/CodeFolding/Prefs.js
thehogfather: Sure - I agree.

src/extensions/default/CodeFolding/foldhelpers/foldcode.js
```diff
@@ -0,0 +1,226 @@
...
+                pos = CodeMirror.Pos(pos.line - 1, 0);
+                range = getRange(false);
+            }
+        }
+        if (!range || range.cleared || force === "unfold" || range.to.line - range.from.line < minSize) {
+            if (range) { range.cleared = false; }
+            return;
+        }
+
+        var myWidget = makeWidget();
```
MiguelCastillo: Convention in Brackets is that variables are declared/defined at the beginning of the method scoping them.

src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
```diff
@@ -0,0 +1,308 @@
...
+     @param {!number} from the starting line for the update
+     @param {!number} to the ending line for the update
+    */
+    function updateFoldInfo(cm, from, to) {
+        var minFoldSize = prefs.getSetting("minFoldSize") || 2;
+        var opts = cm.state.foldGutter.options;
+        var fade = prefs.getSetting("fadeFoldButtons");
+        var gutter = $(cm.getGutterElement());
+        var i = from;
+        var isFold = function (m) {
```
MiguelCastillo: Can we expand `isFold` and `clear` function expressions to function declarations?  I havent spotted more, but function expressions aren't really used in Brackets much.

src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
```diff
@@ -0,0 +1,308 @@
...
+            return m.__isFold;
+        }, clear = function (m) {return m.clear(); };
+
+        /**
+            helper function to check if the given line is in a folded region in the editor.
+            @param {Number} line the
+            @return {from: {ch: Number, line: Number}, to: {ch: Number, line: Number}} the range that hides the specified line or undefine if the line is not hidden
+        */
+        function _isCurrentlyFolded(line) {
```
MiguelCastillo: Functions that start with `_` are conventionally private...  Can we add `@private` to the jsdoc?

src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
```diff
@@ -0,0 +1,308 @@
...
+            var keys = Object.keys(cm._lineFolds), i = 0, r;
```
MiguelCastillo: Can we please name `r` to a little more descriptive name?  It helps us mere humans read the code a bit better :)

src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
```diff
@@ -0,0 +1,308 @@
...
+                }
+                cm.setGutterMarker(i, opts.gutter, mark);
+                i++;
+            }
+        }
+    }
+
+    function updateInViewport(cm, from, to) {
+        var vp = cm.getViewport(), state = cm.state.foldGutter;
+        from = !isNaN(from) ? from : vp.from;
```
MiguelCastillo: NIT: double negation tends to be harder to read.

``` javascript
from = isNaN(from) ? vp.from : from;
to = isNan(to) ? vp.to : to;
```

src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
```diff
@@ -0,0 +1,308 @@
...
+            var state = cm.state.foldGutter;
+            var lineChanges = changeObj.text.length - changeObj.removed.length;
+            //update the lineFolds cache
+            updateFoldsCache(cm, changeObj.from.line, lineChanges);
+            if (lineChanges !== 0) {
+                updateFoldInfo(cm, changeObj.from.line + lineChanges, changeObj.from.line + lineChanges + 1);
+            }
+            state.from = changeObj.from.line;
+            state.to = 0;
+            window.clearTimeout(state.changeUpdate);
```
MiguelCastillo: If we use `lodash` we can clean this setting/clearing timeout with the debounce function.  Very nifty little function. https://lodash.com/docs#debounce

src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
```diff
@@ -0,0 +1,308 @@
...
+        }
+    }
+
+    /**
+        Triggered on viewport changes e.g., user scrolls or resizes the viewport.
+        @param {!CodeMirror} cm the CodeMirror instance for the active editor
+    */
+    function onViewportChange(cm) {
+        var state = cm.state.foldGutter;
+        window.clearTimeout(state.changeUpdate);
```
MiguelCastillo: Same `debounce` suggestion as above

src/extensions/default/CodeFolding/main.js
```diff
@@ -0,0 +1,326 @@
...
+    var CodeMirror              = brackets.getModule("thirdparty/CodeMirror2/lib/codemirror"),
+        Strings                 = brackets.getModule("strings"),
+        AppInit                 = brackets.getModule("utils/AppInit"),
+        CommandManager          = brackets.getModule("command/CommandManager"),
+        DocumentManager         = brackets.getModule("document/DocumentManager"),
+        EditorManager           = brackets.getModule("editor/EditorManager"),
+        ProjectManager          = brackets.getModule("project/ProjectManager"),
+        KeyBindingManager       = brackets.getModule("command/KeyBindingManager"),
+        ExtensionUtils          = brackets.getModule("utils/ExtensionUtils"),
+        Menus					= brackets.getModule("command/Menus"),
```
MiguelCastillo: NIT: A little indentation detail

src/extensions/default/CodeFolding/main.js
```diff
@@ -0,0 +1,326 @@
...
+
+    //load code mirror addons
+    brackets.getModule(["thirdparty/CodeMirror2/addon/fold/brace-fold"]);
+    brackets.getModule(["thirdparty/CodeMirror2/addon/fold/comment-fold"]);
+    brackets.getModule(["thirdparty/CodeMirror2/addon/fold/markdown-fold"]);
+
+    //still using slightly modified versions of the foldcode.js and foldgutter.js since we
+    //need to modify the gutter click handler to take care of some collapse and expand features
+    //e.g. collapsing all children when 'alt' key is pressed
+    var foldGutter              = require("foldhelpers/foldgutter"),
```
MiguelCastillo: Move these require statements along with the other one in line 51.

src/extensions/default/CodeFolding/main.js
```diff
@@ -0,0 +1,326 @@
...
+
+    function saveBeforeClose() {
+        saveLineFolds(EditorManager.getCurrentFullEditor());
+    }
+
+    /**
+        Initialise the extension
+    */
+    function init() {
+        if (!CodeMirror.fold.combine && _prefs.getSetting("enabled")) {
```
MiguelCastillo: In the name of readability, can we make this `if` statement be an early `return`?  This is merely to get rid of the deep indentation in the body of the `if` statement.

src/extensions/default/CodeFolding/main.js
```diff
@@ -0,0 +1,326 @@
...
+        Strings                 = brackets.getModule("strings"),
+        AppInit                 = brackets.getModule("utils/AppInit"),
+        CommandManager          = brackets.getModule("command/CommandManager"),
+        DocumentManager         = brackets.getModule("document/DocumentManager"),
+        EditorManager           = brackets.getModule("editor/EditorManager"),
+        ProjectManager          = brackets.getModule("project/ProjectManager"),
+        KeyBindingManager       = brackets.getModule("command/KeyBindingManager"),
+        ExtensionUtils          = brackets.getModule("utils/ExtensionUtils"),
+        Menus					= brackets.getModule("command/Menus"),
+        _prefs                  = require("Prefs"),
```
MiguelCastillo: Why underscore this `_prefs` but not in an other file/module?

src/extensions/default/CodeFolding/main.js
```diff
@@ -0,0 +1,326 @@
...
+        var editor = EditorManager.getFocusedEditor();
+        if (editor && editor._codeMirror) {
+            var cm = editor._codeMirror;
+            CodeMirror.commands.unfoldAll(cm);
+        }
+    }
+
+    function createGutter(editor) {
+        var cm = editor._codeMirror;
+        if (cm) {
```
MiguelCastillo: Can we also do an early `return` here?

src/extensions/default/CodeFolding/Prefs.js
MiguelCastillo: @thehogfather Are you proposing that the integrated and the one in the registry should have the same preferences name?

thehogfather: @MiguelCastillo no. Just proposing that the preference name should match the name in the package.json. So currently the version in the registry stores preferences using the "code-folding" key. I'll update it to "brackets-code-folding" and leave the version integrated into brackets as "code-folding". This also means any preferences saved in the old extension will be kept in the integrated version.

MiguelCastillo: Gotcha.  That makes sense.  I was just curious if we wanted users to just transparently use the settings they have.  In which case I expect them to remove the extension.

MiguelCastillo:
@thehogfather this code is definitely landable right now and it is looking very good!

There are a few changes that seem manageable to make the code fit better with brackets style.  I think we are almost there :)

thehogfather:
@MiguelCastillo just pushed an update for all those fixes + merged with the latest commit from master.

abose:
@thehogfather  Thanks for the help. Merged into master.
Any suggestion on the unit tests that needs to be added?

MiguelCastillo:
@thehogfather yeah man, your last changes look very good. Thank you, this is a great addition to brackets!  Now you gotta come up with another really cool extension :)

@abose I really like the idea of adding the package.json for attribution :)

le717:
Dancing banana

Congrats on the merge, @thehogfather! :smile:

thehogfather:
:+1: thanks all. Glad I could help.

@abose I guess I would mainly check that folded states are persisted and restored correctly, and that the preference options  (enabled, minFoldSize etc.) work as intended.

src/nls/fi/strings.js
peterflynn: @abose See https://trello.com/c/EwLGRkYe/541-native-submenus.  The only reason we don't have it is that it's a sizeable chunk of work: we'd have to decide on the JS `Menus` API, implement submenus natively on Mac and Win, _and_ implement them HTML for Linux / in-browser before the feature could land in master.

src/extensions/default/CodeFolding/foldhelpers/foldcode.js
peterflynn: @MiguelCastillo Fwiw, not necessarily true. Because of JSLint, it's safe (often actually saf<i>er</i>) to declare vars further down in a function at the point when they're initialized.  We do have plenty of code in core where vars are not all grouped at the top of the scope.

src/extensions/default/CodeFolding/foldhelpers/foldcode.js
```diff
@@ -0,0 +1,268 @@
+/**
+ * Based on http://codemirror.net/addon/fold/foldcode.js
+ * @author Patrick Oladimeji
```
peterflynn: I think we still need to keep the original copyright from CodeMirror here

src/extensions/default/CodeFolding/foldhelpers/foldcode.js
MiguelCastillo: Hmmm, the only reason I say this is because JSLint would complain if your vars werent declared at the top of your functions.  I am guessing this changed.  Good to know though. :)

peterflynn: We use the `vars` JSLint option, which lets you declare vars anywhere in the scope -- as long as they're not duplicated or referenced before the declaration.  It's sort of a best of both worlds, because it effectively makes var declarations work more like Java or C++ (assuming you don't ignore lint errors).

src/extensions/default/CodeFolding/main.js
```diff
@@ -0,0 +1,352 @@
...
+/*jslint vars: true, plusplus: true, devel: true, nomen: true, indent: 4, maxerr: 50 */
+/*global define, require, $, brackets*/
+
+require.config({
+    paths: {
+        "text" : "lib/text",
+        "i18n" : "lib/i18n"
+    },
+    locale: brackets.getLocale()
+});
```
peterflynn: This should be unnecessary now that this is a part of core

src/extensions/default/CodeFolding/strings.js
```diff
@@ -0,0 +1,38 @@
...
+ * to display strings should should load this module by calling var Strings = require("strings").
+ * The i18n plugin will dynamically load the strings for the right locale and populate
+ * the exports variable. See src\nls\strings.js for the master file of English strings.
+ */
+define(function (require, exports, module) {
+    "use strict";
+
+    module.exports = require("i18n!nls/strings");
+
+});
```
peterflynn: Ditto -- I think this whole file is no longer used

src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
peterflynn: @MiguelCastillo Though I wonder if it'd be better to try to keep this as close as possible to the original CM code, with the goal of getting our few changes merged back upstream so we don't have to maintain a fork of it...

src/extensions/default/CodeFolding/foldhelpers/latex-fold.js
```diff
@@ -0,0 +1,59 @@
+/**
+ * Fold latex code regions
```
peterflynn: I wonder if a better approach would be to make this more extensible -- an important goal of the LanguageManager work is to make it so any feature in Brackets can be enabled for additional languages just by installing an extension.  Latex and the custom regions feature both seem like potential candidates to move out into extensions.

Custom regions also adds a menu item that I suspect few users will understand, and that's concerning to me.  Since refactoring this to be more extensible is probably a post-1.3 thing, I wonder if it'd be better to just disable the custom regions feature for now?  Or have the menu item hidden by default, and accessible only by setting a preference?

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,116 @@
...
+        return inflate(store[path]);
+    }
+
+    /**
+      * Saves the line folds for the specified path
+      * @param {!string} path the path to the document
+      * @param {Object} folds the fold ranges to save for the current document
+      */
+    function setFolds(path, folds) {
+        store[path] = simplify(folds);
```
peterflynn: This _really_ should be in view state and not the main preferences file -- it's exactly the kind of thing view state was intended for.  This seems like something we should fix for 1.3 lest we start polluting everyone's preferences file (which should be easily human-readable) with folding state.

peterflynn: Filed as #10846 to make sure we track this

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,116 @@
...
+      */
+    function getSetting(key) {
+        return prefs.get(key);
+    }
+
+    /**
+      * Clears all the saved line folds for all documents.
+      */
+    function clearAllFolds() {
+        prefs.set(foldsKey, {});
```
peterflynn: This function appears to be unused, but I think it also has a bug: this doesn't reset the `store` var used elsewhere, so any call to `setFolds()` later seems like it would immediately restore _all_ the folds that had been cleared previously.

It's probably safest just to re-get the pref before setting it each time.  There may be other bugs lurking with the current approach too (e.g. if you edit the prefs file... does it blow those changes away?)

src/extensions/default/CodeFolding/Prefs.js
abose: @thehogfather #10846 Instead of using the prefs , could we migrate the folding state to state.json?

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,116 @@
...
+                           {name: strings.ENABLE_CODE_FOLDING, description: strings.ENABLE_CODE_FOLDING});
+    prefs.definePreference("minFoldSize", "number", 2,
+                           {name: strings.MIN_FOLD_SIZE, description: strings.MIN_FOLD_SIZE_HELP});
+    prefs.definePreference("saveFoldStates", "boolean", true,
+                           {name: strings.SAVE_FOLD_STATES, description: strings.SAVE_FOLD_STATES_HELP});
+    prefs.definePreference("alwaysUseIndentFold", "boolean", true,
+                           {name: strings.ALWAYS_USE_INDENT_FOLD, description: strings.ALWAYS_USE_INDENT_FOLD_HELP});
+    prefs.definePreference("enableRegionFolding", "boolean", true,
+                           {name: strings.ENABLE_REGION_FOLDING, description: strings.ENABLE_REGION_FOLDING});
+    prefs.definePreference("fadeFoldButtons", "boolean", false,
```
peterflynn: I wonder if this preference should be on by default.

I can't actually get it to work though -- when I set it to true, it seems to have no effect (at least on JS or LESS files).  Will file a bug.

peterflynn: Ok, false alarm with the preferences issue.

I wonder if the best UX is actually in between this setting (hidden until mouseover) and the default (always visible) -- having the triangles be fainter until you mouse over the gutter, then getting darker.  I may mock that up and see how it feels...

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -0,0 +1,116 @@
...
+
+    //default preference values
+    prefs.definePreference("enabled", "boolean", true,
+                           {name: strings.ENABLE_CODE_FOLDING, description: strings.ENABLE_CODE_FOLDING});
+    prefs.definePreference("minFoldSize", "number", 2,
+                           {name: strings.MIN_FOLD_SIZE, description: strings.MIN_FOLD_SIZE_HELP});
+    prefs.definePreference("saveFoldStates", "boolean", true,
+                           {name: strings.SAVE_FOLD_STATES, description: strings.SAVE_FOLD_STATES_HELP});
+    prefs.definePreference("alwaysUseIndentFold", "boolean", true,
+                           {name: strings.ALWAYS_USE_INDENT_FOLD, description: strings.ALWAYS_USE_INDENT_FOLD_HELP});
```
peterflynn: What does it mean to have this enabled?  What is the behavior e.g. in a JS file if this conflicts with the standard brace-fold/comment-fold helpers?

src/extensions/default/CodeFolding/foldhelpers/latex-fold.js
thehogfather: @peterflynn I agree - and I've already tested moving the latex-fold function into the latex extension and that is works as expected. They can be safely removed from the main.js file (if you haven't sealed changes to 1.3 yet).
I also think the custom region should be disabled by default (and the menu option hidden).

src/extensions/default/CodeFolding/Prefs.js
thehogfather: @abose yes - for the folds. I suspect the other settings would need to be in the regular prefs (brackets.json).

src/extensions/default/CodeFolding/Prefs.js
thehogfather: @peterflynn yes editing the prefs file for documents that are currently open in brackets (has no effect - the changes are restored one the document becomes the active editor). This is a bug - that  your safe approach should fix.

src/extensions/default/CodeFolding/Prefs.js
thehogfather: It means that line indentation is used as a suggestion for fold regions. Although this shouldn't take priority over range finders based on the current mode. So where there is conflict, the brace-fold-finder should take precedence. Having said that this might be another pref that could be better off disabled by default.

src/extensions/default/CodeFolding/main.js
```diff
@@ -0,0 +1,352 @@
...
+      */
+    function saveBeforeClose() {
+        saveLineFolds(EditorManager.getCurrentFullEditor());
+    }
+
+    /**
+        Initialise the extension
+    */
+    function init() {
+        if (CodeMirror.fold.combine || !prefs.getSetting("enabled")) {
```
peterflynn: When would `CodeMirror.fold.combine` be true?

thehogfather: It would be truthy if the code-folding plugin is already loaded. This is mainly  for the extension in the registry- to prevent registering key bindings twice etc. I could not find an api to check that an extension has loaded. That code fragment can be removed without affecting the logic of initialisation.

src/extensions/default/CodeFolding/Prefs.js
abose: @thehogfather could use PreferencesManager.getViewState("codefolding.states"); and PreferencesManager.setViewState("codefolding.states", state) to put to the state.json file.

src/nls/fi/strings.js
abose: OK, got it. we'll have to address it some day.

abose:
@thehogfather I was verifying the unit tests and some of the integration tests in the unit test setup is failing. 
Most is with this message- Cannot read property 'lastIndexOf' of undefined 
Could you take a look at debug>run tests>integration>all .

abose:
Code folding is enabled in all files by default, and in some files like txt files it looks a bit odd.
Would it be a good thing to enable code folding in only HTML,XML,js Etc. and disabled in all other file types?

thehogfather:
@abose wrt unit tests, I am unable to run them completely. They freeze with about 125 tests to go. Can you tell me the file throwing the error or show a stack trace? I remember spotting one a while back in one of the imported codeMirror range finders.

src/extensions/default/CodeFolding/main.js
peterflynn: Ah, in that case -- core extensions are always loaded before user-installed extensions, so I think this logic is only needed in the registry copy, not here.

abose:
There is a memory issue with the unit tests currently, so we won't be able to run all tests in a single go.
So in order to run it, we have to the tests in individual sections.
First restart brackets; then open test suite, click on integration tab, click on all; it would start running integration tests. If brackets slows down, Restart brackets again to run tests.
In integration test, specifically the following test are failing[ FindInFiles(3 test cases),FindReplace - Integration(1),	InlineEditorProviders(3), QuickOpen(3)]. 
I am checking out these errors and the stack.

But Issue#10857 seems to be much more important than this. My understanding of the extension is limited; it might take a while for me to fix it. Could you take a look at the issue?

peterflynn:
@abose I took a quick look at those unit tests and it seems like the Health Report notification dialog might be causing problems in some of them (you can see it popping up in each test run, even though there's code that should suppress it).

The 3 QuickOpen failures might be different -- I can repro them on my old smart-autocompelte branch (which doesn't have code folding _or_ Health Report), so those failures might be caused by my recent PR there.  I can try to take a look tomorrow -- lmk if you find anything out before then.

abose:
The health notifications dialogs will be disabled when unit tests are nunning, but i guess that PR is still under review.
For the time being, it can be disabled by going to HealthDataNotification.js > AppInit.appReady()> 
params.parse();if (params.get("testEnvironment")) return;

redmunds:
There's a wiki page with [Tips and Tricks for Running Brackets Unit Tests](https://github.com/adobe/brackets/wiki/Running-Brackets-Unit-Tests) so just add any new tips there and share this link.

Pooria-H:
Why there is no option to disable this feature? It's ugly.

le717:
@Pooria-H There is an option. In your `brackets.json` file, set `code-folding.enabled` to `false`. You can get more details on the [wiki](https://github.com/adobe/brackets/wiki/How-to-Use-Brackets#preferences).

ryanstewart:
Thanks @le717 !

Pooria-H:
@le717 Oh I didn't noticed. thanks

PR #10813, opened 2015-03-31
Update React library to latest 0.13.1

zaggino:
I'd like to use React when writing extensions, but the one bundled with Brackets is quite outdated. Including another version of React into Brackets introduces conflicts between the two so that's not a good way to go. I have gone through API renames and breaking changes between 0.11 (current) and 0.13 (latest) so this should be pretty safe to upgrade.

### DIFF
--- a/src/project/FileTreeView.js
+++ b/src/project/FileTreeView.js
@@ -33,6 +33,7 @@ define(function (require, exports, module) {
     "use strict";
 
     var React             = require("thirdparty/react"),
+        Classnames        = require("thirdparty/classnames"),
         Immutable         = require("thirdparty/immutable"),
         _                 = require("thirdparty/lodash"),
         FileUtils         = require("file/FileUtils"),
@@ -448,7 +449,7 @@ define(function (require, exports, module) {
             }
 
             var nameDisplay,
-                cx = React.addons.classSet;
+                cx = Classnames;
 
             var fileClasses = cx({
                 'jstree-clicked selected-node': this.props.entry.get("selected"),
--- a/src/project/FileTreeView.js
+++ b/src/project/FileTreeView.js
@@ -120,9 +120,8 @@ define(function (require, exports, module) {
         handleClick: function (e) {
             e.stopPropagation();
             if (e.button !== LEFT_MOUSE_BUTTON) {
-                return false;
+                e.preventDefault();
             }
-            return true;
         },
 
         /**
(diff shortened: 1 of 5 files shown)


peterflynn:
The project tree is awfully fiddly -- e.g. look at how many bugs fell out of the original React migration... I think > 50 despite very careful coding work and lots of unit tests.  Given that, I think we'd want to carve out some significant testing time (beyond just code review) before landing this.  That might potentially wind up pushing it into 1.4.

zaggino:
Yes, agreed Peter. Ideally it be merged soon after 1.3 is out. I'd be willing to rewrite this into JSX to make it more readable and easier to fix future bugs but I didn't want to do that in this PR.

ryanstewart:
I went ahead and added this to a new Release 1.4 milestone. Trying to merge it early after we lock and load 1.3 seems ideal.

nethip:
@ryanstewart Sure! we should get this in in 1.4 . Thanks for jumping into this and prioritizing :smile:

zaggino:
This should be now good to test/merge as 1.3 is already out.

ingorichter:
@zaggino I think we should update `immutable.js` too. In the past we updated both dependencies in one PR.

zaggino:
Ok, I'll update that too and commit to this.

nethip:
@zaggino Also please run unit tests and update this thread with your results? Thanks!

zaggino:
@ingorichter `immutable.js` updated to the latest release `3.7.2` from https://github.com/facebook/immutable-js/releases + updated react to `0.13.2` from https://fb.me/react-with-addons-0.13.2.js
@nethip all tests passing fine
image

MarcelGerber:
You kind of agreed on merging this "shortly after 1.3 is out". So shouldn't we go ahead merge this now?

MiguelCastillo:
The changes look good.  I am going load them up and test it for the next day or so.

MiguelCastillo:
BTW, I am in favor of moving to JSX...  But linting will probably need to some attention.

zaggino:
@MiguelCastillo I'd be in favor of switching to http://eslint.org/ ; I've been using it at work and it works perfect on ES6 and JSX; also is very actively developed with new rules

I'd be great if one of you could test this and merge it.

MiguelCastillo:
@zaggino yeah eslint is what I am using a few of my projects.  It has a couple of bugs around properly detecting react component that are defined and then rendered at tags via React.render.

E.g

```
var Component = React.createClass({
...
});

React.render(<Component/>, ...);
```

That generate a warning that `Component` is define but never used...  Other than that, eslint has solid support for JSX.  But as I mentioned, this will be a migration of the linting system, which I think would be a good thing :)

I am currently running/testing your react update branch...  If I don't find anything by tomorrow, I will report back and we should probably be ok merging.

zaggino:
Thanks @MiguelCastillo

MiguelCastillo:
I haven't had any issues with your changes, @zaggino.  I am going to merge this in now and get more testing milage early on.

PR #10820, opened 2015-04-02
Use undefined instread of void. Fixes JSLint.

busykai:
Simple fix to avoid JSLint complaining about JSHint workaround (sic!).


busykai:
Closed. It's actually makes the JSLint to shut up, but causes runtime error...

PR #10830, opened 2015-04-05
Install Launch Script changes(MAC only) for installing Brackets launch script at /usr/local/bin

nethip:
Install Launch Script changes: This change creates a new menu entry under Debug menu, choosing which will create a launch script at /usr/local/bin folder allowing the user to launch Brackets from terminal or any comm


nethip:
This command is going to plainly write out a file to /usr/local/bin folder and make it an executable. If the folder does not exist, we will not be able to create the folder programmatically as that requires admin access. I have changed the error message, in case of failures, to be more informative.
Also part of the change I have added another api `chmod()` to `AppshellFileSystem.js`. 

@peterflynn @MarcelGerber Please review this when you get a chance. Thanks!

src/extensions/default/DebugCommands/main.js
```diff
@@ -104,7 +108,53 @@ define(function (require, exports, module) {
     function handleNewBracketsWindow() {
         window.open(window.location.href);
     }
-    
+
+    function handleScriptMessages(err) {
+        if (err) {
+            var errMsg = Strings.ERROR_CREATING_LAUNCH_SCRIPT + err;
```
MarcelGerber: Better use `StringUtils.format` here

src/nls/root/strings.js
```diff
@@ -574,6 +574,7 @@ define({
     "CMD_REFRESH_WINDOW"                        : "Reload With Extensions",
     "CMD_RELOAD_WITHOUT_USER_EXTS"              : "Reload Without Extensions",
     "CMD_NEW_BRACKETS_WINDOW"                   : "New {APP_NAME} Window",
+    "CMD_LAUNCH_SCRIPT_MAC"                     : "Install Launch Script",
```
ryanstewart: Could we change this to "Add Brackets Command Line"?

It's 20 characters versus 26 but I think it will still fit in the menu because there is extra space there.

src/nls/root/strings.js
```diff
@@ -582,6 +583,14 @@ define({
     "CMD_RESTART_NODE"                          : "Restart Node",
     "CMD_SHOW_ERRORS_IN_STATUS_BAR"             : "Show Errors in Status Bar",
     "CMD_OPEN_BRACKETS_SOURCE"                  : "Open {APP_NAME} Source",
+    "ERROR_CREATING_LAUNCH_SCRIPT"              : "An erorr occured while creating {APP_NAME} launcher script at /usr/local/bin. Please refer to https://github.com/adobe/brackets/wiki/Command-Line-Arguments for troubleshooting. Reason: ",
```
ryanstewart: What about:

`An error occurred while creating the {APP_NAME} command line tool at /usr/local/bin. Please refer to https://github.com/adobe/brackets/wiki/Command-Line-Arguments for troubleshooting. Reason: `

src/nls/root/strings.js
```diff
@@ -582,6 +583,14 @@ define({
...
+    "ERROR_CLTOOLS_RMFAILED"                    : "Failed to remove existing {APP_NAME} symlink at /usr/local/bin.",
+    "ERROR_CLTOOLS_MKDIRFAILED"                 : "Failed to create /usr/local/bin directory structure",
+    "ERROR_CLTOOLS_LNFAILED"                    : "Failed to create {APP_NAME} symlink at /usr/local/bin",
+    "ERROR_CLTOOLS_SERVFAILED"                  : "Failed to create authorization",
+    
+    "LAUNCH_SCRIPT_CREATE_SUCCESS"              : "Launch script succesfully installed! You can now launch {APP_NAME} from terminal.",
+    "CREATING_LAUNCH_SCRIPT_TITLE"              : "Add Brackets Command Line",
```
ryanstewart: How about: 

`Add Brackets Command Line Tool`

src/nls/root/strings.js
```diff
@@ -582,6 +583,14 @@ define({
     "CMD_RESTART_NODE"                          : "Restart Node",
     "CMD_SHOW_ERRORS_IN_STATUS_BAR"             : "Show Errors in Status Bar",
     "CMD_OPEN_BRACKETS_SOURCE"                  : "Open {APP_NAME} Source",
+    "ERROR_CREATING_LAUNCH_SCRIPT"              : "An erorr occured while creating {APP_NAME} launcher script at /usr/local/bin. Please refer to https://github.com/adobe/brackets/wiki/Command-Line-Arguments for troubleshooting. Reason: ",
+    "ERROR_CLTOOLS_RMFAILED"                    : "Failed to remove existing {APP_NAME} symlink at /usr/local/bin.",
+    "ERROR_CLTOOLS_MKDIRFAILED"                 : "Failed to create /usr/local/bin directory structure",
+    "ERROR_CLTOOLS_LNFAILED"                    : "Failed to create {APP_NAME} symlink at /usr/local/bin",
+    "ERROR_CLTOOLS_SERVFAILED"                  : "Failed to create authorization",
+    
+    "LAUNCH_SCRIPT_CREATE_SUCCESS"              : "Launch script succesfully installed! You can now launch {APP_NAME} from terminal.",
```
ryanstewart: And here:
`Command line tool successfully installed! You can now launch {APP_NAME} from terminal.`

src/nls/root/strings.js
```diff
@@ -582,6 +583,14 @@ define({
     "CMD_RESTART_NODE"                          : "Restart Node",
     "CMD_SHOW_ERRORS_IN_STATUS_BAR"             : "Show Errors in Status Bar",
     "CMD_OPEN_BRACKETS_SOURCE"                  : "Open {APP_NAME} Source",
+    "ERROR_CREATING_LAUNCH_SCRIPT"              : "An erorr occured while creating {APP_NAME} launcher script at /usr/local/bin. Please refer to https://github.com/adobe/brackets/wiki/Command-Line-Arguments for troubleshooting. Reason: ",
+    "ERROR_CLTOOLS_RMFAILED"                    : "Failed to remove existing {APP_NAME} symlink at /usr/local/bin.",
```
ryanstewart: These will be different for Mac versus Win, right? Do we need Windows-specific strings?

src/nls/root/strings.js
nethip: Changed the menu entry to Add Brackets Command Line.

src/nls/root/strings.js
nethip: Sure! Right now I have this as `Command Line Tool.

src/nls/root/strings.js
nethip: I have made the string change and now the dialog looks like this.

screen shot 2015-04-08 at 6 08 37 pm

Does this look fine?

src/nls/root/strings.js
nethip: This is a MAC only change. We would not need strings for Win as the command line stuff is taken care of by our installer itself. And I have made change to the new shell api which would return an error if that API is called from platforms other than MAC

nethip: Also we are creating menu entry only on MAC.

nethip:
@ryanstewart Could you check the final strings, if they look fine?

nethip:
The dialog boxes now look as the following. 

screen shot 2015-04-08 at 6 33 55 pm

Error dialog:

screen shot 2015-04-08 at 6 35 13 pm

nethip:
This requires https://github.com/adobe/brackets-shell/pull/512

MarcelGerber:
I know I'm late (at least for the string changes), but normally commands are all-lowercase (that is, `brackets` instead of `brackets`), right?

src/extensions/default/DebugCommands/main.js
```diff
@@ -105,6 +107,63 @@ define(function (require, exports, module) {
...
+            errorString = Strings.ERROR_CLTOOLS_LNFAILED;
+            break;
+        case appshell.app.ERR_CL_TOOLS_SERVFAILED:
+            errorString = Strings.ERROR_CLTOOLS_SERVFAILED;
+            break;
+        case appshell.app.ERR_CL_TOOLS_NOTSUPPORTED:
+            errorString = Strings.ERROR_CLTOOLS_NOTSUPPORTED;
+            break;
+        default:
+            errorString = "Unknown Error.";
```
prksingh: Localize this also.

nethip:
@MarcelGerber Thanks for the suggestion! I did a quick check both on windows and Mac and I am able to launch Brackets from command line just by typing `brackets`. By default all commands on Mac are case insensitive but IIRC there is a setting to override this in which case the command should be `Brackets`. On windows all commands are case insensitive in **cmd.exe** and I not sure if this can be overridden in **GitBash** though.

nethip:
@MarcelGerber There is no option on  Mac to make terminal case sensitive. Mac has HFS+ file system which is case insensitive but case preserving meaning you can have two files with the same names but with different cases.

src/extensions/default/DebugCommands/main.js
nethip: I think we should do this. But there is no generic Unknown Error string which we can use. We might want to add a string. I would suggest we do this in the next release as we have already sent the strings for translation.

prksingh:
Merging.

MarcelGerber:
What I meant is that we should better name the newly created executable `brackets` instead of `Brackets`. That's also an issue should we ever port this to Linux, which is case sensitive.

src/filesystem/impls/appshell/AppshellFileSystem.js
```diff
@@ -489,6 +489,19 @@ define(function (require, exports, module) {
...
+     * specified mode in numeric format (ie 0777)
+     * @param {string} path
+     * @param {number} mode
+     * @param {function(string)=} callback
+     */
+    function chmod(path, mode, callback) {
+        appshell.fs.chmod(path, mode, function (err) {
+            callback(_mapError(err));
+        });
+    }
```
peterflynn: This is unused, right?  It's not accessible from anywhere else in Brackets (since FileSystem doesn't expose the fs impl object), so I think we should remove it.

PR #10831, opened 2015-04-06
Making External changes Dialog Box clearer

rroshan1:
Fix for #10341 .
External changes conflict dialog is confusing/unclear. Adding "outside of Brackets" to add clarity to these type of messages.
The updated message would be:-
external_changes


peterflynn:
@rroshan1 This looks good to me, except please revert the changes to brackets-concat.js -- that is a performance testcase file that should never be updated.

Issue #9847, opened 2014-11-06
PASCAL language highlight syntax support


MOlszowski:
How about PASCAL support?

PR #10844, opened 2015-04-07
Added Haskell as a language

VenessaJohansenBarrera:
Added at the end of the languages.json file so that Haskell will have
syntax highlighting in Brackets.

### DIFF
--- a/src/language/languages.json
+++ b/src/language/languages.json
@@ -277,5 +277,13 @@
             "swf", "tar", "tif", "tiff", "ttf", "woff", "xls", "xlsx", "zip"
         ],
         "isBinary": true
+    },
+
+    "haskell": {
+        "name": "Haskell",
+        "mode": "haskell",
+        "fileExtensions": ["hs"],
+        "blockComment": ["{-", "-}"],
+        "lineComment": ["--"]
     }
 }


peterflynn:
@VenessaJohansenBarrera We need you to sign the [Brackets contributor agreement (CLA)](http://dev.brackets.io/brackets-contributor-license-agreement.html) before we can merge this.

VenessaJohansenBarrera:
The agreement has been signed.

Thanks!

abose:
validated change, merging. @VenessaJohansenBarrera Thanks!

busykai:
as Brackets is the editor for the web, Haskell should rather not be part of the core. there's a syntax highlighting extension for Haskell, readily available via registry. there are many languages which are intentionally *not* part of Brackets, but still available as extensions. why Haskell has been made an exception?

abose:
This change only brings in the file extension awareness in brackets and very basic comment block awareness for haskell, similar to the way we have basic support for c/c++/java... which are not primarily 'web' langs. Full blown support is not part of brackets as said.

busykai:
@abose, I'm sorry for being stubborn, but this deserves some more discussion. I am pretty sure I understand what this PR does (exactly what [the corresponding](https://github.com/ForbesLindesay/brackets-language-extensions) would do). We also could have support for Cobol, Pascal, Fortran etc., but we [did not add these](https://github.com/adobe/brackets/issues/9847) before. I would understand including these if there were no extensibility, but I'd much rather add support for these via extensions. Now, that's my personal preferences... I guess from the product standpoint, we just have to be consistent add [all of them](https://codemirror.net/mode/) or only those which are relevant to web development. Strong focus on web is what makes Brackets so great. Having these irrelevant languages floating around would only create false expectations on the customer side and steals time from development.

Usability-wise, we have a drop-down box to pick a language in the editor status bar. I'd rather not contaminate that list with arbitrary languages (it's long enough already). Those who need support for extra languages have extensions at their disposal.

Just my couple of cents.

abose:
I see your point @busykai , and now that i think of it , i agree with you. Haskell is still an obscure language (compared to c/c++/etc.. in our language dropdown), so supporting that is best left to the extensions as they could give in much more support for those languages than the basic support. If no one objects, i could revert the commit by wednesday.

PR #10859, opened 2015-04-09
Code folding UI & code style tweaks

peterflynn:
~~We should land PR #10850 first, and then the diff for this will be simplified.~~ _(done)_

~~For now, this link previews what the diff will be after 10850 lands: https://github.com/thehogfather/brackets/compare/issue10846...adobe:pflynn/codefolding-tweaks~~

Improve visual appearance of code folding:
- Fainter, smaller gutter triangle icons
- Gutter triangle icons become less faint on mouseover (or when collapsed)
- Vertically align gutter triangle icons with text better
- Change inline collapsed placeholder to more readable arrow icon
- Better colors in dark theme
- Less horizontal gap between line numbers and triangle icons (without reducing gap between line numbers and text, if folding gutter disabled)

Usability: rename `"fadeFoldButtons"` pref to clearer `"hideUntilMouseover"`

And tweak code style issues:
- Remove most of ignored, unneeded package.json
- Preserve original CodeMirror copyright on CM-derived files
- Rename a few things for clarity
- Remove some unneeded code
- Tweak formatting & spelling in comments

**Update - adds a few more fixes/cleanups:**

More rigorous enable/disable flow for code folding: instead of piecemeal checking whether each bit has already initialized, atomically track whether init is done yet or not. Init/deinit all extant editors at once instead of doing each editor as it is activated, since splitview means editors may be already visible but not yet active. Fix listener leaks from repeated inits.

And keep menu items in a more predictable order:
- Add menu items at startup before user extensions load (at htmlReady instead of appReady), so core menu items like these don't move around as user extension load order changes
- Place QuickView menu item more explicitly, so it doesn't move unpredictably relative to Code Folding menu items (within the extensions/default folder, load order is arbitrary)
- Properly clean up menu divider, which used to get duplicated on each init pass

### DIFF
--- a/src/extensions/default/CodeFolding/main.js
+++ b/src/extensions/default/CodeFolding/main.js
@@ -29,6 +29,7 @@
 /*global define, $, brackets*/
 define(function (require, exports, module) {
     "use strict";
+    
     var CodeMirror              = brackets.getModule("thirdparty/CodeMirror2/lib/codemirror"),
         Strings                 = brackets.getModule("strings"),
         AppInit                 = brackets.getModule("utils/AppInit"),
@@ -44,7 +45,7 @@ define(function (require, exports, module) {
         COLLAPSE                = "codefolding.collapse",
         EXPAND                  = "codefolding.expand",
         EXPAND_ALL              = "codefolding.expand.all",
-        gutterName              = "CodeMirror-foldgutter",
+        GUTTER_NAME             = "CodeMirror-foldgutter",
         codeFoldingMenuDivider  = "codefolding.divider",
         collapseKey             = "Ctrl-Alt-[",
         expandKey               = "Ctrl-Alt-]",
@@ -52,7 +53,7 @@ define(function (require, exports, module) {
         expandAllKey            = "Shift-Alt-1";
 
     ExtensionUtils.loadStyleSheet(module, "main.less");
-
+    
     //load code mirror addons
     brackets.getModule(["thirdparty/CodeMirror2/addon/fold/brace-fold"]);
     brackets.getModule(["thirdparty/CodeMirror2/addon/fold/comment-fold"]);
@@ -81,7 +82,7 @@ define(function (require, exports, module) {
         cm._lineFolds = cm.getValidFolds(folds);
         prefs.setFolds(path, cm._lineFolds);
         Object.keys(cm._lineFolds).forEach(function (line) {
-            cm.foldCode(+line);
+            cm.foldCode(Number(line));
         });
     }
 
@@ -107,10 +108,10 @@ define(function (require, exports, module) {
       * Event handler for gutter click. Manages folding and unfolding code regions. If the Alt key
       * is pressed while clicking the fold gutter, child code fragments are also folded/unfolded
       * up to a level defined in the `maxFoldLevel' preference.
-      * @param {object} cm the codeMirror object
+      * @param {!CodeMirror} cm the CodeMirror object
       * @param {number} line the line number for the clicked gutter
       * @param {string} gutter the name of the gutter element clicked
-      * @param {object} event the underlying dom event triggered for the gutter click
+      * @param {!KeyboardEvent} event the underlying dom event triggered for the gutter click
       */
     function onGutterClick(cm, line, gutter, event) {
         var opts = cm.state.foldGutter.options, pos = CodeMirror.Pos(line);
@@ -118,7 +119,7 @@ define(function (require, exports, module) {
         var range;
         var _lineFolds = cm._lineFolds;
         if (cm.isFolded(line)) {
-            if (event.altKey) {//unfold code including children
+            if (event.altKey) { // unfold code including children
                 range = _lineFolds[line];
                 CodeMirror.commands.unfoldAll(cm, range.from.line, range.to.line);
             } else {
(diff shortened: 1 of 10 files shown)


src/extensions/default/CodeFolding/main.js
```diff
@@ -83,13 +76,12 @@ define(function (require, exports, module) {
         if (!editor || !saveFolds) { return; }
 
         var cm = editor._codeMirror;
-        if (!cm) {return; }
```
peterflynn: `editor._codeMirror` can never be null for any Editor, so I removed this check (the `editor` arg itself could be null -- e.g. in rare case where the `"documentRefreshed"` listener fires on a Document with no `_masterEditor` -- so I left that check above in place.

src/extensions/default/CodeFolding/main.js
```diff
@@ -232,27 +224,30 @@ define(function (require, exports, module) {
...
-            return;
-        }
         var path = editor.document.file.fullPath, _lineFolds = prefs.getFolds(path);
         _lineFolds = _lineFolds || {};
         cm._lineFolds = _lineFolds;
         var gutters = cm.getOption("gutters").slice(0);
         var lnIndex = gutters.indexOf("CodeMirror-linenumbers");
-        gutters.splice(lnIndex + 1, 0, gutterName);
+        gutters.splice(lnIndex + 1, 0, GUTTER_NAME);
+        $(editor.getRootElement()).addClass("folding-enabled");
```
peterflynn: This enables us to apply different padding to line numbers when the folding gutter is present vs. not.

src/extensions/default/CodeFolding/main.js
```diff
@@ -284,15 +279,18 @@ define(function (require, exports, module) {
     }
 
     /**
-        Initialise the extension
-    */
+     * Initialise the extension
+     */
     function init() {
-        if (CodeMirror.fold.combine || !prefs.getSetting("enabled")) {
```
peterflynn: Per discussion in the PR, it seems like `CodeMirror.fold.combine` should be needed in the extension-registry copy of this, but not in the core version since it's guaranteed to always load first.

peterflynn:
A few other things we should probably do before this story closes down:

* Update wiki docs on writing themes (themes now need to consider changing the gutter triangle & inline placeholder colors)
* Add folding-related preferences to table in "How to Use Brackets" wiki (once the churn has settled)
* Add notes to extension-writing wiki docs about how to add folding support for new languages

I'm happy to help with those three.

Lastly, @abose are you already tracking a task for creating unit tests to go with this feature?  If not, should we create a bug or Trello card to track that work?

src/extensions/default/CodeFolding/main.js
thehogfather: good to know :)

thehogfather:
@peterflynn Happy with these changes. I'll update PR #10850 later this evening and you can merge this.

thehogfather:
@peterflynn just to let you know I updated PR #10850. Merge your changes whenever you're ready.

MarcelGerber:
@peterflynn You got point 2 wrong for dark themes. Collapsed ones are _brighter_ (= _more_ outstanding) than expanded ones.

peterflynn:
@MarcelGerber That's by design -- collapsed should be more prominent.  So in the default light theme collapsed = darker, while in the dark theme collapsed = lighter.  Third-party themes can override these styles though.  I'll reword the bullet at top for clarity.

MarcelGerber:
Oh sorry, I got those wrong. What I meant was:
__Expanded__ ones are brighter (= more outstanding) than __collapsed__ ones.

peterflynn:
@MarcelGerber Hmm, that doesn't repro for me:

screen shot 2015-04-09 at 5 31 49 pm

Do you have a third-party theme installed that might be interfering with the styles, or anything?

peterflynn:
@thehogfather Fyi, I pushed a few more updates to this PR -- see updated description at top for details and a link to the correct diff view.  It should now enable/disable dynamically much more cleanly, keeping the menus in proper order better and working better with splitview (and hopefully with slightly simpler code).

I've rebased on top of your PR now, so until that's merged, this PR here will show the diff from both combined.

abose:
image
The icon is not rendered in my machine.

MarcelGerber:
@peterflynn Yeah you're right, it's an issue with my third-party theme (https://github.com/Brackets-Themes/Lion).
I'll file an issue over there as soon as this PR lands in master.

peterflynn:
@abose What OS are you on?  Does it show the other arrow icon (that the code on master uses) properly?

abose:
Windows 8.1. yes, it shows the arrow icon properly.

abose:
@peterflynn I was only fixing the unit test failures caused by the code folding extension in master.

I think it's best to add an issue to add unit test cases here as this is one of the core features now.

abose:
Now that PR #10850  has been merged, can this be merged?

MarcelGerber:
@abose Fixing the "code folding state doesn't get saved" issue (https://github.com/adobe/brackets/pull/10850#discussion_r28110161) is probably also something @peterflynn wants to fix in this PR (it's easy, even).

abose:
@peterflynn pull #10885 has the fix for the issue (#10850 (comment)), being a simple it can be done as part of this PR itself - or merge that pr after merging this .

peterflynn:
Let me just fix the Unicode issue @abose pointed out -- then I think this is good to merge.

peterflynn:
Ok, I spent a while playing with this and decided to switch from arrows to a simple "...":

screen shot 2015-04-13 at 7 16 24 pm

I actually like this a bit better than the wide arrows this PR had previously:

screen shot 2015-04-13 at 7 17 02 pm

And it's definitely way more legible than the original narrow arrows:

screen shot 2015-04-13 at 7 18 17 pm

abose:
Thanks @peterflynn for the quick fix.
+1 for 3 dots . 
Merging.

PR #10873, opened 2015-04-10
Fix for #10715 

### LINKED ISSUE #10715: Saving untitled document corrupts MRU order
I haven't been able to nail down solid repro steps for this yet, but I've seen it at least twice...

Symptom 1: messages like `/_brackets_26272468/Untitled-1 duplicated in mru list` appear in the console

Symptom 2: unsaved changes that you choose to drop when switching projects sometimes come back if you return to the same project without restarting Brackets in between

----

**Found repro steps:**

1. Open any file in your project in the working set
2. Create two new untitled files in a row (working set will now show: \<some file>, Untitled-1, Untitled-2)
3. Switch back to Untitled-1
4. Save the file anywhere
    * Result: ...`Untitled-1 duplicated in mru list` message in console
5. Press Ctrl-Tab -- switches to Untitled-2 as expected (but prints warning again)
6. Press Ctrl-Tab again -- switches to the file formerly called Untitled-1 as expected
7. Press and hold Ctrl, and hit Tab twice -- gets stuck on Untitled-2 and does not continue to the 3rd MRU entry, \<some file>)... and prints warning a 3rd time

prksingh:
Fixes #10715 Untitled documents aren't added to MRU until visited a 2nd time.

The issue is that no file Entry is added to File Index for Untitled documents. This causes a second file entry to be created and returned for untitled files from File System on File open (also on save of untitled file). Duplicate file entries for the same file in the MRU order causes the issue.

### DIFF
--- a/src/document/DocumentManager.js
+++ b/src/document/DocumentManager.js
@@ -454,6 +454,8 @@ define(function (require, exports, module) {
             now = new Date(),
             file = new InMemoryFile(fullPath, FileSystem);
         
+        FileSystem.addEntryForPathIfRequired(file, fullPath);
+
         return new DocumentModule.Document(file, now, "");
     }
     
--- a/src/filesystem/FileSystem.js
+++ b/src/filesystem/FileSystem.js
@@ -526,6 +526,22 @@ define(function (require, exports, module) {
     };
 
     /**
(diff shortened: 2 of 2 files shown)


prafulVaishnav:
Merging.. Thanks..!!

PR #10887, opened 2015-04-13
Adding a new first time popup for health data

prksingh:
This popup code is borrowed from Extract first time popup. This will be a non modal popup. Some details may still change

### DIFF
--- a/src/extensions/default/HealthData/HealthDataNotification.js
+++ b/src/extensions/default/HealthData/HealthDataNotification.js
@@ -33,6 +33,7 @@ define(function (require, exports, module) {
         Dialogs                      = brackets.getModule("widgets/Dialogs"),
         UrlParams                    = brackets.getModule("utils/UrlParams").UrlParams,
         HealthDataPreview            = require("HealthDataPreview"),
+        HealthDataPopup              = require("HealthDataPopup"),
         HealthDataManager            = require("HealthDataManager"),
         HealthDataNotificationDialog = require("text!htmlContent/healthdata-notification-dialog.html");
     
@@ -84,7 +85,7 @@ define(function (require, exports, module) {
             var isShown = PreferencesManager.getViewState("healthDataNotificationShown");
 
             if (!isShown) {
-                showDialogHealthDataNotification()
+                HealthDataPopup.showFirstLaunchTooltip()
                     .done(function () {
                         HealthDataManager.checkHealthDataSend();
                     });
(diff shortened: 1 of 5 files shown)


abose:
can you attach a screenshot of the new popup.

prksingh:
Good idea. Here it is: 
bracketshealthdatapopup

MarcelGerber:
Please make sure no Health Data is sent before the popup is accepted (that is, don't send Health Data when it's open).
One approach would be to wait 24 hours again before the first data is sent.

peterflynn:
I'd like to review this, but I may not be able to get to it tonight -- ok if I post feedback tomorrow?

prafulVaishnav:
@MarcelGerber We will send the healthData when the user will cancel the popup. Since its not modal, users have option to opt-out.

petetnt:
Shouldn't the popup have a way of opting out especially if the first set of data is sent immediately after the popup closes?

peterflynn:
@petetnt I think the idea was that you can leave the popup open while you opt out, but this doesn't seem obvious enough.  I'm proposing we wait at least 1 hour, even if the popup was closed, before any data could be sent.

I think Firefox waits 24 hours, but their popup is much easier to miss... while some other open-source apps have _zero_ wait time before sending metrics, which we definitely don't want to do.  Hopefully we'll land at a happy medium here :-)

petetnt:
@peterflynn I think that any reasonable delay would suffice, but like you said leaving the popup open doesn't really seem obvious as I'd say that the train of thought would be "Oh, I don't want to/can't/shouldn't send any information, I'll just close this popup and go change the settings..." :8ball:

prksingh:
I will have a discussion with @ryanstewart (should be today) and update the thread.

ryanstewart:
Thanks for the feedback @petetnt, you're right. I think 30 minutes is a good compromise. 30 minutes should be enough time for people to change it but will also give us information from people who are kicking the tires on Brackets. 

I don't want to wait an hour because if we wait that long we may not ever end up counting people who open it once, try it out, then leave. I think it's important to still count those users. 

Thanks again for the feedback @petetnt (and @peterflynn )

prksingh:
@peterflynn Has initiated another Pull request with this [here](https://github.com/adobe/brackets/pull/10912) which improves the UI and fixes some other health report issues as well. I am closing this pull request to avoid duplication

petetnt:
:+1:

peterflynn:
@ryanstewart Yep, very good point -- I remember that being an important motivation in the Firefox Health Report discussions as well.  Learning about users who try the app briefly but never come back is in some ways _more_ important than users who stay with the app for a long time, because it helps tell you what you're missing (in terms of usability, feature set, locale support, etc.).

Issue #10857, opened 2015-04-09
Code folding markers broken after deleting selection across collapsed section


peterflynn:
1. Start with this JS code:
```
define(function () {
    var foo;
    
    function one() {
        foo = 1;
    }
    
    function two() {
        foo = 2;
    }
    
    function three() {
        foo = 3;
    }
    
});
```

1. Collapse "two"
2. Select from the start of the `function` keyword in "two", to the start of the `function` keyword in "three"
3. Press delete
    * Result: folding triangle for "one" now shows it's collapsed, even though function "one" is _not_ collapsed
4. Click the folding triangle for "one" repeatedly
    * Result: nothing happens; triangle is nonresponsive
5. Undo
    * Result: triangle for "one" still show's it's collapsed. Triangle for "two" shows it's _not_ collapsed even though it _is_
6. Click triangle for "two" repeatedly
    * Result nothing happens on first click, but on 2nd click it starts working correctly
7. Click triangle for "one" repeatedly
    * Result: it still does nothing

peterflynn:
In other cases, this can cause phantom code-folding triangles to appear in places where no triangle was present before.  E.g.

1. Open CodeFolding/main.js
2. Collapse the block that starts with `var foldGutter`
3. Select from just before the "f" in `foldGutter` to just after the "n" in `function restoreL...`
4. Delete
    * Now there's a triangle on line 57 (`COLLAPSE_CUSTOM_REGIONS = ...`)

nethip:
@peterflynn Thanks for testing this extensively.

abose:
The problems are not present in https://codemirror.net/demo/folding.html

thehogfather:
The culprit function here is in foldgutter.js `updateFoldsCache`. Changes to the document and their corresponding effects on the line folds are not been correctly handled. The issue isn't present in the codemirror demo because it doesn't attempt to cache the lineFolds. 

A related issue is to hit `Ctrl-Z` after either one of the scenarios described above. The result is that the collapsed region is restored but the indicator on the gutter is wrong.

abose:
@thehogfather #10888 proposes a fix for the issue. can you check it out.

Also the issue of Ctrl-Z after either one of the scenarios described above. The result is that the collapsed region is restored but the indicator on the gutter is wrong.-- i still have not fixed that, any clue on what could be going wrong in that case?

thehogfather:
@abose checking #10888 at the moment.

[...]

Issue #10856, opened 2015-04-08
brace-fold frequently throws exceptions


peterflynn:
Put this in a JS file:
```
var test = {
    "foo": {
        "one": true,
        "two": true,
        "three": true
    },
    "bar": false
};
```

1. Select lines 3-5 entirely (i.e. everything inside the foo `{}`s)
2. Press delete

Result: `Uncaught TypeError: Cannot read property 'lastIndexOf' of undefined`

1. Undo
2. Select lines 2-6 entirely (i.e. the entire `"foo": { ... },` block)
3. Press delete

Result: same exception thrown again

Nothing immediately obvious is broken afterwards, but it spams the console and looks bad, so I think we need to fix this for 1.3.

peterflynn:
Other examples:

```
define(function () {
    "use strict";
    
    function one() {
        foo = 1;
    }
    
    function two() {
        foo = 2;
    }
    
});
```

Deleting lines 2-3 will throw an exception.
Deleting lines 4-7 or 4-11 will throw an exception.

```
define(function () {
    "use strict";
    
    var foo;
});
```

Deleting lines 2-3 (or 2-4) even in this very simple case will throw an exception.

abose:
The unit test failures after the code folding integration and this bug has the same root cause.

abose:
Verified the fix, closing.

PR #10888, opened 2015-04-13
Code folding fixes

abose:
**Do not merge yet** - we should land PR #10859(Code folding UI & code style tweaks) first.

Fixes Unit test failures after code folding integration and issues #10857 , #10856 

point 6 in issue #10857 is still fix in progress.

### DIFF
--- a/src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
+++ b/src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
@@ -211,8 +211,8 @@ define(function (require, exports, module) {
             var lineChanges = changeObj.text.length - changeObj.removed.length;
             //update the lineFolds cache
             updateFoldsCache(cm, changeObj.from.line, lineChanges);
-            if (lineChanges !== 0 && changeObj.from.line + lineChanges >= 0) {
-                updateFoldInfo(cm, changeObj.from.line + lineChanges, changeObj.from.line + lineChanges + 1);
+            if (lineChanges !== 0) {
+                updateFoldInfo(cm, Math.max(0, changeObj.from.line + lineChanges), Math.max(0, changeObj.from.line + lineChanges) + 1);
             }
             state.from = changeObj.from.line;
             state.to = 0;
--- a/src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
+++ b/src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
@@ -151,6 +151,30 @@ define(function (require, exports, module) {
         });
     }
 
+    function syncDocToFoldsCache(cm, from, lineDiff) {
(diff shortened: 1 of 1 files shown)


abose:
The only change is in /foldhelpers/foldgutter.js, All other changes are part of PR #10859

thehogfather:
This looks a good fix for the phantom fold marks issue.

abose:
@peterflynn waiting for PR #10859 to be merged.

peterflynn:
@abose I think there might be something funny with the git history in this PR -- if I view a [comparison against my branch](https://github.com/adobe/brackets/compare/pflynn/codefolding-tweaks...CodeFoldingFixes), it appears to also contain a bunch of ALF changes.  Please check to make sure the history is in an ok state before anyone merges this.

Also, in the future please use `abose/` as a prefix on your branch names -- since all committers are pushing branches to this repo, having informal "namespaces" is important to help keep branches organized.

src/extensions/default/CodeFolding/foldhelpers/foldgutter.js
```diff
@@ -214,7 +211,7 @@ define(function (require, exports, module) {
             var lineChanges = changeObj.text.length - changeObj.removed.length;
             //update the lineFolds cache
             updateFoldsCache(cm, changeObj.from.line, lineChanges);
-            if (lineChanges !== 0) {
+            if (lineChanges !== 0 && changeObj.from.line + lineChanges >= 0) {
```
peterflynn: I'm not sure I understand this part of the change. The only time `changeObj.from.line + lineChanges < 0` could happen is when a bunch of lines were deleted at the start of the file.  Do we actually want to skip running `updateFoldInfo()` in that case?

I'm wondering if what we really want is something more like `Math.max(0, changeObj.from.line + lineChanges)` on the line below this instead... and similar for the 2nd arg too.

abose:
changed, `Math.max(0, changeObj.from.line + lineChanges)` and calling `updateFoldInfo()` sounds more logical.

abose:
point 6 in issue #10857 is Also fixed with the change.

abose:
Thanks @thehogfather for verifying.
Merging the fix.

peterflynn:
Ah, ok so @thehogfather got a chance to look at the additional changes here that occurred after my review?  There's a whole bunch of new code here that was added (`syncDocToFoldsCache()` and a bunch of other changes in commits 45c7a99b, d69ade37, & 684a507e)...  Normally that would require a re-review by someone before merging.

thehogfather:
I'll get some more time to test and review more rigorously this evening and update the thread.

abose:
@thehogfather Could you help with fixing some of these issues now that we are nearing the release date (probably next week), It would help very much if you could!

thehogfather:
Yes I can. I'll have a look this evening.

abose:
Thanks :)

PR #10898, opened 2015-04-14
Getting Started project: remove useless whitespace, fix an arrow

valtlai:
In Getting Started project
– remove useless whitespace
– replace “an arrow” (`&lt;`) with “->”

### DIFF
--- a/samples/root/Getting Started/index.html
+++ b/samples/root/Getting Started/index.html
@@ -1,6 +1,5 @@
 <!DOCTYPE html>
 <html>
-    
     <head>
         <meta charset="utf-8">
         <meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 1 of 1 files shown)


abose:
:+1:  :smiley: 
merging.
image

peterflynn:
Hmm, I actually disagree with all of the changes in this pull request.

We want the whitespace and indentation in the file to be consistent with what Brackets generates by default -- it was before, and now it's not.

All Brackets help documentation has standardized on using ">" when referring to menu items -- this changes it to "->" in one place only.

We want the Getting Started content for all locales to stay in sync -- this changes one locale but not the others.

So, I propose we revert this before 1.3 ships.  @abose any objections?
@valtlait does that reasoning make sense to you?  Let me know if you'd like to discuss further.

valtlait:
@peterflynn It’s ok, because of #10920 (2<sup>th</sup> comment)

PR #10908, opened 2015-04-15
Brackets Box/Flex CSS Update

puzzledShark:
Presently brackets is using an outdated version of flex, due to this the sidebar is not rendered properly within firefox.

Refer to: https://github.com/humphd/brackets/issues/180 for more information on this issue/fix explaination

### DIFF
--- a/src/htmlContent/extension-manager-dialog.html
+++ b/src/htmlContent/extension-manager-dialog.html
@@ -2,11 +2,11 @@
     <div class="modal-header">
         <ul class="nav nav-tabs no-focus">
             {{#showRegistry}}
-                <li><a href="#registry" class="registry" data-toggle="tab"><img src="styles/images/extension-manager-registry.svg"/><br/>{{Strings.EXTENSIONS_AVAILABLE_TITLE}}</a></li>
-                <li><a href="#themes" class="themes" data-toggle="tab"><img src="styles/images/themes-icon.svg"/><br/>{{Strings.EXTENSIONS_THEMES_TITLE}}</a></li>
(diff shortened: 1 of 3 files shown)


abose:
@KeniMorri it looks like you have not signed the [BRACKETS CONTRIBUTOR LICENSE AND DISTRIBUTION AGREEMENT](http://dev.brackets.io/brackets-contributor-license-agreement.html )

le717:
I'd recommend holding off on merging this until Linux gets CEF 2721 so nothing breaks on that platform (and I think I heard something about that possibly happening in 1.4?).

peterflynn:
Earlier, flexbox caused us some very bad performance issues.  We should check:

a) whether this PR causes any performance regressions
b) whether this PR causes any performance _improvements_ (maybe newer syntax uses a newer, faster impl under the hood) -- and if so, is it safe to start using flexbox more widely in the Brackets UI after this?

KeniMorri:
Updated because I forgot to include prefixes for safari
@abose my bad, I've signed it now

abose:
Marking this for 1.4, will update if there is any performance issues as peter pointed out.

abose:
I see that there is an overlap in hte project tree rendering.
image

redmunds:
@abose I just ran into this problem on an internal project (so I can't post any links). The solution for my particular case was to specify the `min-height` as the `flex-basis` parameter instead of `auto`, which is what Brackets appears to be doing -- see the `.flex-item()` mixin.

abose:
Thanks @redmunds . 
Waiting for PR to be updated with the fix.

PR #10912, opened 2015-04-15
Health Report UX improvements

peterflynn:
This is a superset of #10887.  @prksingh sorry, I hope this doesn't step on anything you'd already started writing... I set out to just tweak the strings & CSS, and this sort of snowballed after that :-P  I tried to split this cleanly into separate commits, so if you really wanted to take only the visual changes I think you could do that.

Here are the differences from 10887:
- Don't send data 30 until minutes after showing the popup, regardless of how quickly the user closes it - this gives users time to read more and opt out if desired, even if they close the popup before changing the pref.
  - Note: I took this opportunity to simplify the send-time calculations a bit by changing viewstate from "last time sent" to "next to to send."
- Tighten wording in Health Data notification popup (ran this by @ryanstewart & @njx for feedback)
- Add similar wording for context to Health Data preferences/preview dialog
- Show full preferences dialog when Help menu item is invoked, even if user is currently opted out (it's actually impossible to opt back in if we showed the popup instead, as 10887 does)
- Improve popup layout & show/hide animation
- Add tooltips to links in popup & dialog
- Fix bug with popup placement on Linux (due to HTML menu bar)
- Remove unused `HealthDataNotification.showDialogHealthDataNotification()` and change unit test to check notification popup instead
- Small refactorings: rename ExtensionManager flag for clarity; use ExtensionManager constant instead of duplicating string literal; share link-tooltip code better.


peterflynn:
Here's what the UI looks like now with this PR:

screen shot 2015-04-15 at 4 40 58 pm

screen shot 2015-04-15 at 4 41 37 pm

peterflynn:
@prksingh et al: regardless of whether we merge this or some other variation of PR #10887, we should probably do another test pass on this feature overall since the notification & the data-send timing is changing significantly from how it used to behave.  (I tested this a bunch locally of course, but it's important to have a 2nd set of eyes).

prksingh:
This looks good @peterflynn Thanks for the UI tweaks! I will take another look at this later and merge this.

@prafulVaishnav Can you also have a look at the changes?

prafulVaishnav:
@peterflynn If user have opt-out then we should not be showing the health data preview as in that case we will not be logging/sending anything. Instead we should show the preference dialog where he can opt-in.

peterflynn:
@prafulVaishnav That change is by design -- if the user has opted out, it's still useful for them to be able to see what data _would_ be sent if they were to opt back in.  Maybe they'll decide it's not so scary after all, for example :-)  The UI is now unified: the dialog contains both the preference and the preview.

prafulVaishnav:
@peterflynn @nethip We had some discussion on it based on which we decided to keep different dialog box for preference(to opt-in) and preview(containing option to opt-out) Previewing the data on opt-out will give the feeling to the user that we are still tracking the data which we aren't. At some later stage, we will be logging more data like user events, workflows, etc.

peterflynn:
It seem important for people to be able to see the data and then decide if they're ok with sending it.  The Firefox one works this way too.  Given that the checkbox is right there, do you think people would really misunderstand?  @ryanstewart, thoughts?

nethip:
@peterflynn I think people would have more confidence when we show them a dialog saying we are not tracking your data anymore as you have opted out against showing them the data even they have opted out. Talking about the Firefox model, I personally would feel a little odd to see the tracking data though I have opted out of it.

peterflynn:
One of the points of the Firefox Health Report -- and hopefully here too -- is that the data can benefit the user directly by showing them stats about their performance, average project size, etc.  So you don't want to refuse the user access to their own data just because they've opted out of sharing it with others.  This is also useful for field diagnostics -- e.g. if a user complains Brackets is running slow, we could ask them to check the Health Report dialog and see if the projects they tend to edit are unusually large, etc.

What about having the string vary depending on opt-in state?  E.g. when opted out, it could say something like: "Below is a preview of the data that _would_ be sent in your next Health Report if enabled (sharing your Health Report is currently disabled)."

nethip:
@peterflynn Yeah I am OK with the changing of the wording. We need to clearly distinguish between diagnostic data and tracking data. I am in for showing the former in case I opt out. 

I can compare this to be a browser cookie just waiting to be used.

nethip:
That being said I am fine if we can express the same with good wording.

src/extensibility/ExtensionManager.js
```diff
@@ -789,7 +789,7 @@ define(function (require, exports, module) {
     exports.updateExtensions        = updateExtensions;
     exports.getAvailableUpdates     = getAvailableUpdates;
     exports.cleanAvailableUpdates   = cleanAvailableUpdates;
-    exports.isRegistryObjectUpdated = false;
+    exports.hasDownloadedRegistry = false;
```
MarcelGerber: nit: indentation.
Also, a newline before this line would separate the functions above and this state variable even more.

src/widgets/Dialogs.js
```diff
@@ -413,6 +413,26 @@ define(function (require, exports, module) {
...
+        if (elementOrDialog.getElement) {
+            $element = elementOrDialog.getElement().find(".dialog-message");
+        } else {
+            $element = elementOrDialog;
+        }
+        $element.find("a").each(function (index, elem) {
+            var $elem = $(elem);
+            var url = $elem.attr("href");
+            if (url && url !== "#") {
+                $elem.attr("title", url);
```
MarcelGerber: Maybe you should make sure the element doesn't have a title yet.

peterflynn:
@nethip Ok, I'll push up the dynamic wording changes tonight.

src/extensions/default/HealthData/HealthDataPopup.js
```diff
@@ -0,0 +1,84 @@
+/*
```
prafulVaishnav: Later we may want to generalize this for the popup other than HealthDataNotification.

peterflynn:
@nethip It felt weird having the wording jump around as the checkbox is toggled, so I went with a more static set of wording that emphasizes the previewed data isn't sent if you've opted out:

screen shot 2015-04-16 at 11 49 25 pm

(this also simplifies the checkbox label itself, after some discussion with NJ & @ryanstewart)

@MarcelGerber Thanks for the feedback -- changes pushed to fix those nits too

@prafulVaishnav I agree we may want a more generalized Brackets notification UI widget in the future.  Are you ok with waiting until a future PR to generalize this, though?  It's good to have more concrete use cases in mind before generalizing anyway...

prafulVaishnav:
@peterflynn Yes in future we will take care of generalizing it. Thanks a lot Peter. Merging this.

PR #10914, opened 2015-04-16
Codefolding linefolds syncing fixes

thehogfather:
This PR addresses some inconsistencies with updating the folded states in the folds cache when lines are removed and added.
cc @abose @peterflynn please review and test.


abose:
Thanks @thehogfather for the quick fix. Will test and update.

abose:
It looks very stable now, all the cases mentioned in the original bug and comments are now working fine.
All the unit tests are alo passing (exept the 3 pending quick edit ones which we are working on actively to fix.)

abose:
Thanks, Hog father, It works well now in my day of testing

Issue #10869, opened 2015-04-10
Bucket code folding states by project in state.json instead of dumping every thing into one bucket.


abose:
rather than dumping folding state for all files together in one map, it might make sense to bucket it by project, like we do for other similar prefs like the scroll position & selection state of editors.

MainViewManager has an example of how to do that:

            context = { location : { scope: "user",
                                     layer: "project" } },
            state = PreferencesManager.getViewState(PREFS_NAME, context);
and

            context         = { location : { scope: "user",
                                         layer: "project",
                                         layerID: projectRoot.fullPath } },
        ...
        PreferencesManager.setViewState(PREFS_NAME, state, context);
Add a line note

peterflynn:
Nominating for 1.3 because it would be a pain to change this later...

abose:
PR merged, So closing the issue.

PR #10919, opened 2015-04-17
Store code-folding view state in project-specific buckets

peterflynn:
Store code-folding view state in project-specific buckets for cleaner organization. If we later offer a way for users to delete stored data for a project, this ensures code folding data will be cleaned up along with
everything else. (This is also how we store Quick Edit expand/collapse state, which is very similar).

I suggested this originally in https://github.com/adobe/brackets/pull/10850#discussion_r28110431.  It's not super critical, but it will be much easier to do now (before code folding initially ships) -- if we did this later, we'd have to write and test additional preferences migration code.

@thehogfather does this look ok to you?

### DIFF
--- a/src/extensions/default/CodeFolding/Prefs.js
+++ b/src/extensions/default/CodeFolding/Prefs.js
@@ -8,9 +8,11 @@
 /*global define, brackets*/
 define(function (require, exports, module) {
     "use strict";
-    var PreferencesManager          = brackets.getModule("preferences/PreferencesManager"),
+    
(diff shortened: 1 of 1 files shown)


abose:
#10869 will be fixed by this change list. Testing the changes as this would be good to go in today's prerelease.

abose:
Thanks for submitting this fix @peterflynn. The state.json has the fold stated bucketed in the project prefs now, Also all the unit tests are passing. Also the fold states are persisted across project switch and brackets restart.
Merging .

thehogfather:
looks great :+1:

PR #10920, opened 2015-04-17
Revert Getting Started index.html changes

peterflynn:
Reverts adobe/brackets#10898 -- see my comments there

@abose, if this sounds ok to you could you merge?

### DIFF
--- a/samples/root/Getting Started/index.html
+++ b/samples/root/Getting Started/index.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html>
+    
     <head>
         <meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


peterflynn:
We could still remove the trailing whitespace on _non-blank_ lines -- but we should update all locales at once for thoroughness.  It also seems reasonable to remove that one stray blank line before the `</p>`.  All the others reverts here I think we definitely want though.

valtlait:
@peterflynn There’s no need to escape “>” to “&gt;”. It could be better to use “>” because it looks nicer.

abose:
Merging as per discussion in pull #10898

PR #10922, opened 2015-04-17
Fix failing QuickOpen unit tests

peterflynn:
Fix QuickOpen unit tests after introduction of QuickSearchField - listens for Enter on keydown, not keyup; expects an "input" event to follow every text change (Enter key processing now waits until text change events have caught up so Enter key never takes effect if change event never comes)


rroshan1:
Thanks Peter. The Tests are now passing through on both Windows and Mac.

PR #10926, opened 2015-04-17
Sample projects: remove trailing whitespace on non-blank lines, fix arrow

valtlai:
- remove trailing whitespace on non-blank lines
- fix an arrow: `&lt;` to `>`
- remove the blank line just before `</p>`

See #10920, comments 2–3


samples/fr/Premiers pas/index.html
```diff
@@ -57,7 +57,7 @@
         <samp>
             Une petite démonstration ? Placez le curseur de la souris sur la balise <!-- <samp> --> et tapez <kbd>Cmd/Ctrl + E</kbd>. Un éditeur rapide CSS apparaît en superposition, avec la règle CSS applicable. La fonction Edition rapide est également utilisable pour les attributs de classe et d’ID. Vous pouvez aussi vous en servir sur vos fichiers LESS et SCSS.
             
-            Vous pouvez créer de nouvelles règles en procédant de la même manière. Cliquez sur l’une des balises <!-- <p> --> en haut du document et appuyez sur <kbd>Cmd/Ctrl + E&lt;/kbd>. Il n’existe aucune règle associée pour le moment, mais vous pouvez cliquer sur le bouton Nouvelle règle afin d’ajouter une règle pour <!-- <p> -->.
+            Vous pouvez créer de nouvelles règles en procédant de la même manière. Cliquez sur l’une des balises <!-- <p> --> en haut du document et appuyez sur <kbd>Cmd/Ctrl + E</kbd>. Il n’existe aucune règle associée pour le moment, mais vous pouvez cliquer sur le bouton Nouvelle règle afin d’ajouter une règle pour <!-- <p> -->.
```
peterflynn: Good catch!

peterflynn: @pantkowiak @ybayer Looks like there is an ALF issue here -- both the French & Japanese translations had this character escaped when it should not have been.  We should make sure the next ALF update doesn't cause this problem to come back.

peterflynn:
The HTML spec says `>` "SHOULD" be escaped as `&gt;`, but since it also says the motivation for that is buggy older browsers, I think we can ignore that advice -- that change here is ok.

samples/root/Getting Started/index.html
```diff
@@ -28,11 +28,10 @@
...
         <p>
             <em>Brackets is a different type of editor.</em>
-            Brackets has some unique features like Quick Edit, Live Preview and others that you may not find in other 
-            editors. And Brackets is written in JavaScript, HTML and CSS. That means that most of you using Brackets 
-            have the skills necessary to modify and extend the editor. In fact, we use Brackets every day to build 
+            Brackets has some unique features like Quick Edit, Live Preview and others that you may not find in other
+            editors. And Brackets is written in JavaScript, HTML and CSS. That means that most of you using Brackets
+            have the skills necessary to modify and extend the editor. In fact, we use Brackets every day to build
             Brackets. To learn more about how to use the key features, read on.
-
```
peterflynn: Several other languages have this extra line also.  Do you want to remove it in all of them for consistency?  (Please squash commits if you do, for history cleanliness).

samples/fr/Premiers pas/index.html
pantkowiak: @peterflynn  I did some digging, and this is not an ALF issue. Somehow the opening bracket for </kbd> tag was escaped during our pre-process for translation and post-process. This was missed in post-process, because we never had such issue. I've added info about this to our instructions, so this should not be an issue in the future:)

samples/root/Getting Started/index.html
valtlait: @peterflynn Done.

PR #10934, opened 2015-04-18
Switch "Brackets" to "{APP_NAME}" in strings

dvorapa:
Please switch these in root strings

### DIFF
--- a/src/nls/root/strings.js
+++ b/src/nls/root/strings.js
@@ -602,11 +602,11 @@ define({
 
     // extensions/default/HealthData
     "HEALTH_DATA_NOTIFICATION"                  : "Health Report Preferences",
-    "HEALTH_FIRST_POPUP_TITLE"                  : "Brackets Health Report",
-    "HEALTH_DATA_DO_TRACK"                      : "Share anonymous information on how I use Brackets",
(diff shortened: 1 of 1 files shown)


busykai:
Nice catch! Thank you! This looks good! I'm not merging this only because I'm not sure where in the release cycle we are.

ryanstewart:
Agreed, great catch @dvorapa. I think we're all set to merge this since it's a pull request against the release branch. The only thing I'm not sure about is the impact on translations. Once we get the final translations we should do a scrub to make sure we're using `{APP_NAME}` there as well. 

@nethip do you know what the best way to handle this is?

busykai:
@ryanstewart, @nethip :

> The only thing I'm not sure about is the impact on translations. Once we get the final translations we 
> should do a scrub to make sure we're using {APP_NAME} there as well.

This is why I wasn't sure. It would be preferable if this was merged to `master` and then re-translated for the next release... It may make it a little difficult to catch (for the translator) if it was merged to the `release` branch since normally root strings are frozen when `release` is cut. Or may be I'm over complicating the logistics...

nethip:
@ryanstewart I agree with @busykai that we should be merging this with `master` and have the translations happen there. That way, we would not be effected with the string translations for this release, as we are put in a *string freeze* on `release` branch.

@dvorapa Could you please re-target this to `master`. And thanks for catching this :smile:

nethip:
@dvorapa Don't bother re-targeting this to `master`. I am merging this I had a word with LOC team and they are going to send the strings for translations after merging this.

dvorapa:
I can not manage to re-target that. It looks like my `patch-2` branch is more than 5 commits ahead from Brackets' `master` (forked from the `release` branch after first commits with german translation were merged). And I don't want to make a mess in Brackets commit history. Nevertheless, I tried it in #10991. Feel free to close #10991, if it disn't what

peterflynn:
@dvorapa Fwiw, the reason we didn't use `{APP_NAME}` originally was because it seemed unlikely strings about sending data to an Adobe server would apply to anyone else's spinoff apps -- so forks with their own distribution will probably need to replace these strings anyway.

PR #10939, opened 2015-04-18
Added JSP language object based on htmlEmbedded mode

noragrossman:
Added a JSON object for JavaServer Pages files, using the HtmlEmbedded mode from CodeMirror

### DIFF
--- a/src/language/languages.json
+++ b/src/language/languages.json
@@ -36,7 +36,7 @@
     "html": {
         "name": "HTML",
         "mode": ["htmlmixed", "text/x-brackets-html"],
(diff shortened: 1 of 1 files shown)


ficristo:
Based to the [CodeMirror documentation](http://codemirror.net/mode/htmlembedded/index.html) the change seems fine.
I wonder if you should also add `"text/x-brackets-html"` to the `mode` field.
Meanwhile @noragrossman could you updated the PR?

My only concern, its true that Brackets targets html technologies, but I wonder if java (and jsp) should be added as extension.

@MarcelGerber WDYT?

ficristo:
@noragrossman thank you for your contribution.
As of now we don't accept new languages.

PR #10952, opened 2015-04-20
Implement zooming by Ctrl + mouse wheel

valtlai:
This PR implements zooming by Ctrl + mouse wheel feature. Works also by pinching a touchpad.

~~There’s an possible issue: the font size changes by 2, not 1.~~

### DIFF
--- a/src/view/ViewCommandHandlers.js
+++ b/src/view/ViewCommandHandlers.js
@@ -308,6 +308,18 @@ define(function (require, exports, module) {
         setFontSize(DEFAULT_FONT_SIZE + "px");
     }
 
+    /** Zooming by mouse wheel */
+    window.addEventListener("wheel", function(e) {
+        // Detect Ctrl key on Linux and Windows, or Cmd key on Mac
+        if (e.ctrlKey || e.metaKey) {
+            if (e.deltaY < 0) {
+                _handleIncreaseFontSize();
+            } else {
+                _handleDecreaseFontSize();
+            }
+        }
+    });
+
     /**
      * @private
      * Updates the user interface appropriately based on whether or not a document is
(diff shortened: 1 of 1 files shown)


MarcelGerber:
Your "possible issue" should have been fixed by #10930.

valtlait:
@MarcelGerber Yes, so it is.

redmunds:
When using mouse wheel, I want editor to scroll (which is a common operation for me) not change font size (which I rarely change) so this should be a preference.

valtlait:
@redmunds User needs to press Ctrl when zooming by mouse wheel.

redmunds:
@valtlait I missed that -- sounds good!

peterflynn:
See also #3250 for an earlier attempt.  Sounds like our newer CEF doesn't have the Ctrl key problem on Windows, though.

On Mac I think the usual modifier for zoom is Alt though, not Cmd.  Should probably check the platform rather than accepting both modifiers on both OSes.

TomMalbran:
In #3250 there are a couple of test for this feature which could still work after changing line 86 to:
```oEvent.initMouseEvent("DomMouseScroll", true, true, doc.defaultView, detail, 0, 0, 0, 0, true, false, false, false, 0, null);```.

This line will have to be changed if we use alt on mac and ctrl on win or do not accept both modifiers.

valtlait:
Now, Ctrl is for Windows and Linux and Alt is for Mac. Does pinch zoom work when using Alt key?

jesobreira:
Sorry for necro... This PR is what made this possible? So it's not a bug, but a feature? Or this PR cares about different things?

captura de tela 2018-06-07 as 21 49 37

PR #10957, opened 2015-04-20
ALF Automation

walfgithub:
Brackets: Please merge this pull request created by ALF. @ArturOledzki @pantkowiak @ybayer FYI

### DIFF
--- a/src/nls/fr/strings.js
+++ b/src/nls/fr/strings.js
@@ -416,7 +416,7 @@ define({
 	"CMD_RELEASE_NOTES": "Notes de mise à jour",
 	"CMD_GET_INVOLVED": "Participer",
 	"CMD_SHOW_EXTENSIONS_FOLDER": "Afficher le dossier d’extensions",
-	"CMD_HEALTH_DATA_STATISTICS": "Rapport des données sur l’état de santé",
+	"CMD_HEALTH_DATA_STATISTICS": "[8033837] !é=Health Report=!",
 	"CMD_HOMEPAGE": "Page d’accueil de {APP_TITLE}",
 	"CMD_TWITTER": "{TWITTER_NAME} sur Twitter",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
     "name": "Brackets",
-    "version": "1.3.0-0",
-    "apiVersion": "1.3.0",
+    "version": "1.4.0-0",
```
nethip: @walfgithub This bump up in version no is not required. Is this an automation error?

src/config.json
```diff
@@ -23,8 +23,8 @@
         "healthDataServerURL" : "http://healthdev.brackets.io/healthDataLog"
     },
     "name": "Brackets",
-    "version": "1.3.0-0",
-    "apiVersion": "1.3.0",
+    "version": "1.4.0-0",
+    "apiVersion": "1.4.0",
```
nethip: Same here.

peterflynn:
I think we should close this start over with clean commits -- this pulls in part of the history of `master` which should not be on the release branch, but it's hidden from the diff because another commit reverts that history.

To see this, click on the first commit above (94b72f9), then click the "parent" link in the upper right. You'll see that the parent is 9264e29c, which is the commit that changes version from 1.3 to 1.4 -- that commit should stay on master.

To fix this, ALF needs to start a new branch from scratch, branched off of release instead of master, then create a new commit with the changed files there.

peterflynn:
I think btw one problem with merging this as-is is that it would case problems when we later merge release back into master: since the commit that reverts 9264e29's diff is "past" 9264e29 in the history, it will essentially overrule 9264e29, and the version number on master would jump back to 1.3 after merging in the release branch.  Fixing that would mean master gets 3 commits that bump up, revert, then bump up again the version number.

nethip:
Closing this as we have already merged this. https://github.com/adobe/brackets/pull/10967

Issue #10959, opened 2015-04-20
Health Report *does not work* in release builds


peterflynn:
1. Download & run the 1.3 prerelease build.  Leave Health Report enabled and then wait.

No data is ever sent.  If you look in dev tools, it's sending POSTs to the local Brackets index.html file instead.

This is because config.json is actually machine-generated and gets overwritten by the build script (see [write-config task](https://github.com/adobe/brackets/blob/master/tasks/write-config.js)).  The health report URL needs to go into brackets.config.json as well, so that when the script regenerates config.json it is preserved.

See discussion in PR #2654, where config.json was introduced, about this exact pitfall :-/

prafulVaishnav:
@peterflynn Thanks a lot Peter for figuring it out before release. I have created [PR](https://github.com/adobe/brackets/pull/10961)

prafulVaishnav:
PR merged so closing it..

PR #10961, opened 2015-04-21
Adding HealthDataServerURL to brackets.config.json

prafulVaishnav:
Initially it was added to config.json. I was not aware it was machine generated file and picks up value from brackets.config.json. Fix for Issue #10959

### DIFF
--- a/src/brackets.config.json
+++ b/src/brackets.config.json
@@ -20,6 +20,7 @@
         "extension_registry"         : "https://s3.amazonaws.com/extend.brackets/registry.json",
         "extension_url"              : "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default" : true,
-        "build_timestamp"            : ""
+        "build_timestamp"            : "",
+        "healthDataServerURL"        : "http://healthdev.brackets.io/healthDataLog"
     }
 }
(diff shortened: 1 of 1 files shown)


prafulVaishnav:
@peterflynn Can you please review this.

peterflynn:
@prafulVaishnav Looks good.  Do we still have another PR coming to change this over to an https URL?

prafulVaishnav:
@peterflynn Yes we will be changing the URL to https but it is not available at present. But definitely we will changing it before release. Should we wait for that before merging this?

abose:
We could wait to merge this, The SSL certs are under processing and would be available within the week.

prafulVaishnav:
@abose Can you take a look and merge it?

abose:
@prafulVaishnav Is there a timeout after which the health data would be sent? I tested the pr with clean preferences and i couldnt find the logs being seen at the server. I tried closing and reopening brackets and the logs was still not being sent.

pvaishnav:
@abose First time data will be sent after 30 minutes and then after time period of 24 hours. You need to update the config.json as well if you are working on this pr as it still contains 'dev' url.

prafulVaishnav:
@abose First time data will be sent after 30 minutes and then after time period of 24 hours. You need to update the config.json as well if you are working on this pr as it still contains 'dev' url.

abose:
Have you verified that the data is sent to the production server for the source at the release branch? 
Starting from a fresh preferences?

prafulVaishnav:
@abose Yes I have verified it. Are you not able to still send the logs?

abose:
We are getting the test logs at the server, merging.

Issue #10977, opened 2015-04-23
Popup boxes are off center


CPCookieMan:
This is driving me up a wall. It looks terrible. Something changed between 1.1 and 1.2 that made popup boxes go all wonky. It makes brackets-git look really weird as well, as the popup boxes got smaller for some reason, and the titles for popup boxes go on to two lines.
![centering](https://cloud.githubusercontent.com/assets/1424353/7289410/b1c981e6-e93d-11e4-9b52-9caed3cd17af.PNG)
Upon further review, it's still happening to me all the way back to 1.0. I'm almost sure this wasn't always the case. What could cause that?
And on yet more review, it would seem that .modal-wrapper is on a position:absolute system. It has left:0 and right:0 set, with margins of auto. Theoretically, that should center it, but for some reason, 0 is actually 360px out into the editor (which as far as I can tell, doesn't correlate with anything. It's not like the size of the sidebar or anything.) This can be solved by setting left:-360px, but that seems pretty hacky. Any ideas?
![centering2](https://cloud.githubusercontent.com/assets/1424353/7289738/467fec2c-e942-11e4-9c87-4826a176f01a.PNG)

CPCookieMan:
I've submitted a pull request which has fixed the problem. Request #10978.

nethip:
@CPCookieMan This problem got introduced because of a change in our Extract for Brackets plugin. Basically the plugin was overriding some CSS styles. Could you try this with our latest build. This must not be there.

ficristo:
Closing as was an Extract bug.
If you can still reproduce on latest release please open a new issue.

PR #10978, opened 2015-04-23
Modal positioning fix

hollinsky:
Fixing my bug, Issue #10977, this properly centers modals within the Brackets window.

The old:
centering2

The new:
newlook

It looks so much cleaner with this fix.

### DIFF
--- a/src/styles/brackets_patterns_override.less
+++ b/src/styles/brackets_patterns_override.less
@@ -734,11 +734,13 @@ a:focus {
 /* Dialog-related styles */
 
 .modal-wrapper {
-    width: 100%;
+    width: 100% !important;
+    max-width: 75%;
     height: 100%;
     position: absolute;
     top: 0;
     left: 0;
+    padding-left: 0;
     overflow: auto;
     display: table;
 }
--- a/src/styles/brackets_patterns_override.less
+++ b/src/styles/brackets_patterns_override.less
@@ -888,6 +888,8 @@ a:focus {
 
 .dialog-title {
     color: @bc-text-thin;
+    padding-top: 10px;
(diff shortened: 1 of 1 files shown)


src/styles/brackets_patterns_override.less
```diff
@@ -734,11 +734,13 @@ a:focus {
 /* Dialog-related styles */
 
 .modal-wrapper {
-    width: 100%;
+    width: 100% !important;
+    max-width: 75%;
```
prafulVaishnav: Why we need max-width ; 75% here as It will not centrally align the popup to the whole Brackets window.

src/styles/brackets_patterns_override.less
```diff
@@ -886,6 +888,8 @@ a:focus {
 
 .dialog-title {
     color: @bc-text-thin;
+    padding-top: 10px;
+    padding-bottom: 15px !important;
```
prafulVaishnav: padding will create unnecessary space

prafulVaishnav:
@CPCookieMan Thanks for the fix but This PR is not needed as the bug was introduced due to some extension. Can you please close this PR?

CPCookieMan:
The problem was introduced by the Extract for Brackets extension.

js0511:
NAVER - http://www.naver.com/
--------------------------------------------

oiup2006@naver.com 님께 보내신 메일 <Re: [brackets] Modal positioning fix (#10978)> 이 다음과 같은 이유로 전송 실패했습니다.

--------------------------------------------

받는 사람이 회원님의 메일을 수신차단 하였습니다. 


--------------------------------------------

PR #10993, opened 2015-04-24
Fix 2 translation errors in the extension manager (Italian language)

fredev:
I am an Italian user of brackets, the first error I found is in the extension installed tab title, the word 'installata' is the singular form, the simple fix is to use the plural 'installate'.
The 2nd error is in the button 'extension_installed' where the plural form 'installate' is used, here we refer to a single extension, so the fix is to use the singular form 'installata'.

Hope this help, best regards!


Denisov21:
I'm an Italian user, for me is a good translation!

prafulVaishnav:
@Denisov21 Just to clarify, are you saying changes suggested by @Fredev are good?

Denisov21:
Yes exactly @prafulVaishnav!

PR #11043, opened 2015-05-02
Refresh editor when reloading file to avoid scroll area getting out of sync

njx:
This is a possible fix for #9632. My suspicion is that without this refresh, CodeMirror's idea of the scroll area size gets out of sync somehow if the text has a significantly different number of lines. It's not clear why this should be necessary - there might be some bug in CodeMirror - but it seems like a fairly straightforward workaround.

With this change in place I haven't seen the bug in #9632 at all today, whereas it shows up pretty often for me on master. It seems like a relatively harmless change, although it could add some small delay when switching documents.

@peterflynn any thoughts?

### DIFF
--- a/src/editor/Editor.js
+++ b/src/editor/Editor.js
@@ -931,6 +931,7 @@ define(function (require, exports, module) {
         
         // This *will* fire a change event, but we clear the undo immediately afterward
         this._codeMirror.setValue(text);
+        this._codeMirror.refresh();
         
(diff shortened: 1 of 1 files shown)


njx:
I've been running with this for the past few days and haven't seen the scrolling problem, and I also haven't seen any other issues with it, so I think it's a good fix.

zaggino:
Tested and seems to fix the issue, merging.

PR #11128, opened 2015-05-16
[Review Only] Find in Files incremental update

redmunds:
This is a branch I started during "Innovation Days" in November, 2014.

Starting this pull request to use for discussion while this is being reviewed/developed.

DONE:
- update FiF panel incrementally
- update spinner during search
- search (sequentially) groups of files (in parallel) for balance of speed and responsiveness

TODO:
- add Cancel button to FiF panel
- Closing FiF panel should stop the search
- Unit tests for `Async.doInSequentialBatches()`

Note:
- Search speed refers to _first_ search. Subsequent searches are fast because all files are loaded to memory

Testing
- unit tests
- Hit count is updated while search is in progress
- Next Page and Last Page buttons should work while search is in progress
- Spinner animation works while search is in progress
- Search can be cancelled
- Cancel button removed after search completes


src/search/FindInFilesUI.js
```diff
@@ -102,10 +109,45 @@ define(function (require, exports, module) {
...
+                    progressCount += results.matches.length;
+                    if (!progressFirstUpdate) {
+                        // First update
+                        _resultsView.open();
+                        if (_findBar) {
+                            _findBar.close();
+                        }
+                        progressFirstUpdate = true;
+                    } else if (!progressFirstPage) {
+                        // First page
```
redmunds: 100 should be defined in a constant (e.g. ENTRIES_PER_PAGE) if it's not already defined somewhere.

TomMalbran: constant is in SearchResultsView

src/search/FindInFilesUI.js
```diff
@@ -102,10 +107,45 @@ define(function (require, exports, module) {
...
+                        if (progressCount >= 100) {
+                            progressCount -= 100;
+                            progressFirstPage = true;
+                        }
+                        _resultsView._updateResults();
+                    } else {
+                        // Subsequent full pages
+                        if (progressCount >= 100) {
+                            progressCount -= 100;
+                            _resultsView._showSummary();
```
redmunds: @TomMalbran The reason `_updateResults()` was being called is so user can click "next page" or "last page" button while search is in progress. If only summary is being updated, then `_updateResults()` will have to be called when either "next page" or "last page" button is clicked, right?

TomMalbran: Since the files are now sorted before searching, and since we are only updating every 100 results there shouldn't be any problem. The idea is that after I press "next page" or "last page" the results shown in that page are complete and won't change when searching over more files.

The point is that the first 100 results go in the first page, the next 100 results in the following and so on. So if I watching the results from 301-400 and the search is over 1000, the next results will go on the page 1001-1100 and nothing on the page I am on right now.

_updateResults() basically just renders the panel and then restores the scroll position and item selected. So it is just needed when updating the page that the user is watching, which we shouldn't need to do. But I am not sure if we can assume the same if we try to process x files in parallel instead of doing all sequentially.

TomMalbran: I still noticed that clicking on next is kind of hard since the buttons are constantly being updated. So we can do something fancier than rendering the summary and just update the text and the state of the buttons.

TomMalbran: Actually I think that there might be a case where we need to update the panel. That case is when  instead of getting 100 results we get 120 (it can happen since we are searching on files), and we are seeing the results from 101-200. In this case if we find more results, we need to update the current page, but we can know when we hit this case and update the results.

redmunds:
@larz0 Can you try out this branch and make a suggestion for a Cancel button?

There's already an (x) button in upper-right corner that dismisses panel (and *should* also stop the search). But sometimes I do a search and forget to use an Exclusion set, and I see the thing I'm looking for, so I'd like to stop the search without dismissing the panel or clearing the results.

TomMalbran:
@redmunds Should we change the close button in the Find in Files panel for a cancel button? I think that we need to change how we update the summary so that it is not fully rerendered.

It would be really nice to also show a progress bar (in how many more files we need to search). It could be just a simple percentage calculated between the total amount of files to search and the files already searched. This will make it easier to know when the search finished.

larz0:
@redmunds @TomMalbran how about showing "(Cancel)" while it's still searching, like this:

screen shot 2015-05-19 at 9 42 32 am

redmunds:
@TomMalbran A progress bar would be nice and should be easy to implement, but there's a spinner displayed in statusbar so not critical. As I [described above](https://github.com/adobe/brackets/pull/11128#issuecomment-103558228) in the "But sometimes..." sentence, I think we need both Cancel and Close buttons.

redmunds:
@larz0 I was thinking button should be over to right next to Close button because if it's embedded in that text, it may move around (and be tricky to click) as numbers change.

TomMalbran:
@redmunds Oh. I see. Is more like a stop than a cancel. But if it stops should we be able to continue?
I forgot about the statusbar loader. So I guess that we don't need a progress bar.

Are you going to implement something to process n files in parallel? With the change I did, searches with lots of results seem pretty fast even if it is sequentially. Changing the dom is really slow, so the less we change the better. So I am thinking of just changing the text in the summary instead of all.

@larz0 I agree with redmunds about a button. It even took me a while to find the link the first time I saw the design.

larz0:
@redmunds @TomMalbran good point, how about put it after the search term then. The right hand side is a little far away (dependent of screen size) and would seem a little out of context.

redmunds:
@TomMalbran 

> Are you going to implement something to process n files in parallel?

@peterflynn said that there is already code to do this, so I'd like to give it a try. Sending an progress update for every single file seems like overkill. One the other hand, this could cause the file order to change, so maybe it's not worth it.

TomMalbran:
It would be nice to see if that would make it faster. I tried to implement it, but I am not sure if I did it right since I didn't saw any bit difference.

If it works by sending a notify every x consecutive files, it might work. Something like processing the first 50 files and when those are done it sends a notify and it starts processing the second 50 files.  So, if for example, we get 400 results from the first 50 files, we know that those are the first 400 entries that will go on the first 4 pages. So on the second notify we shouldn't need to change the first 4 pages, since the results will start on page 5.

larz0:
I wonder if the panel title should say "Finding…" before the end of the find process.

rroshan1:
I think we should have the message *Finding "\<query string\>"* while the results get populated and the "cancel" button right in front of it- Easy to locate and giving the complete message like *Cancel Finding \<query string\>*.

rroshan1:
The final number of matches is inconsistent/incomplete. It only updates after a click on any of the next/last page buttons.
I am taking it up.

redmunds:
I think UI currently only updates for every 100 items, so the last fraction of 100 are missed. We'll need to know when search is complete, especially if we want to change the message that's displayed.

redmunds:
I changed search to (sequentially) process 50 files at a time (in parallel). I ran some tests

```
batch-size  time(ms)
----------  --------
All (7157)    8124
200           8591
100           8831
 50           9283
 20           9816
  1          13016
```

**Notes:**
- Machine used for measurements: Win7 lenovo W530 w/SSD
- Used *Debug > Show Performance Data* "FindIn" data
- not very accurate, only 1-2 samples for each test
- Search I used was:
  - "brackets" in Brackets code base (7157 files)
  - case insensitive
  - don't exclude any files
- 50 seems like a nice balance of speed and responsiveness
  - note this is *files*, nor FIF *results*

**Questions:**
- Should batch size be a preference?
- Panel is not shown until first results, so there's a delay before panel is first shown. Should we start with loading an empty panel?

abose:
@redmunds we have had a rework on find in files and could you checkout the branches,
prksingh/instantsearch
and prksingh/fifnode
Basically, we moved all the search logic to the node process. it is not final yet, but we are seeing very promising results.

redmunds:
@abose, @prksingh  

> we have had a rework on find in files and could you checkout the branches, prksingh/instantsearch
and prksingh/fifnode Basically, we moved all the search logic to the node process. it is not final yet, but we are seeing very promising results.

I tried out the prksingh/fifnode branch (fyi, I couldn't find prksingh/instantsearch) and ran the same test [as I did above](https://github.com/adobe/brackets/pull/11128#issuecomment-116297972) with the following results. Times reported are in ms for: first-search/subsequent-search.

I ran randy/fif-incr-update branch with batch sizes of 50 & 200.

```
master                       4173/1150
prksingh/fifnode             3070/3070
randy/fif-incr-update  (50)  8447/4077
randy/fif-incr-update (200)  7801/4030
```

The initial search of prksingh/fifnode is significantly faster than master, but subsequent searches are much slower than master.

The goal of the randy/fif-incr-update is to increase performance *perception* by providing a non-blocking interface, so I think it's good that improving Find in Files is being approached from both angles. But it should not be taking so much extra time to update UI, so this still needs some work.

nethip:
@redmunds Thanks for trying out this branch and testing it too. About incremental search, the changes are in this branch `prksingh/instantSearch` where `Search` is with capital `S`.  I agree with you that we should not degrade on performance just to unblock the UI. This work is still in progress. Nevertheless your inputs are really valuable. I am thinking we could integrate your changes along with `fifnode` changes. @abose Could you take a look at this PR and see if we can integrate this to our `fifnode` code.

src/search/FindInFiles.js
```diff
@@ -417,15 +418,26 @@ define(function (require, exports, module) {
...
-            perfTimer = PerfUtils.markStart("FindIn: " + scopeName + " - " + queryInfo.query);
+            perfTimer = PerfUtils.markStart("FindIn: " + scopeName + " - " + queryInfo.query),
+            searchDeferred = new $.Deferred();
         
-        return candidateFilesPromise
+        candidateFilesPromise
             .then(function (fileListResult) {
                 // Filter out files/folders that match user's current exclusion filter
                 fileListResult = FileFilters.filterFileList(filter, fileListResult);
+                fileListResult = FindUtils.getSortedFiles(fileListResult, function (file) { return file.fullPath; }, FindUtils.getOpenFilePath());
```
redmunds: @TomMalbran I've been trying to speed up this code and I discovered that the initial call to `FindUtils.getSortedFiles()` here that you added is taking a lot of time -- ~3000ms in [the scenario I use here](https://github.com/adobe/brackets/pull/11128#issuecomment-116751074). Even worse, this is before panel is displayed the first time, so there's a 3 sec gap with no ui feedback!

I think that the files are all correctly sorted *except* that we want to move current doc to the top of the list. Is that correct? If so, we could write a separate function to do that much faster.

Let me know what you think. I think this is the main blocker to get this into master.

TomMalbran: Is weird that it takes so long, since it is just sorting the files, and sorting is usually fast. And even in master this sorting is done before the panel is displayed, but after the search is done (but only on the files with results). And actually, we are still sorting them on render, which we could try to skip.

If I remember correctly, I don't think that the files were sorted and I had to sort them for the UI. But if it is sorted, then great. And if it is not, we can keep a sorted version of the files if this is an issue.

src/search/FindInFilesUI.js
TomMalbran: @redmunds I think that we should check if `SearchResultsView._currentStart + 100 > SearchModel.numMatches` and if is true we can do `_updateResults()` and if not, we can just update the summary.

src/search/FindInFilesUI.js
```diff
@@ -74,16 +74,21 @@ define(function (require, exports, module) {
...
+        
+        searchInScopePromise
             .done(function (zeroFilesToken) {
                 // Done searching all files: show results
                 if (FindInFiles.searchModel.hasResults()) {
-                    _resultsView.open();
-
-                    if (_findBar) {
-                        _findBar.close();
+                    if (!progressFirstUpdate) {
```
TomMalbran: I removed this since I thought that we could never get to this point. But I think that it would be good to re-add it: (replace this line with):
```
if (progressFirstUpdate) {
    _resultsView._updateResults();
} else {
```
The alternative would be to check in which page we are, and depending on the result update just the summary or the whole panel.

src/search/FindInFiles.js
nethip: @TomMalbran Sorting indeed is taking some time. That is what we figured out when we moved the existing code to a node module.

redmunds: @TomMalbran @nethip Note that it only takes a long time the first time, so I suspect it's causing the project file structure (?) to get loaded into memory. It would be nice if this process could be started when Find in Files dialog is invoked, but it would probably have to be refactored to be done incrementally so it doesn't not prevent smooth typing in the dialog.

I also thought about showing results panel earlier (to show that we're doing something), but it currently doesn't get displayed if there are no results. We'd have either dismiss it after displaying it (yuck!) or change it so panel is always displayed whether or not there are results (which means more UI to design).

TomMalbran: @redmunds It still feels weird. `getSortedFiles` only sorts the files, since the first parameter is the list of files. Maybe the slow part is comparing the paths? But it still makes no sense that it is slow only the first time.

We should already have the list of files from the project model. And if sorting is an issue, we could keep a sorted copy in the project model?

redmunds: @TomMalbran `SearchModel` stores the results as an object with path as the key. `SearchResultsView` needs to re-sort list anyway because iterating over properties of an object does not guarantee any specific order. Also, file searching is done in parallel within batches, so there's no guarantee that search results are returned in sort order. So, sorting needs to be done in the `SearchResultsView`, and it looks like it's already done in `SearchResultsView._render`.

So, order of results added to panel as search progresses may not be linear. Sorting files here may *reduce* (not eliminate) that, but unless we can figure out what the slowdown is we shouldn't do it. Just moving current doc to top of list might be good enough.

TomMalbran: @redmunds I still don't know if `fileListResult` is sorted. If it is, great. If not, we need to figure what to do. Maybe first figure why the sort can take so long.

If the sort in the render is a problem, that can be fixed by using a sorted array for the results and an object to store the indexes.

redmunds: I was able to refactor `FileUtils.compareFilenames()` to eliminate some unnecessary file manipulations and comparisons in commit https://github.com/adobe/brackets/commit/607979620cc8ebc0ebe11856217d51b7fb56c021.

 This reduces time spent in this function by ~40% - from ~3.0s to ~1.8s (in the scenario I'm using on my Win system).

redmunds: The bad news is that the ~1.8s lag is on *every* search (not just the *first* search that I reported earlier). I also verified that `fileListResult` is *not* sorted, and not sorting list provides random results.

TomMalbran: That is great, and it makes more sense. I am guessing that if 1.8s is still too much, then we would need to keep a sorted copy of the file list. Maybe instead of sorting the entire list, we could copy the list, grab the first 50 files, remove them from the list and repeat this on each process, which might be a bit faster than sorting the entire list?

src/file/FileUtils.js
```diff
@@ -434,19 +434,23 @@ define(function (require, exports, module) {
...
-            filename2 = getFilenameWithoutExtension(filename2);
+        function cmpExt() {
+            var ext1   = getFileExtension(filename1),
+                ext2   = getFileExtension(filename2);
+            return ext1.toLocaleLowerCase().localeCompare(ext2.toLocaleLowerCase(), lang, {numeric: true});
+        }
+        
+        function cmpNames() {
+            if (brackets.platform === "win") {
+                filename1 = getFilenameWithoutExtension(filename1);
```
MarcelGerber: This will entirely drop the extension, so `cmpExt` won't work afterwards.

redmunds: Good catch! Need to use local vars.

redmunds: Fixed.

src/search/FindInFiles.js
redmunds: I optimized file list sorting in commit https://github.com/adobe/brackets/commit/d5cf805a0a346da60cf17ebc0a47e58b6df08e8f to get the sort down to ~1.5s, but that's probably as good as we can do.

TomMalbran: The order used for find in files is the same order as the one used on the project tree (on windows), when it is fully open. Is there a way to get the files from the project tree so that we don't need to sort them?

redmunds: The list comes from `ProjectModel._getAllFilesCache()`. It should be easy to change `allFilesVisitor()` function to insert into sorted list (currently just to end), but I'm sure that will cause a slowdown somewhere else.

I also see that `_resetCache()` is called *every* time `ProjectModel.handleFSEvent()` is called, which happens on every write to `state.json`, which happens on every click in file tree. So, that may need to be made smarter about when to clear cache.

TomMalbran: I think that we could also go back to the original implementation where we didn't sort the files upfront and updated the UI on every progress. When I did the change we were updating it after every file searched, but since now we are doing it after 50 files it might not be as slow as before, even if updating the DOM is slow.

redmunds: Note that even though files are searched in batches of 50, each file still notifies when results are found, so we still need to use result count.

The down side of re-sorting list as results come in is that you may try to click on an item that changes under your mouse -- very annoying!

rroshan1:
Requesting all to try this variant of Find In Files- Instant search #11375 and share your feedback.

nethip:
@redmunds With these fif code moved to node, how does this PR fit? Can we reuse some parts of this PR in the new node code? 

I have moved this to the next milestone.

redmunds:
@nethip 

> Can we reuse some parts of this PR in the new node code?

I created a new PR for the `compareFIlenames()` function: https://github.com/adobe/brackets/pull/11530

abose:
Closing in favor of #11530.
Can any other changes from this pr be ported to the new impl apart from this?

PR #11129, opened 2015-05-16
[Review Only] Fix adjusting selection after edit when an edit overlaps subsequent selection

redmunds:
This is for https://github.com/adobe/brackets/issues/10284

The problem seems to be that `Document.adjustPosForChange()` was handling the edit range as _exclusive_ instead of _inclusive_.

Block and Line Comments seemed to rely on this handling, so I added more explicit selection handling for those commands. This broke many unit tests, which almost all these commands, so we need to **identify other commands** (both core and extensions) that insert & remove text and **test** those.

Note: I originally found this bug using my [Quick Markup extension](https://github.com/redmunds/brackets-quick-markup) so that needs to be updated for this fix.

cc @njx @peterflynn

### DIFF
--- a/src/document/Document.js
+++ b/src/document/Document.js
@@ -472,7 +472,7 @@ define(function (require, exports, module) {
     
     /**
      * Adjusts a given position taking a given replaceRange-type edit into account. 
-     * If the position is within the original edit range (start and end inclusive),
+     * If the position is within the original edit range (start and end exclusive),
      * it gets pushed to the end of the content that replaced the range. Otherwise, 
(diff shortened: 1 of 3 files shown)


redmunds:
I played around with this some more and found that the following case is not fixed. Start with multi-cursor selection with cursors 1 char apart:

```html
    <p> | | </p>
```

Then *Edit > Block Comment*

redmunds:
@njx (or anyone else) Can you comment on whether you think this is the right approach?

If so, I can fix the explicit IP handling case [I found](https://github.com/adobe/brackets/pull/11129#issuecomment-103074275).

PR #11130, opened 2015-05-16
Preferences Code Hints

sprintr:
This PR implements code hints for the preferences we use in brackets. Additionally it also shows a little description of the active preference to the user.

prefrence

CC:
@redmunds @ryanstewart @MiguelCastillo


rroshan1:
Glad to see it ready!! :smiley:

petetnt:
This is pretty awesome :+1: 

The `data.json` and it's description field should be localization/internationalization-enabled and it would be nice if extensions could extend that, something in lines of:

``` js
//preference (string), type, description
CodeHintsManager.register("some.preferenceHere", "boolean", Strings.some.preferenceHere.description");
```
-> 
``` json 
"some.preferenceHere: {
  "type": "boolean",
  "description": "Some preference is true / false"
}
```

Not sure if those are feasible to land in 1.4

test/spec/JSONUtils-test.js
```diff
@@ -0,0 +1,257 @@
...
+        it("should detect a correct exclusionList for an object", function () {
+            // Between " and tabSize"
+            ctxInfo = JSONUtils.getContextInfo(testEditor, {line: 15, ch: 13});
+            expect(ctxInfo.exclusionList).toEqual([
+                "useTabChar", "closeBrackets"
+            ]);
+        });
+        it("should detect an array", function () {
+            // After [
+            ctxInfo = JSONUtils.getContextInfo(testEditor, {line: 24, ch: 23});
```
abose: refactoring candidate with a for loop.

abose: Cancel that, i see that this is just a test file :+1:

abose:
+1 
@sprintr Why is this marked as review only?

sprintr:
@abose Wanted to get some input on this so I marked it review only.

@petetnt Going to add i18n support today. Some preferences have got their own UI, so we don't want to have each and every pref in the code hints. I was thinking about providing an API to extension devs, so they can plug in their hints. Something like `PreferencesManager.setHints()` and `PreferencesManager.getHints()` should do the job.

As far as the core brackets preferences, we should use data.json, so translators (who are encouraged to commit on the web) can easily translate descriptions in a single file.

@rroshan1 Glad you liked it. :)

sprintr:
Added internationalization support, moved all descriptions to `nls/root/strings.js`

MarcelGerber:
I'm not too set about adding all these strings to the core nls file, as it's quite a lot of new strings.
Maybe we should consider creating another nls file for this?

sprintr:
@MarcelGerber We can also move it to its own `strings.js` file. I didn't see any other core extension doing it.

abose:
I think it would be ok; if not easier to manage in the core file itself. adding js files for separate features would eventually decrease the maintainability of translations[as i saw in several other large projects]. For instance you would have to search over several files to find a string const; but with just one file, they are generally easy to locate; also people would need to go through several files to see if some strings changed to figure out translations.

sprintr:
- Added a new API `PreferencesManager.getAllPreferences` to return all the preferences defined.
- Added support for preference hinting by third-party extensions using the preferences they have defined. [Here](https://gist.github.com/sprintr/b48c318ced6af39bf123) is a tiny extension to demonstrate how it can be done.

@abose Is there anything else remaining to cover in this PR?

petetnt:
Nice :+1:

src/preferences/PreferencesBase.js
```diff
@@ -1324,6 +1328,15 @@ define(function (require, exports, module) {
         },
 
         /**
+         * Returns a clone of all preferences defined.
+         *
+         * @return {Object}
+         */
+        getAllPreferences: function () {
+            return $.extend(true, {}, this._knownPrefs);
```
TomMalbran: Instead of `$.extend` for cloning you can use `_.clone` and `_.cloneDeep` from lowdash which is already included in the core.

src/extensions/default/PrefsCodeHints/main.js
```diff
@@ -0,0 +1,386 @@
...
+                if (data[ctxInfo.parentKeyName] && data[ctxInfo.parentKeyName].keys) {
+                    keys = data[ctxInfo.parentKeyName].keys;
+                } else if (ctxInfo.parentKeyName === "language") {
+                    keys = LanguageManager.getLanguages();
+                    option.type = "object";
+                } else {
+                    keys = data;
+                }
+            
+                hints = $.map(Object.keys(keys), function (key) {
```
TomMalbran: You could just use `Object.keys(keys).map(function (key) { ... })`

sprintr: `Array.prototype.map` will return undefined if it doesn't match. A possible workaround would be to use a `Array.prototype.filter` on the array returned by `map` to exclude `undefined`. $.map works fine for this scenario.

sprintr:
@abose Can you give this PR a review.

abose:
@sprintr will do some test with the changes and update tomorow :+1: 
Also if you feel that this PR is good to go, You could remove the [review only] tag .

src/extensions/default/PrefsCodeHints/unittests.js
```diff
@@ -0,0 +1,573 @@
+/*
```
abose: Could this file be moved to the `/test/` folder?

src/extensions/default/PrefsCodeHints/main.js
```diff
@@ -0,0 +1,385 @@
...
+     */
+    function formatHints(hints, query) {
+        
+        var hasMetadata = hints.some(function (token) {
+            return token.type || token.description;
+        });
+        
+        StringMatch.basicMatchSort(hints);
+        return hints.map(function (token) {
+            var $hintItem = $("<span>").addClass("brackets-pref-hints"),
```
abose: codehints
The width of the selection keeps jumping; Would it be possible to fix the width and wrap the hint description. Maybe with <divs> and some css with `display: block;    visibility: visible;    white-space: pre-wrap;`

src/extensions/default/PrefsCodeHints/main.js
```diff
@@ -0,0 +1,385 @@
...
+    
+    // List of parent keys for which no key hints will be provided.
+    var parentKeyBlacklist = [
+        "language.fileExtensions",
+        "language.fileNames",
+        "path"
+    ];
+    
+    // Define a preference for code hinting.
+    PreferencesManager.definePreference("codehint.PrefHints", "boolean", true);
```
abose: How about more argument is added to the definePreference API and (maybe the last parameter as description/ object with possible values for the prefs) and then raise a `preferences- code hints missing` warning so that extension devs can directly incorporate code hints to their preferences?

src/extensions/default/PrefsCodeHints/unittests.js
sprintr: This is a standard practice that most of the core extensions do.

src/extensions/default/PrefsCodeHints/main.js
sprintr: The last parameter is optional, we can ping extension authors about incorporating their hints if they want to.

sprintr: We also don't want to have all the preferences in code hints, because some of the preferences are managed from UI.

redmunds: > We also don't want to have all the preferences in code hints, because some of the preferences are managed from UI.

Some people prefer to edit preferences manually, so I think all preferences should be in code hints. Also, it's tricky to set a UI managed preference when it's language or path specific, so that's another case for manually editing preferences that have GUI controls.

sprintr: Preferences related to themes and fonts don't work if changed via brackets.json which is why they don't appear in code hints.

nethip:
@sprintr Thanks for putting up this PR :+1: Really appreciate it!

I had a small suggestion. I think right now one has to manually go about editing the json file to make sure the preference name, along with its values kind of popup in code hints, while editing the preference file. How about adding another parameter to `PreferencesManager.definePreference()` that would take the list of preference values, along with the description, as an optional value. 

This way extension writers can make sure their extension's preference names and values popup in the code hint. 

Also we need to clearly document what needs to be done, going forward, to create a new preference in Brackets core. I was thinking if we can automate the job, but I might sound crazy :smirk_cat:

sprintr:
@nethip I wrote a [tiny extension](https://gist.github.com/sprintr/b48c318ced6af39bf123) to demonstrate how extension authors can plugin their hints. Just modified `PreferencesManager.definePreference`.

abose:
@sprintr 
250px works very well now :simple_smile:
I still think that all the preferences that can be configured using the preferences file can be hinted. Maybe we could specify in the wiki or in the hints itself that some preferences need a relaunch of brackets to take effect.

sprintr:
@abose I am going to do two things tomorrow.
1. Allow all the preferences in code hints.
2. Move the descriptions of preferences to the files where they are defined. So we won't need updating data.json with new prefs in future.

abose:
+1 :)
But that would result in a major rewrite of this pr i guess ..

sprintr:
1. Each and every preference defined is now available through code hints.
2. Added `src/extensions/default/PrefCodeHints/unittest-files/preferences.json` which carries a few preferences to run unit tests against. This approach will allow us not to break unit tests if we add new preferences in the future.
3. The descriptions of all the preferences have now been moved to where they are defined.
4. The descriptions of preferences that are not defined still lies in `data.json`.
5. The code hints won't jump to different width if a code hint doesn't have description.
6. The value of a **value token** will be replaced if the very next token is a **value token**.

---
@abose Rebased it against master. If it looks good to you then you can proceed to merge it.

MarcelGerber:
I have one extension whose preference name is so long that a horizontal scrollbar is shown.
The horizontal scrollbar now covers the info text:
image

src/view/ThemeSettings.js
```diff
@@ -156,8 +156,12 @@ define(function (require, exports, module) {
         prefs.set("themeScrollbars", defaults.themeScrollbars);
     }
 
-    prefs.definePreference("theme", "string", defaults.theme);
-    prefs.definePreference("themeScrollbars", "boolean", defaults.themeScrollbars);
+    prefs.definePreference("theme", "string", defaults.theme, {
+        description: Strings.DESCRIPTION_THEME
```
MarcelGerber: For some reason, this pref doesn't show up for me.

src/brackets.js
```diff
@@ -437,7 +438,9 @@ define(function (require, exports, module) {
         }
 
         if (brackets.platform === "win" && !brackets.inBrowser) {
-            PreferencesManager.definePreference("_windowsScrollFix", "boolean", true).on("change", enableOrDisableWinScrollFix);
+            PreferencesManager.definePreference("_windowsScrollFix", "boolean", true, {
```
MarcelGerber: Is there some way to exclude a preference?
This is more or less a "hidden" preference that no one should ever need to change himself (and it will also go away at some point), so I'd appreciate it if this pref wasn't shown.
It wouldn't be much of an issue normally, but right now, this pref is the first to show up in the list...

abose: Exclude certain preferences from code hints:  will add an issue to track this after this is merged.

src/view/ThemeSettings.js
MarcelGerber: Sorry, my fault. It does show up.

src/extensions/default/PrefsCodeHints/styles/brackets-prefs-hints.css
```diff
@@ -0,0 +1,29 @@
+.brackets-pref-hints .matched-hint {
+    font-weight: 500;
+}
+.dark .brackets-pref-hints .matched-hint {
+    color: #ccc;
+}
+
+.brackets-pref-hints .hint-obj {
+    display: inline-block;
+    width: 250px;
```
abose: Changing this to `min-width 300px` and the below `width` to 300px could be a solution to the horizontal scrollbar problem.

abose:
@sprintr works very well now :+1:  .will merge this today after the horizontal scrollbar merge is in.

sprintr:
@abose We should wait until #11195 get merged since that seems to be a nice way to do it. I'll add marquee effect to hints exceeding limits and functionality to exclude hints. Have to remove few merge conflicts too.

abose:
@sprintr Right now the scroll-bar fix truncates long code hints. Instead of truncating, if we change the css for `.brackets-pref-hints .hint-obj ` from `width: 250px;` to `min-width 300px` ; Then instead of truncating the hint; the popup will take the size of the largest preference variable. The hint description will be wrapped at 250px. This would avoid any truncation and also resolve the layout problems. I think this might be more suited than truncating variable names.

MarcelGerber:
@sprintr LanguageManager actually defines its preferences, so you can modify the code there instead of including them in `data.json`: https://github.com/adobe/brackets/blob/5d4269ca359fc52095d5e7704a6b8a70c74dcd4c/src/language/LanguageManager.js#L1147-L1152

With #11197, pretty much all preferences (except the meta `language` and `path`) will be defined.

sprintr:
@MarcelGerber You can now check marquee and exclusion of code hints. for exclusion add `exclude: true` in the options.

MarcelGerber:
Very nice! I also love the marquee effect

src/extensions/default/PrefsCodeHints/main.js
```diff
@@ -0,0 +1,413 @@
...
+     */
+    function PrefsCodeHints() {
+        this.ctxInfo = null;
+
+        // Add all the preferences defined except the excluded ones.
+        var preferences = PreferencesManager.getAllPreferences(),
+            preference;
+        Object.keys(preferences).forEach(function (pref) {
+            preference = preferences[pref];
+            if (preference.excluded) {
```
MarcelGerber: Maybe something more distinct, like `excludeFromHints`?

src/brackets.js
```diff
@@ -437,7 +438,9 @@ define(function (require, exports, module) {
...
+                description: Strings.DESCRIPTION_WINDOW_SCROLL_FIX
```
MarcelGerber: You could just exclude that pref (and remove the corresponding string) :+1:

sprintr: @MarcelGerber Cool. I'll do that. There is another pref that is not used `defaultExtension`, we can exclude that one two.

sprintr:
@MarcelGerber Removed the width of hint object and it seems we don't need marquee anymore. Can you check it.

abose:
@sprintr some jumps will be there with just `width`. If you change it to `min-width` it will work good .

sprintr:
@abose Yeah, it works that way. It only happens where no description is available.

abose:
Verified. All unit tests passing. Merging.

abose:
@sprintr Thanks for this awesome contribution :tada:

sprintr:
@abose Awesome!

busykai:
@sprintr, great! this is really helpful!

nethip:
Great collaboration work on this PR! @abose @sprintr Does it make sense to update our existing documentation with the updated usage of `definePreference()` and also share [this](https://gist.github.com/sprintr/b48c318ced6af39bf123) as an example?

abose:
Issue #11200 tracks the doc changes required and hints for the new preferences API
@ryanstewart is reviewing the code hints being shown for preferences.

nethip:
:cool:

PR #11148, opened 2015-05-21
added Slack channel and re-ordered contact order

mackenza:
re-ordered the list to reflect the best chance of actually receiving help.

Should we remove the IRC link? It seems to be not well-populated anymore and is redundant based on Slack usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,9 @@ Not sure you needed the exclamation point there, but we like your enthusiasm.
 
 #### Contact info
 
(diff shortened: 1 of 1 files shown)


ryanstewart:
I like it. Merging.

mackenza:
@ryanstewart any thoughts on the IRC question?

ryanstewart:
I'm not sure we want to take it completely out at the moment, but I think we eventually will. I just can't imagine an open source project without an IRC channel :)

But if Slack continues to work out well for us, I don't see any reason we'd need/want to keep it.

PR #11157, opened 2015-05-23
Revert "Rename CodeMirror2 folder to CodeMirror"

abose:
Reverts adobe/brackets#11150


nethip:
@abose Could we close this PR now as we have already done a due diligence for the side effects of renaming CodeMirror2 to CodeMirror?

PR #11195, opened 2015-06-01
Fix horizontal scrolling of code hints

sprintr:
This removes the horizontal scroll bars appearing when hints exceed the size.

@abose This should get in before we add #11130 

@MarcelGerber Can you merge this branch in your local `pref-code-hints` branch and check if it works for you?

### DIFF
--- a/src/styles/brackets_patterns_override.less
+++ b/src/styles/brackets_patterns_override.less
@@ -379,6 +379,9 @@ a:focus {
         box-shadow: 0 3px 9px @bc-shadow;
         border: none;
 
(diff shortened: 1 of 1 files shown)


MarcelGerber:
Apparently, this only works for unfiltered code hints?
image

MarcelGerber:
Ouch. Looks like I forgot to restart after editing the `.less`.
Works fine, thank you.

It also works well with other Code Hint providers. Merging.

sprintr:
@MarcelGerber Cool.

PR #11235, opened 2015-06-09
[MAC] Enabling back sub pixel antialiasing for code view

nethip:
Defaulting the code view antialiasing to sub-pixel which can be overridden using a new preference `fontSmoothing`. This preference takes `subpixel-antialiased` or `antialiased`. With these changes, the UI would have gray scale anti-aliasing but the code view container `editor-holder`, would have subpixel-antialiasing on. 

The reason to disable subpixel antialiasing for UI is that, with subpixel AA on, SourceSansPro is not rendering fine on low res monitors in Brackets UI.

Historically sub pixel AA was turned off to circumvent the [flickering bug](https://github.com/adobe/brackets/issues/1993). Looks like this is being fixed in CEF2171 as I am unable to repro this bug anymore.

I tested this PR on variety of hardware (retina, non-retina, external monitor) and the code rendering is a little crisper now on MAC.


nethip:
Fixes https://github.com/adobe/brackets/issues/11013, https://github.com/adobe/brackets/issues/4640 and https://github.com/adobe/brackets/issues/10169

nethip:
@busykai @MarcelGerber Would you mind reviewing this PR? This is targeted for 1.4.

src/brackets.js
```diff
@@ -142,7 +142,42 @@ define(function (require, exports, module) {
...
+
+    // Define a preference for font smoothing mode on Mac.
+    // By default fontSmoothing is set to subpixel-antialiased
+    // for the text inside code editor. It can be overridden
+    // to "antialiased", that would set text rendering AA to use
+    // gray scale antialiasing.
+
+    if (brackets.platform === "mac") {
+
+        PreferencesManager.definePreference("fontSmoothing", "string", "subpixel_antialiased", {
```
MarcelGerber: Please replace the underscore in `subpixel_antialiased` with a hyphen.

nethip: @MarcelGerber Good catch! Will do that.

src/styles/brackets.less
```diff
@@ -53,18 +53,14 @@ html, body {
...
+    
     &.platform-mac {
-        -webkit-backface-visibility: hidden;
-        backface-visibility: hidden;
+        // Use gray scale antialiasing for UI. Code view editor-holder
+        // overrides this to use subpixel antialiasing, which then
+        // can be overridden by setting "fontSmoothing" preference to
+        // "antialiased". Gray scale AA is used for UI has SourceSansPro
+        // font does not look good with subpixel AA well on low res monitors.
+        -webkit-font-smoothing: antialiased;
```
MarcelGerber: This setting applies to `html, body` while the pref applies to `#editor-holder`. Is this intended?

src/brackets.js
```diff
@@ -142,7 +142,42 @@ define(function (require, exports, module) {
...
+
+            $("#editor-holder").css("-webkit-font-smoothing", aaType);
+
+        });
+
+        AppInit.htmlReady(function () {
+            // The code editor's text anti-aliasing should be defaulted to sub pixel
+            // antialiasing on mac
+            if (brackets.platform === "mac") {
+                $("#editor-holder").css("-webkit-font-smoothing", "subpixel-antialiased");
```
MarcelGerber: Presumably, it's easier (and more semantically correct) to just add this as an extra `#editor-holder.platform-mac` selector to the stylesheet.
Chrome may even optimize that case.

MarcelGerber:
Looks good to me otherwise.

nethip:
@MarcelGerber Thanks for taking a look at this PR!  I will incorporate the changes you have suggested and update this PR.

src/styles/brackets.less
nethip: Yes that is correct. The entire UI would have gray scale anti aliasing except for the code editor. Subpixel AA for the UI was not looking good. It exposed some artifacts which was making the entire editor look out of place. I will update the PR with my findings and PR, that will help understand the scenario. Thanks!

nethip:
After analyzing how fonts render on both retina and non-retina and also comparing the rendering with other apps I have come to the following conclusion.
 
1)      Editors like Sublime Atom and Visual Studio code enable sub pixel rendering by default, irrespective of whether the display is retina or non-retina.
2)      Some editors, like Sublime, allow users to override this default subpixel rendering. In sublime, “font_options”preference can be used to control the code rendering to use sub pixel AA or gray scale AA. This override is limited to just the code area alone. The rest of the UI is still using sub-pixel rendering.
 
So the logical conclusion I have arrived at was that, we should be enabling sub-pixel rendering by default and give an option to user to override this behavior. So I went ahead and added a new preference “fontSmoothing”, which would take “subpixel-antialiased” or “antialiased”  as the values for this preference.
 
Now with this in mind I went ahead and enabled subpixel-AA for the entire app(apply –webkit-font-smoothing:”subpixel-antialised” to the body tag). After that I looked at how the overall rendering looked like. The code area was looking a bit more crisp.
 
But the problem was with other parts of the UI like project tree, find in files pane, JS linting pane e.t.c. I could immediately see extra artifacts being rendered. The following screenshot was taken on an external monitor with a resolution of 1680 x 1050, connected to Mac Pro, with subpixel-AA on.
 
image

image

 
 
This is the same rendered with gray pixel on AA, which is the current scheme on master.
 
image

image

 
 
I am putting the screenshots of Brackets with subpixel-AA on and off. The first one is with subpixel-AA on. Just by glancing we can see that the first one(subpixel-AA on) has some extra pixels which is making it look a little out of place.
 
image

image

 
I tried substituting the UI font SourceSansPro with system fonts like Lucida Grande and Helvetica Neue and I saw that using system fonts, the fonts render a little better. Looks like SourceSansPro is not rendering fine with text color #ffffff on dark backgrounds after turning on subpixel-AA. I checked on how this is with other apps and found out that Sublime, Atom and Visual Studio Code use system default fonts, which render relatively well for subpixel-AA on cases. Also none of them uses text with color #ffffff and that could be the reason why they text is rendering fine in other apps as white text on black background is producing artifacts even with other apps.
 
So with the above test results, I went ahead and added subpixel-AA only to the code area. (to #editor.holder alone). Now the following screenshot is the one where subpixel-AA is enabled only to the code area.
 
image

src/brackets.js
nethip: @MarcelGerber I could do a `.platform-mac #editor-holder` but that will make AA non-configurable. The idea is to make the `subpixel-antialiased` controlled through a preference. 

I am thinking of creating another selector and attach/unattach it to `#editor-holder` based on the preference. Does this make sense? Or the current scheme of things is better?

nethip:
Tagging @larz0 for his comments on enabling sub pixel AA only for the code area.

src/brackets.js
busykai: That'd be nicer, I think.

busykai:
@nethip, I'm wondering if `brackets.js` is the right place to do this. Perhaps either `view/MainViewManager.js` or `view/WorkspaceManager.js` are more appropriate (`MainViewManager.js` more so, `WorkspaceManager.js` manages resizing, really). Both react to `htmlReady` and manipulate `#editor-holder`.

Also, if moved to another module, I'd reduce the use of string literals and define modules vars for values and names.

src/brackets.js
busykai: Shouldn't this line set the pref value instead of presumed default?

MarcelGerber: We don't have access to the prefs at this early point, IIRC

src/brackets.js
```diff
@@ -142,7 +142,42 @@ define(function (require, exports, module) {
...
+            description: Strings.DESCRIPTION_FONT_SMOOTHING,
+            values: ["subpixel-antialiased", "antialiased"]
+        });
+
+        PreferencesManager.on("change", "fontSmoothing", function () {
+
+            var aaType = PreferencesManager.get("fontSmoothing");
+
+            // For any other value, set this to subpixel-antialiased
+            if (aaType !== "subpixel-antialiased" && aaType !== "antialiased") {
```
MarcelGerber: Taking another look at this line, I think we could also make it a Boolean pref, as we only really have two values (and I don't think there will be more in the future)

MarcelGerber:
Maybe the place where other font prefs like `font.fontSize` are defined is best.
Can't look it up right now.

src/brackets.js
busykai: You're right, only view state is ready by then. My point, though, that it probably should be done when the prefs are ready.

nethip: @busykai I evaluated setting the value when the prefs are ready. But I am thinking that experience would then be the user first sees the gray antialised text and then sub pixel AA text. The might give a feeling of a flicker effect. And that is why I did not go with that approach. Correct me if I am wrong.

src/brackets.js
nethip: Sure! we could do this to be a bool variable. I only went this approach as I had earlier plans to add `none` to the list. Now I am thinking should we add this option, if not we could go with the `bool` approach.

nethip:
@MarcelGerber I could move the preference to some place where `font` prefs are getting initialised.

nethip:
@busykai I think you are right about the code to be present in `MainViewManager.js`. I will move it to that place.

nethip:
@busykai @MarcelGerber While I was unit testing with this change on MAC, the text now looks crisper with white text on black background. However it looks a little fuzzy on black text on white background. After doing some thinking I came up with this idea of why not add a new entry in the metadata section of themes. That way theme authors can decide on whether they would want the AA to be sub pixel or grayscale. That way we could then enable sub pixel AA only for our default dark theme. What do you guise think about this?

@MiguelCastillo Do you think it is a good idea to add another property to the `package.son` that instructs brackets to either enable sub pixel AA or grayscale AA?

CC @ryanstewart

MiguelCastillo:
@nethipn I had implemented this in a PR a while ago. https://github.com/adobe/brackets/pull/8635...  You can checkout the thread.  So my thoughts on this now are that we can just create an extension for this instead of putting it in core.  What do you think?

MiguelCastillo:
I can create the extension and rope you guys in for review if you would like.

nethip:
@MiguelCastillo I checked out your PR #8635. So essentially you are doing what I was thinking about in my latest comment! 

I am curious to know what you would be packaging as an extension. Even if we decide to go with the code, that you are planning to package as an extension, IMO that should be added to the core. My understanding of an extension is something that the app wills still function as is, even if that extension is disabled/deleted. In this case, the rendering itself is broken and I am thinking it must in the core. But this is just my opinion and if we have a strong reason to ship this as an extension, that is fine too.

MarcelGerber:
@nethip I don't have a Mac, so I can't repro myself, but is the rendering on dark backgrounds really that much worse than it is on light ones?

The thing is, I can think of scenarios like "The font looks awful with your theme", where you can't really blame the developer because he presumably either didn't know or didn't think about including the antialiasing pref in his theme (maybe just because they don't have a Mac themself).

MiguelCastillo:
@nethip we can always put things in core...  Not a problem.  But the reason this was not put into core in the first place is because it's **not** a setting that applies across platforms. So it didn't seem like a good approach for us to add a setting that only worked on Mac and would potentially cause confusion when users in different platforms wouldn't see a difference when switching the setting.  This whole trouble was not worth our effort of including it in core to solve a relatively niche problem.

I am not sure what you mean by rendering is broken...  If anti-aliasing is breaking rendering, then we fix it and don't make it an option to break it again.

The extension would possibly be JS that exposes a way to switch the anti-aliasing setting on/off via preferences, a small UI, or even a menu setting similar to settings like `Word Wrapping`.  The extension would have in the repo a large note stating that it is ONLY for Mac.

* I myself run a tiny css that disables font smoothing, which is what I expect the extension to do.

``` css
html, body {
   -webkit-font-smoothing: subpixel-antialiased !important;
}
```

nethip:
@MiguelCastillo 

> we can always put things in core... Not a problem. But the reason this was not put into core in the first place is because it's not a setting that applies across platforms. So it didn't seem like a good approach for us to add a setting that only worked on Mac and would potentially cause confusion when users in different platforms wouldn't see a difference when switching the setting. This whole trouble was not worth our effort of including it in core to solve a relatively niche problem.

We could make the preference name mac specific, something like `font.fontSmoothingMac`, making it easier for the user to figure out that this setting is mac specific. Also now that we have code hints for preferences, we could add `Mac only` in the description.

> I am not sure what you mean by rendering is broken... If anti-aliasing is breaking rendering, then we fix it and don't make it an option to break it again.

Looks like people who are on Mac are having a real bad experience with gray scale AA, especially with white text on black background inside Brackets. This is even becoming a deciding factor for users to either switch with Brackets or migrate to other editors. So this is a very serious problem and I am afraid there is no generic solution to this.

nethip:
@MarcelGerber I understand that the theme developers now have to worry about whether to do this use the new setting, if we decide to add it to the theme. Now I am thinking maybe this is a not a good idea to add it as a theme setting. How about just changing sub pixel AA for dark theme alone? That should fix all the issues that I have put forth above.

MiguelCastillo:
@nethip So with users on Mac complaining that AA is causing a problem, then I would pick the default that works consistently across all platforms... Meaning, pick a default that looks/behaves the same across windows/linux/mac, which is subpixel.  I considered doing subpixel *just* for dark themes, but I strongly believe that providing users with a consistent experience is more important.  I think it would be a bit odd if people saw different font rendering for light themes vs dark themes.

I probably wouldn't provide an option to change it back - but this is less important to argue about so I can go either way.

nethip:
@MiguelCastillo Even I am thinking why don't we just enable it by default.
Users anyway have a preference, which they could use to go back to grayscale AA.

nethip:
@busykai @MarcelGerber  @MiguelCastillo I have incorporated all the code review comments. The PR is up for review now. Please take a look at it when you get a chance. Thanks!

src/styles/brackets.less
```diff
@@ -73,7 +69,9 @@ html, body {
     }
 }
 
-
+.subpixel-aa{
```
MarcelGerber: Why don't we just make it `body.platform-mac #editor-holder`?

src/brackets.js
```diff
@@ -358,8 +358,8 @@ define(function (require, exports, module) {
         }
         
         // Localize MainViewHTML and inject into <BODY> tag
-        $("body").html(Mustache.render(MainViewHTML, Strings));
-        
+        $("body").html(Mustache.render(MainViewHTML, { shouldAddAA: (brackets.platform === "mac"), Strings: Strings }));
```
MarcelGerber: This doesn't work right now, you need to add the `Strings.` namespace in `main-view.html` to the template vars there.
But read my other comment first.

src/styles/brackets.less
MarcelGerber: Oh, I see now. Disregard that.

src/view/ViewCommandHandlers.js
```diff
@@ -266,6 +266,24 @@ define(function (require, exports, module) {
...
     }
 
+
+    /**
+     * Font smoothing setter to set the anti-aliasing type for the code area on Mac.
+     * @param {string} aaType The antialiasing type to be set.It can take either "subpixel-antialiased" or "antialiased"
+     */
+    function setMacFontSmoothingType(aaType) {
+        var $editor_holder  = $("#editor-holder"),
+            hasClass        = $editor_holder.hasClass("subpixel-aa");
```
MarcelGerber: We don't really need this `hasClass`. `add/removeClass` will no-op in case the class is not present.

src/view/ViewCommandHandlers.js
```diff
@@ -515,6 +533,20 @@ define(function (require, exports, module) {
         setFontFamily(prefs.get("fontFamily"));
     });
 
+    // Define a preference for font smoothing mode on Mac.
+    // By default fontSmoothing is set to "subpixel-antialiased"
+    // for the text inside code editor. It can be overridden
+    // to "antialiased", that would set text rendering AA to use
+    // gray scale antialiasing.
+    if (brackets.platform === "mac") {
+        prefs.definePreference("fontSmoothing", "string", "subpixel_antialiased", {
```
MarcelGerber: Still, make it default to `subpixel-antialiased` (undercore -> hyphen)

nethip: My bad. I had changed it at one place but forgot to change it here. I will make the changes here. Thanks!

src/view/ViewCommandHandlers.js
nethip: Sure! I will change this.

src/brackets.js
nethip: This is an real good catch Marcel :clap:

nethip:
@MarcelGerber This is up for review again.

src/styles/brackets.less
```diff
@@ -53,18 +53,14 @@ html, body {
...
-    // It seems that we have to put it here rather than on the animated element in order to prevent the
-    // entire window from flashing.
-    // See: http://stackoverflow.com/questions/3461441/prevent-flicker-on-webkit-transition-of-webkit-transform
-    // Mac-only though, since this would turn off subpixel antialiasing (ClearType) on Windows
+    
     &.platform-mac {
-        -webkit-backface-visibility: hidden;
-        backface-visibility: hidden;
+        // Use gray scale antialiasing for UI. Code view editor-holder
+        // overrides this to use subpixel antialiasing, which then
```
MarcelGerber: add: to use subpixel antialiasing __on Mac__

src/styles/brackets.less
```diff
@@ -53,18 +53,14 @@ html, body {
...
+        // can be overridden by setting "fontSmoothing" preference to
+        // "antialiased". Gray scale AA is used for UI has SourceSansPro
+        // font does not look good with subpixel AA well on low res monitors.
```
MarcelGerber: _Gray scale AA is used for the UI parts which use SourceSansPro font, which doesn't look good with subpixel AA_

src/view/ViewCommandHandlers.js
```diff
@@ -266,6 +266,23 @@ define(function (require, exports, module) {
         }
     }
 
+
+    /**
+     * Font smoothing setter to set the anti-aliasing type for the code area on Mac.
+     * @param {string} aaType The antialiasing type to be set.It can take either "subpixel-antialiased" or "antialiased"
```
MarcelGerber: nit: Add a space: _set. It_

src/view/ViewCommandHandlers.js
```diff
@@ -22,7 +22,7 @@
  */
 
 /*jslint vars: true, plusplus: true, devel: true, nomen: true, indent: 4, maxerr: 50 */
-/*global define, $ */
+/*global define, brackets: true, $ */
```
busykai: `brackets` is not writable here, so it should be `brackets: false`. additionally, it would probably be better if either it was just `brackets` or other globals should be explicitly marked as read-only as well.

src/brackets.js
```diff
@@ -142,7 +142,7 @@ define(function (require, exports, module) {
     
     // read URL params
     params.parse();
-    
+
```
busykai: Could you undo these changes to the whitespace lines? Brackets by default does not trim whitespaces. There's no reason to introduce a change to this. If you have a trim extension, it should probably be disabled for brackets project.

src/nls/root/strings.js
```diff
@@ -743,5 +743,6 @@ define({
     "DESCRIPTION_USE_THEME_SCROLLBARS"               : "True to allow custom scroll bars",
     "DESCRIPTION_LINTING_COLLAPSED"                  : "True to collapse linting panel",
     "DESCRIPTION_FONT_FAMILY"                        : "Change font family",
-    "DESCRIPTION_FONT_SIZE"                          : "Change font size; e.g, 13px"
+    "DESCRIPTION_FONT_SIZE"                          : "Change font size; e.g, 13px",
+    "DESCRIPTION_FONT_SMOOTHING"                     : "\"subpixel-antialiased\" to enable sub-pixel antialiasing and \"antialiased\" for gray scale antialiasing on Mac"
```
busykai: "Mac-only: " should probably be the first words of the description ("on Mac" at the end could be removed then).

Also use of *and* conjunction is confusing. It should be "or" instead.

src/view/ViewCommandHandlers.js
```diff
@@ -266,6 +266,23 @@ define(function (require, exports, module) {
...
+     * @param {string} aaType The antialiasing type to be set.It can take either "subpixel-antialiased" or "antialiased"
+     */
+    function setMacFontSmoothingType(aaType) {
+        var $editor_holder  = $("#editor-holder");
+
+        // Add/Remove the class based on the preference. Also
+        // default to subpixel AA in case of invalid entries.
+        if (aaType === "antialiased") {
+            $editor_holder.removeClass("subpixel-aa");
+        } else if (aaType === "subpixel-antialiased" || aaType !== "antialiased") {
```
busykai: This does not seem right. It does not make sense to have literal `subpixel-antialiased` if its value does not matter. If the value of the setting is not recognized, no action should be taken and a warning should be logged.

Perhaps the entire construct would be better if organized as a `switch` statement.

src/view/ViewCommandHandlers.js
nethip: Sure! I will make the change.

src/view/ViewCommandHandlers.js
nethip: I could change the code not to react to a bad preference but that might lead to unexpected behavior. Let's suppose the current value is `antialiased`, which makes the code editor render using gray scale AA. Say the preference is now changed to `zantialiased`, an invalid entry. At this point of time, if we don't change  AA to default because of the bad preference, on the next launch of Brackets the user would see a rendering different than the one in the previous session. IMO we should default to subpixel AA in case of bad preference.About moving this to a switch statement, that seems to be a good idea and I will move it to a switch statement.

src/brackets.js
nethip: @busykai There was some piece of code here in the earlier commits and i just deleted that piece of code based on the code review comments and that is why it is showing as `blank space removal` change. I could revert this, but I think it is a good idea to remove the extra spaces wherever possible.

nethip:
@busykai @MarcelGerber  Did you get a look at the updated code in this PR? As this is targeted for 1.4, we need to send the strings for translation and I would really appreciate if we could merge this into master as early as possible.

@prafulVaishnav.

rawat11:
@nethip I did some testing with new SP AA, The editor part seemed more readable specially in the black background. But some of the UI elements are not looking crisp (as they are in Windows) ( Eg. Lines and Column info shown on the bottom side of editor) Can we do something on that front ?

screen shot 2015-07-08 at 23 50 19
screen shot 2015-07-08 at 23 49 50

The text in white background looks totally washed away. I think darken them a bit might make them look better.

nethip:
@rawat11 Thanks for testing it out! As discussed on this thread, we would go with subpixel AA for all the themes. Users now have a preference to turn this off. And also rendering is not in our control to darken the text as it is completely managed by `chrome`. AA is the only option using which we can control rendering to some extent. 

Could you change the font from Source Code Pro to some system font and update us, how it is looking with lighter theme. Also would you mind checking the same in other editors like Sublime Text and Atom (You might want to change the font and the theme in these editors to look like what it is in Brackets so that we have an apples to apples comparison)

Regarding the UI elements: No change has been done there, so the rendering should be as it is on `master`.

rawat11:
@nethip I tested with some system fonts and it did not look much of the difference on Black background
screen shot 2015-07-09 at 02 45 49

With lighter theme (AA on)

screen shot 2015-07-09 at 02 46 53
screen shot 2015-07-09 at 02 47 48


With Sub pixel AA on

screen shot 2015-07-09 at 02 48 32


White theme looks better with SP AA

Also, I tried the same on Atom, but i could only change the theme, could not find the option for font

screen shot 2015-07-09 at 02 50 45

I guess Atom has SP AA on by default


According to me White themes aren't looking that good with many of the system fonts but changing it to SP AA makes it look a bit better but still not unto the par. With Black it looks good and SP is definitely a good option to move to.

For the UI, I think we have maintained Gray AA but is there no option to make some change on this, it would only add to readability ad well as rendering

prafulVaishnav:
LGTM.. Merging..

prafulVaishnav:
@nethip I think there are some conflicts.. Can you please resolve it..

nethip:
@prafulVaishnav I will resolve the conflict and merge the PR.
I have added this [waffle card](https://github.com/adobe/brackets/issues/11370) to track adding unit tests to this PR.

nethip:
@prafulVaishnav I have resolved all the issues and this PR is now merged :smile:

PR #11254, opened 2015-06-12
Instead of clearing the search on tab change, update the search across the tab

prafulVaishnav:


abose:
I think there was some reason behind why the search string is cleared by brackets on every tab switch  #5856 #7388. I have marked this to review to be sure before going ahead with this.

abose:
Closing as the decision to clear the search string on tab switch was an XD choice.

PR #11245, opened 2015-06-10
Allow JavaScript Code Hints in PHP Files

sprintr:
Fixes #11240

PR #11263, opened 2015-06-13
Allow JavaScript Code Hints only in its context

sprintr:
Fixes a regression I made in #11245 where JavaScript code hints were globally available in a PHP document. This will make sure they are only available inside `script` tags.

### DIFF
--- a/src/extensions/default/JavaScriptCodeHints/main.js
+++ b/src/extensions/default/JavaScriptCodeHints/main.js
@@ -56,6 +56,8 @@ define(function (require, exports, module) {
         noHintsOnDot   = false, // preference setting to prevent hints on dot
         ignoreChange;           // can ignore next "change" event if true;
 
+    // Languages that support inline JavaScript
+    var _inlineScriptLanguages = ["html", "php"];
     
     // Define the detectedExclusions which are files that have been detected to cause Tern to run out of control.
     PreferencesManager.definePreference("jscodehints.detectedExclusions", "array", [], {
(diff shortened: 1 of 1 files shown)


src/extensions/default/JavaScriptCodeHints/main.js
```diff
@@ -341,10 +341,11 @@ define(function (require, exports, module) {
     };
 
     /**
-     * @return {boolean} - true if the document is a html file
+     * @return {boolean} - true if the document is a html or php file
      */
     function isHTMLFile(document) {
```
abose: Function name is now misleading. maybe change including usages to
isHTMLOrPHPFile(document)
or alternatively move the check to a more localized place.

src/extensions/default/JavaScriptCodeHints/main.js
```diff
@@ -56,6 +56,9 @@ define(function (require, exports, module) {
         noHintsOnDot   = false, // preference setting to prevent hints on dot
         ignoreChange;           // can ignore next "change" event if true;
 
+    // Languages that support inline JavaScript
+    var _inlineScriptLanguages = ["html", "php"];
```
abose: Would it make sense to define this as a preference?

sprintr: Don't think so. These are the only two languages that support inline JS.

sprintr: In case someone wants to disable JavaScript code hints in PHP, they can use language/php layer to disable them.

abose: jsp/asp I remember can do similar js mixin as PHP.

sprintr: @abose jsp/asp files will be highlighted as [html files](https://github.com/adobe/brackets/blob/master/src/language/languages.json#L47).

TomMalbran: Maybe is something that could be defined as part of the language in the LanguageManager?

abose:
Looks good. Merging.

PR #11275, opened 2015-06-15
Fix unittests

sprintr:
Fixes errors in unit tests caused by #11201.

CC: @abose


abose:
Fixed, merging.

sprintr:
@abose Did all of them pass this time?

abose:
Yes all tests passed. 
If you noticed that `live Preview` tests failed, that intermittently happens, usually some timeout issues, but will pass eventually on rerums.

sprintr:
@abose Only 1 live preview test passed for me at first run.

abose:
yes i too get that at times, but reruns usually does the trick. Might be some timing issue.

PR #11302, opened 2015-06-23
Fixed a few typos

sprintr:
Fixes a few typos in preferences description

CC: @abose


abose:
Thanks @sprintr . merging.

rawat11:
@sprintr Could you please check these strings (Do they hold specific meaning ?)
- "True to allow unfiltered for in" (should it be translated as "True to allow unfiltering for inbound elements"?)
- "True to allow stupidity" 


@abose Can you check as well

sprintr:
> "True to allow unfiltered for in" (should it be translated as "True to allow unfiltering for inbound elements"?)

`True to allow unfiltered for in` seems fine to me. [Reference](https://jslinterrors.com/option-forin)

> "True to allow stupidity"

This can be changed to `True if blocking ('...Sync') methods can be used.` [Reference](http://jslint.it/lint.html)

rawat11:
@sprintr Thanks. I do not have much idea on this.
If you are fine with first. then let it be like that. For second we can change it, leaving it to you.

sprintr:
In the review of #11318, All of `True` were changed to the JavaScript Boolean `true`, so should we do the same for default locale?

@rawat11 @abose @MarcelGerber

abose:
@sprintr seems logical to use js boolean for me too.

Issue #11056, opened 2015-05-05
Improve Preferences Experience


ryanstewart:
# Background
Until now we've kept the preferences menu item disabled because we'd planned on doing a more robust UI around preferences. But in the few years that Brackets has been around, more and more editors have adopted the JSON file as a preference and in general our users seem comfortable with that approach.

# Proposal
* Enable the Preferences menu item in the default menu on Mac and Windows
* Show *all* of the available preference items in the brackets.json file with their default values. (So it's easier for users to change them)
* Make sure the [preferences documentation](https://github.com/adobe/brackets/wiki/How-to-Use-Brackets#preferences) is up to date

ryanstewart:
@prafulVaishnav @nethip @rroshan1 Would it be possible to add a `"documentation"` property to `brackets.json` that points to our documentation?

I know JSON doesn't support comments so we can't show something like:

`\\ Go here for documentation` 

But could we do something like:

`{
   "documentation":"https://github.com/adobe/brackets/wiki/How-to-Use-Brackets#preferences"
    "steelbisondev.fixmyjs.on_save": false,
    "linting.collapsed": true,
    "linting.enabled": false,
    "BracketsNewProjectExtension.newProjectsFolder": "/Users/ryanstewart/Library/Application Support/Brackets/extensions/user",
    "livedev.multibrowser": true,
    "me.drewh.jsbeautify.on_save": false,
    "themes.theme": "light-theme",
    "fonts.fontSize": "11px",
    "wordWrap": false
}
`

And just ignore that property? Is that clear enough for users who want to know more about preferences?

nethip:
:+1: 

@ryanstewart About documentation, we could do that.

sprintr:
I am going to submit a PR to allow code hints for preferences next week.

@ryanstewart I was thinking about adding a little description of each preference beneath it when it is active.

ryanstewart:
@sprintr yes! The code hints PR would be awesome!

rroshan1:
@sprintr :+1:

sprintr:
The PR for code hints is #11130

TomMalbran:
Should we consider shipping with a JSON lint extension and warn users that the preference will not work if there are errors in the file.

I also find it as a bad user experience that after the popup saying that there is an error in preference, fixing that error, saving the file, it doesn't switches to the preference file and I have to reload to trigger it.

mackenza:
[...]

PR #11307, opened 2015-06-24
Opening preferences brackets.json in split view along with default preferences

nethip:
With this PR, upon choosing `Debug > Open Preferences file`, preferences file `brackets.json` will open in the second pane of split view while the default preferences will load in the left-pane. The default preferences are generated on the fly by iterating through all valid preferences and writing into the default preferences file, `defaultSettings.json` created inside `brackets` preferences folder. Here is the [work card](https://github.com/adobe/brackets/issues/11056) for the same

This is the sample screenshot of how the split view looks like.

image

I am going to add unit tests for these. Meanwhile I would like to get some eyes on the changes and also make some code changes based on the review comments.

@sprintr @MarcelGerber @abose Please review this when you get a chance.Thanks!

cc @ryanstewart 
### Pending Tasks
- [ ] ~~Add unit tests~~: Adding unit tests to this PR is not straight forward as we are using shell apis for this PR and they are not working with our unit test framework. I have added a [waffle card](https://github.com/adobe/brackets/issues/11370) to track this specific task.


src/extensions/default/DebugCommands/main.js
```diff
@@ -48,6 +51,12 @@ define(function (require, exports, module) {
     
     var KeyboardPrefs = JSON.parse(require("text!keyboard.json"));
     
+    // default preferences file name
+    var DEFAULT_SETTINGS_FILENAME = "defaultSettings.json";
```
MarcelGerber: We really only use the term "preferences", so we should do so here, too.

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +272,467 @@ define(function (require, exports, module) {
...
+            // it to rebuild the UI.
+            WorkingSetView.refresh(true);
+        }
+
+        CommandManager.execute(Commands.FILE_OPEN, { fullPath: prefsPath, paneId: "second-pane"});
+
+    }
+
+    function _isSupportedPrefType(prefType) {
+        if (prefType !== undefined &&
```
MarcelGerber: You don't really need the check for `undefined` here.
It makes the method harder to read imho.

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +272,467 @@ define(function (require, exports, module) {
...
+                        }
+                    }
+                }
+            }
+
+            if (_type) {
+                // preference object's type
+                // is defined. Check if that is valid or not.
+                prefType = _type;
+                if ((prefType !== "number"  &&
```
MarcelGerber: Why don't you use the `_isSupportedPrefType` method here?

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +272,467 @@ define(function (require, exports, module) {
...
+                prefType = "array";
+            } else if (typeof (prefObj.initial) !== "undefined"  ||
+                       typeof (prefObj.keys) !== "undefined") {
+
+                // OK looks like this preference has
+                // no explicit type defined. instead
+                // it needs to be deduced from initial/keys
+                // variable.
+                var _prefVar;
+                if (typeof (prefObj.initial) !== "undefined") {
```
MarcelGerber: You don't need `typeof` to check for `undefined`, a more simple `if (prefObj.initial !== undefined)` suffices.

This is not only true for this particular line, but for the overall changes.

MarcelGerber:
When I just tried this branch, I noticed that `Debug > Open preferences file` no-ops when you're not in Split View and none of the files is already open.

src/extensions/default/DebugCommands/main.js
nethip: Sure! will do this change.

src/extensions/default/DebugCommands/main.js
nethip: OK! will do that. I have added `_isSupportedPrefType` recently and this code was there before.

src/extensions/default/DebugCommands/main.js
nethip: I ran into problems using combination of `undefined` and `"undefined"` and hence followed a single approach to use `typeof`. 

About having this change done across the file, you meant just the usage of `typeof()` for undefined alone or for all the occurences of `typeof`

nethip:
@MarcelGerber That was quick :smile: Thanks for started to test this! I just checked. My bad. I pushed some bad change in one of the last commits that is causing this. For now to see what it looks like, be in the split view and then try `Debug > Open Preferences File`. I will check the code and push in a fix. Thanks!

nethip:
@MarcelGerber I have addressed the code review comments update the PR. Also I have fixed the bug with opening defaultPreferences and brackets.json in split view. Could you take a look at it and let me know if anything else needs to be changed?

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +272,461 @@ define(function (require, exports, module) {
...
+        CommandManager.execute(Commands.FILE_OPEN, { fullPath: prefsPath, paneId: "second-pane"});
+
+    }
+
+    function _isSupportedPrefType(prefType) {
+        if (prefType === "number" ||
+                prefType === "boolean" ||
+                prefType === "string"  ||
+                prefType === "array"   ||
+                prefType === "object") {
```
sprintr: I think we should use an array of supported preference types here and just check whether or not `prefType` is in it.

nethip: Sure! will do that.

nethip:
@MarcelGerber @abose @sprintr I have made the suggested changes. This is up for review.

src/nls/root/strings.js
```diff
@@ -744,5 +744,8 @@ define({
     "DESCRIPTION_USE_THEME_SCROLLBARS"               : "True to allow custom scroll bars",
     "DESCRIPTION_LINTING_COLLAPSED"                  : "True to collapse linting panel",
     "DESCRIPTION_FONT_FAMILY"                        : "Change font family",
-    "DESCRIPTION_FONT_SIZE"                          : "Change font size; e.g, 13px"
+    "DESCRIPTION_FONT_SIZE"                          : "Change font size; e.g, 13px",
+    "DESCRIPTION_OPEN_PREFS_IN_SPLIT_VIEW"           : "False to disable opening preferences file in split view",
+    "DEFAULT_PREFERENCES_JSON_HEADER_COMMENT"        : "/*\n * This is a read-only reference file for what preferences you can\n * override in your own file. Modify your own preferences on the\n * right and if you have any questions, please refer to\n * https://github.com/adobe/brackets/wiki/How-to-Use-Brackets#preferences\n * for more information on how to use preferences inside {APP_NAME}.\n */",
```
abose: A bit of refactoring in the lines of: "This is a read-only file with the preferences supported by brackets. Use this as a reference to modify your own preferences file(brackets.json) opened in the right-hand pane. Refer to https://github.com/adobe/brackets/wiki/How-to-Use-Brackets#preferences  for more information on how to use preferences inside Brackets."

nethip: @ryanstewart What do you think about the new wording?

abose:
There are too many commits in this PR, maybe squash some of them for good measure :)

src/nls/root/strings.js
ryanstewart: Looks good @nethip !

MarcelGerber:
We should make sure nothing can change the contents of this read-only file, no core features (like Replace, Replace in Files) and presumably extensions, too

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,457 @@ define(function (require, exports, module) {
         brackets.app.showOSFolder(dir);
     }
 
+    function _getDefaultPrefsFilePath() {
+        // Default preferences
+        return defaultPreferencesFullPath;
+    }
```
sprintr: Since `_getDefaultPrefsFilePath` is used only once and is also not exported, it might be better to use `defaultPreferencesFullPath` where `_getDefaultPrefsFilePath` is being used.

nethip:
@MarcelGerber It is a good idea to make sure none of the other features can modify this file. Opening the file in read only mode was an attempt to make sure this file does not get modified by other features. We could also make it read only. (For this I will have to add `chmod` to our app shell file system api). Does adding a `chmod` step alone, help in making sure none of the other features will be able to make a change to this file? I do not have any other ideas on how to make sure other features don't make any change to `defaultPreferences`?

src/extensions/default/DebugCommands/main.js
nethip: I will remove this entry.

nethip:
@abose I will do an `git -i rebase` once you guys are OK with the PR.

MarcelGerber:
I just tried using "Replace" in this file and you can indeed change the file contents that way.

The best solution would probably be a read-only flag for the Document, but I suppose it'd be a rather disruptive change at this point.
Thus, I think it's ok as it is right now. At least the most obvious ways cannot be used to alter the file.

src/editor/Editor.js
```diff
@@ -279,9 +281,14 @@ define(function (require, exports, module) {
...
      */
-    function Editor(document, makeMasterEditor, container, range) {
+    function Editor(document, makeMasterEditor, container, range, options) {
         var self = this;
-        
+
+        var isReadOnly = false;
+        if (options !== undefined && typeof (options.isReadOnly) === "boolean") {
+            isReadOnly = options.isReadOnly;
+        }
```
MarcelGerber: You can easily squash these 4 lines to a single `var isReadOnly = !!(options && options.isReadOnly)`.

src/extensions/default/DebugCommands/main.js
```diff
@@ -48,6 +51,13 @@ define(function (require, exports, module) {
     
     var KeyboardPrefs = JSON.parse(require("text!keyboard.json"));
     
+    // default preferences file name
+    var DEFAULT_PREFERENCES_FILENAME = "defaultPreferences.json",
+        SUPPORTED_PREFERENCE_TYPES   = ["number", "boolean", "string", "array", "object"];
+
+    var reComputeDefaultPrefs        = true,
```
MarcelGerber: nit: `recompute` should be all-lowercase (not camelCase). See https://en.wiktionary.org/wiki/recompute

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+            }
+
+            if (_type) {
+                // preference object's type
+                // is defined. Check if that is valid or not.
+                prefType = _type;
+                if (!_isSupportedPrefType(prefType)) {
+                    prefType = "undefined";
+                }
+            } else if (Array.isArray(prefObj)) {
```
MarcelGerber: I don't think there's any case where the `prefObj` is an array.

MarcelGerber:
@nethip I noticed you use a mixture of `_type` and `prefType` with no clear differentiation.
I guess we can improve that.

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+        // if the preference type is valid or not.
+        if (pref && !pref.excludeFromHints && _getObjType(pref) !== "undefined") {
+            return true;
+        }
+
+        return false;
+    }
+
+   /*
+    * This method tries to matach between initial objects
```
MarcelGerber: nit: `match`

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+        if (typeof (prefObj.keys) === "object") {
+            // iterate through the list.
+            var allKeys = prefObj.keys;
+            keysFound = true;
+            for (property in allKeys) {
+                if (allKeys.hasOwnProperty(property)) {
+                    finalObj[property] = prefObj.keys[property];
+                }
+            }
+        }
```
MarcelGerber: The common logic you use 3 times (that is: setting `keysFound = true`, looping over all keys and aggregating them) can be put into a function.

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+
+        return finalObj;
+    }
+
+    function _formatDefault(prefObj, prefName, tabIndentStr) {
+
+        if (!prefObj || typeof (prefName) !== "string" || _getObjType(prefObj) === "object") {
```
MarcelGerber: As stated above, the `prefObj` itself should always be an object.

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+            // return empty string in case of
+            // object or pref is not defined.
+            return "";
+        }
+
+        var prefDescription = prefObj.description || "",
+            prefDefault     = prefObj.initial,
+            prefFormatText  = tabIndentStr + "\t// {0}\n" + tabIndentStr + "\t\"{1}\": {2}",
```
MarcelGerber: Why do you add a tab char (`\t`)?

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+            prefObjType     = _getObjType(prefObj);
```
MarcelGerber: Same here. `prefObj` is _always_ an entry of `getAllPreferences()`, correct?

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+
+        if (prefDefault === undefined && !prefObj.description) {
+            // This could be the case when prefObj is a basic JS variable.
+            if (prefObjType === "number" || prefObjType === "boolean" || prefObjType === "string") {
+                prefDefault = prefObj;
```
MarcelGerber: As `prefObj` is, well, an object, this doesn't really make sense.

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+                }
+
+                if (Array.isArray(_prefVar)) {
+                    // In cases of array the
+                    // typeof is returning a function.
+                    prefType = "array";
+                }
+
+            } else {
+                prefType = typeof (prefObj);
```
MarcelGerber: Still, `prefObj` is always an object.

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+                // Defaulting the preference to false,
+                // in case this is missing.
+                prefDefault = false;
+            } else {
+                // for all other types
+                prefDefault = "";
+            }
+        }
+
+        if ((prefDescription === undefined || prefDescription.length === 0)) {
```
MarcelGerber: Just check `if (!prefDescription)`

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+            prefObjDesc    = prefObj.description || "",
+            prefObjType    = _getObjType(prefObj),
+            hasKeys        = false,
+            tabIndents     = "",
+            numKeys        = 0;
+
+
+        // Generate the indentLevel string
+        for (iLevel = 0; iLevel < indentLevel; iLevel++) {
+            tabIndents = tabIndents + "\t";
```
MarcelGerber: `tabIndents += "\t";`

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+        }
+
+        // Check if the preference is an object.
+        if (_getObjType(prefObj) === "object") {
+            prefObjKeys = _getObjKeys(prefObj);
+        }
+
+        if (prefObjKeys && Object.keys(prefObjKeys).length > 0) {
```
MarcelGerber: We can drop the first condition and move the second one in the if block above.

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,485 @@ define(function (require, exports, module) {
...
+        }
+
+        // Indent the beginning of the object.
+        tabIndents = tabIndents + "\t";
+
+        if (prefObjDesc && prefObjDesc.length > 0) {
+            entireText = tabIndents + "// " + prefObjDesc + "\n";
+        }
+
+        entireText = entireText + tabIndents + "\"" + prefName + "\": " + "{";
```
MarcelGerber: Use `+=`

MarcelGerber:
This code is quite complex for this not-that-complex task.
We should really try to make it more comprehensible.

src/extensions/default/DebugCommands/main.js
nethip: I will change this.I guess I was using `computeDefaultPrefs` before and kind of blindly added a `re`.

src/extensions/default/DebugCommands/main.js
nethip: Well, then I should change the name of the variable. This function is supposed to return an empty string in case it is called by passing an object and generates formatted strings only for non object types.

src/extensions/default/DebugCommands/main.js
nethip: I think the confusion here is again with the variable name. I will change it.

src/extensions/default/DebugCommands/main.js
nethip: Will do this.

nethip:
@MarcelGerber I will address all the code review comments that you have proposed(Thanks!).

When I started with this task I have written barely 50-100 lines to get the task done but it was great amount of unit testing that made me end up with ~800 lines of code. The following are some of the reasons why the code is really looking complicated.

* Strict type checking to make sure preference type and preference default value match.
* Object level formatting, including nesting
* Deducing an object's initial value by looking at various parameters like `keys` and `initial` object. I found out that, in some extensions, authors are sometimes declaring default values in the `initial` object itself. 
* Entire formatting including formatting comments. Otherwise we could have used `JSON.stringify`
* Robustness of every method. Make sure right params are passed and every method can behave independently.

If there are some things, like variable names, that make the code not comprehensible, which do not fall in the above categories, please call them out so thatI can work on those areas.

src/editor/Editor.js
nethip: Yeah I think a stringent type check is not needed here. Did you mean the above condition to be `var isReadOnly = (options && options.isReadOnly)`

src/extensions/default/DebugCommands/main.js
nethip: This method could be called for tested object types as well and that is when we could get a valid array object. 
Consider something like this
```
{
    Pref11: {
        NestedPref: [];
        AnotherPref: false
    }
}
```

So this method is going to be called to determine `NestedPref` type as well and that is the case where we might get an array. I think the name `prefObject` is misleading and is effecting the readability. I will change the name of the variable to something more contextual.

src/extensions/default/DebugCommands/main.js
nethip: This function basically returns formatted text along with tabs.

src/extensions/default/DebugCommands/main.js
nethip: @MarcelGerber Most of the stringent type checks that I have put in here are to make sure that we generate correct code. We definently don't want to end up with a `[object object]` strings in the default preferences file. Consider `if(!prefDescription)` is defined and is set to false. In this case, the condition evaluates to `true` while the check in the first place was to see if this was `undefined` and if so then generate a default value. So we could easily define a `description` property inside a preference and assign a `false` to it. That is good enough to break the logic. I am not saying I have handled all these type of situations, but did try best to handle whatever I can think of. :smile:

nethip:
@MarcelGerber I have incorporated your code review comments into this PR and this is up for review again.

MarcelGerber:
@nethip Travis didn't pass.

nethip:
@MarcelGerber Fixed it.

mackenza:
sorry that I am late to this PR but I tried it out yesterday and wanted to make a suggestion that I don't think has been discussed yet, but if it has, I apologize in advance ;)

I think the editable preferences file (`brackets.json`) should be on the left not the right. It fits in better with the left-to-right workflow most Western languages have and it makes much more sense to me that I am editing on the left and glancing to my right to look at the reference material.

sprintr:
:+1: for @mackenza's suggestion. Another think I would like is to close `defaultPreferences.json` once `brackets.json` is closed, and restore the layout scheme to what is was before opening the preferences.

src/extensions/default/DebugCommands/main.js
MarcelGerber: Well, in my observations, even nested prefs were always objects. Did you really spot a case where this was not the case?

src/extensions/default/DebugCommands/main.js
MarcelGerber: Shouldn't `tabIndentStr` already do that?

src/extensions/default/DebugCommands/main.js
nethip: This is one of the examples.

```
// Sets the tag closing options
	"closeTags": {
		// Close when > of opening tag is typed
		"whenOpening": false,

		// Close when / of closing tag is typed
		"whenClosing": false,

		// An array of tags that when opened has a blank line
		"indentTags": [],

		// An array of tags that should not be auto-closed
		"dontCloseTags": []
	},
```

src/extensions/default/DebugCommands/main.js
nethip: This function is responsible for formatting the code it generates. That, basically, is the principle I have used. For first level preferences like `closeBrackets: false`, `tabIndentStr` is going to be empty.  Consider a preference which has nested tags.

```
"closeTags": {
		// Close when > of opening tag is typed
		"whenOpening": false,

		// Close when / of closing tag is typed
		"whenClosing": false,

		// An array of tags that when opened has a blank line
		"indentTags": [],

		// An array of tags that should not be auto-closed
		"dontCloseTags": []
	},

```

When the same function is called for `"whenOpening: false"`, that is when `tabIndentStr` has some value.

nethip:
@mackenza Sure! How about a preference?

@sprintr We could do that too.

sprintr:
@nethip It would be nice if both of them have preferences. `defaultPreferences.json` should not close for someone who opened Preferences just to read them.

mackenza:
let's go easy on this, though... it doesn't need to be a Cadillac feature. Preferences are typically changed on first install and then rarely changed after that. The PR is a good idea, but complicating it with all kinds of use cases around opening/closing and restoring views seems a bit... too much imo.

src/extensions/default/DebugCommands/main.js
```diff
@@ -257,6 +273,475 @@ define(function (require, exports, module) {
...
+    }
+
+    function _getDefaultPreferencesString() {
+
+        var property,
+            allPrefs       = PreferencesManager.getAllPreferences(),
+            headerComment  = Strings.DEFAULT_PREFERENCES_JSON_HEADER_COMMENT + "\n\n{\n",
+            entireText     = "";
+
+        for (property in allPrefs) {
```
sprintr: Better use `Object.keys(allPrefs).forEach(function (pref) {})` Here. Sorting the preferences here would also be nice.

nethip:
@sprintr I have changed the for() to forEach(). Also now I am sorting the keys.
@mackenza This PR is now updated with a new preference to display the preferences either on the right side or the left side. Please check it out and let me know if this looks fine.

@MarcelGerber Let me know if the changes look good to you. I could then rebase this entire branch to a very limited no of commits before we merge this to `master`.

mackenza:
@nethip I get an error trying to open up the split view with your last changes:

image

Console says:
` Uncaught TypeError: Cannot read property 'removeView' of null` in `view/MainViewManager.js:810`

sprintr:
@nethip I am also facing the issue @mackenza reported above, so am not able to try it out. You may add default values of `closeTags` prefs as well [here](https://github.com/adobe/brackets/blob/prashant/open-prefs-in-split-view/src/editor/Editor.js#L133) like you did for `jslint.options`

nethip:
@mackenza @sprintr Thanks a lot guys for trying out this PR and letting me know about the issue. This kind of crept in with the addition of the new preference to display the user preferences either in the `first-pane` or `second-pane`. I now have updated my branch with the fix. Also I have added default values to `closeTags`. Could you take the latest changes and checkout the feature and let me know if you guys see anything fishy. I really appreciate you guys taking some time in trying this feature out. Thanks!

sprintr:
@nethip :cool: Works just as expected. :+1:  for making the new layout default. The opening/closing/restoring layout scheme issue can be skipped for the moment, it looks more like an extension idea.

nethip:
Since this needs to make into 1.4 we would want to send the strings for translations. I would like to know if there any parts of this PR that needs to be worked on. Or are we good to go with this change, in which case I will go ahead and merge this change.

@MarcelGerber @sprintr @abose @mackenza @

src/editor/Editor.js
```diff
@@ -279,9 +283,11 @@ define(function (require, exports, module) {
      * @param {{startLine: number, endLine: number}=} range If specified, range of lines within the document
      *          to display in this editor. Inclusive.
      */
```
rroshan1: Please add "@param options" in comments for the function.

src/editor/EditorManager.js
```diff
@@ -194,8 +194,8 @@ define(function (require, exports, module) {
      *          to display in this editor. Inclusive.
```
rroshan1: Please add "@ param editorOptions" in comments for the function.

rroshan1:
Currently, the comments at the beginning of defaultPreferences.json looks untidy when opened in split view,
current
I would suggest a cleaner 
changed

rroshan1:
@nethip Everything else looks good from my side. :+1: 
I have run unit tests and scenario tests and all have passed.
I can merge it once the comments are addressed.

src/nls/root/strings.js
```diff
@@ -744,5 +744,9 @@ define({
     "DESCRIPTION_USE_THEME_SCROLLBARS"               : "True to allow custom scroll bars",
     "DESCRIPTION_LINTING_COLLAPSED"                  : "True to collapse linting panel",
     "DESCRIPTION_FONT_FAMILY"                        : "Change font family",
-    "DESCRIPTION_FONT_SIZE"                          : "Change font size; e.g, 13px"
+    "DESCRIPTION_FONT_SIZE"                          : "Change font size; e.g, 13px",
+    "DESCRIPTION_OPEN_PREFS_IN_SPLIT_VIEW"           : "False to disable opening preferences file in split view",
+    "DESCRIPTION_OPEN_DEFULT_PREFS_IN_FIRST_PANE"     : "False to open preferences in right pane/bottom pane",
```
mackenza: spelling error... should be `DESCRIPTION_OPEN_DEFAULT_PREFS_IN_FIRST_PANE`

mackenza:
@nethip I am really not crazy about the preference name for whether the editable window is left or right.

`alwaysOpenUserPrefsinSecondPane` is a real mouthful and semantically questionable.  A simple fix might be to remove the "always". It's not needed. We don't have it in the other preference related to this: `openPrefsInSplitView` so I feel it's redundant.

So, in the interest of getting this feature merged, I propose we simply go with:

`openUserPrefsInSecondPane` (notice I also camel-cased the `I` in `In`

src/nls/root/strings.js
```diff
@@ -744,5 +744,9 @@ define({
...
+    "DEFAULT_PREFERENCES_JSON_HEADER_COMMENT"        : "/*\n * This is a read-only file with the preferences supported by brackets. \n * Use this as a reference to modify your own preferences file(brackets.json)\n * opened in the right-hand pane. \n * Refer to https://github.com/adobe/brackets/wiki/How-to-Use-Brackets#preferences \n * for more information on how to use preferences inside Brackets.\n */",
```
mackenza: now that we have a preference for which pane the default prefs opens in, we cannot say `opened in the right-hand pane.` I suggest:

`opened in the other pane.`

nethip:
@mackenza I agree with you on changing the preference name. Just did it. Also changed the description variable name along with other string changes.

@rroshan1 I incorporated your changes as well.

MarcelGerber:
There's still a merge conflict.

nethip:
@MarcelGerber I am on it. Looking into the merge conflict!

nethip:
I have created another PR which is basically a rebase of this PR, as this PR has some 30 odd commits. Would you mind having a quick look and merging it. Thanks!

nethip:
@MarcelGerber Here is the new [PR] (https://github.com/adobe/brackets/issues/11369) with the rebased commits.

nethip:
Since there are other PRs that are dependent on this I have gone ahead and merged this PR. Please feel free to call out any specific pieces of code that need to be reworked on. I will make sure that is addressed.

A big thank you to @MarcelGerber @sprintr @mackenza @abose @rroshan1 for providing valuable feedback because of which this PR is in a very good shape. Thanks!

Issue #10289, opened 2014-12-29
Can i see live preview on my ipad ?


MickeyBadBad:
I know there is two way to get this .
1,use something like Air Display .
2,do something like apache allow deny mechanism , and use browser in ipad .
i want to know , does there have option 2 ? if not , Could you guys make it ? ( it's will be a good function i think ~ )

webRat:
You can already do this now without any additional work? Your iPad has to be on the same lan as your machine using Brackets. When it browses to (localhost):(port)/file, you'll browse to (ip address of machine running backets):(port)/file You have to keep live preview running on the machine running backets at all times or the connection will break?

ILYG:
the problem is the port start from Brackets is only allowed by localhost , i can't find how to change it to allowed all ip requests .

webRat:
No, your machine has an internal ip.

Assuming you're on a mac, go to Apple->System Preference->Network->Look at your wifi connction and it should say, "Wi-Fi is connected to {{ Network name }} and has the IP address {{ Something like 192.168.1.x. }}"

Then go to http://192.168.1.x/path/to/your/files on your ipad.

webRat:
Nevermind, I see what you mean now. If you're not familiar with setting up servers and such, perhaps CodeKit is the way to go for you.

ILYG:
I just had a pad , not Mac @-@ . and i think the setting up things i'm good at it . can you tell me what can i do specificly , please ?

redmunds:
@busykai Is this a possibility for Live Preview MultiBrowser? If not, then feel free to close it.

busykai:
@redmunds yeah, actually, i have a patch to enable remote connectivity to live preview (as an option). i will submit a PR to start the discussion. it will have its implications though such as security, need for manual launch etc.

busykai:
@redmunds, @ILYG there's a functional implementation for *multi-browser LivePreview only* in `kai/ld-remote-connectivity`.

JKAussieSkater:
Bump!

PR #11338, opened 2015-07-02
Enable setting the LivePreview server address in Brackets preferences.

tony2nite:
Add functionality to configure the LiveServer address in Brackets preferences file. This enables setting like this to let you access the LiveServer across a local network:

```
"staticserver.address": "0.0.0.0"
```

Addresses #6424

### DIFF
--- a/src/extensions/default/StaticServer/StaticServer.js
+++ b/src/extensions/default/StaticServer/StaticServer.js
@@ -46,6 +46,10 @@ define(function (require, exports, module) {
         description: Strings.DESCRIPTION_STATIC_SERVER_PORT
     });
 
+    _prefs.definePreference("address", "string", "127.0.0.1", {
+        description: Strings.DESCRIPTION_STATIC_SERVER_ADDRESS
+    });
+
     /**
      * @constructor
      * @extends {BaseServer}
@@ -135,8 +139,11 @@ define(function (require, exports, module) {
         }
 
         var port = sanitizePort(_prefs.get("port"));
+        var address = _prefs.get("address");
 
-        this._nodeDomain.exec("getServer", self._root, port)
+        console.error('addresssssss', address);
+
+        this._nodeDomain.exec("getServer", self._root, address, port)
             .done(function (address) {
 
                 // If the port returned wasn't what was requested, then the preference has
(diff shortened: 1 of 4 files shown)


nethip:
@tony2nite You need to sign [Brackets CLA](http://dev.brackets.io/brackets-contributor-license-agreement.html).

@busykai Would you mind taking a look, when you get a chance?

tony2nite:
@nethip I've signed-up. Twice actually. Let me know if it's not registering correctly.

nethip:
@tony2nite The CLA checks seem to be have passed. The errors now are JS linting errors. Please fix the linting errors and update this PR. Thanks!

tony2nite:
Hi @nethip. Fixed the linting errors and updated. Looks green to me.

nethip:
Thanks @tony2nite . Let us wait to hear from @busykai.

busykai:
oh, right. this has been a long-standing issue. thanks @tony2nite for bringing this up!

it has been discussed more recently in #10289. i actually created a branch which implemented the necessary changes, it's here: `kai/ld-remote-connectivity`. in that branch i'm just switching the implementation to `INADDR_ANY`, non-configurably (actually that's about the only piece that is missing in that branch). i'm wondering if we could reconcile the two. it seems that i've created a PR #10517 at some point, but it's gone now and github does not let me to create another one from the same branch. anyways, here is the [comparison](https://github.com/adobe/brackets/compare/kai/ld-remote-connectivity).

things i dislike in this PR:
1. address should not be passed as is from configuration to `listen`. there are really two options here: you either listen to `INADDR_ANY` (`0.0.0.0`) or to `localhost`. a more elaborate third choice should be expressed in terms of the physical/virtual interfaces to listen to, but it's not necessary.
2. i doubt this is functional in the first place since the url-to-document translation wouldn't work for the remote URLs. listening to `INADDR_ANY` is not enough.
3. multi-browser LivePreview implementation is not addressed.

it seems to me that the best course of action would be to take `kai/ld-remote-connectivity` as the base and make it configurable (with only one `boolean` switch indicating if listen to all interfaces or loopback only, defaulting to loopback). i will be able to finish this up and create a pull request in a couple of weeks or so. on the other hand, if @tony2nite is willing to take the work done in `kai/ld-remote-connectivity` and finish it up, that would be cool as well!

JKAussieSkater:
Bump!

josezulu:
What is the status on this?

Issue #11333, opened 2015-07-01
Find/Replace dialog closes on cursor out


nlourenco:
If I have a block of code in which I need to batch replace within a folder of files, it is not possible since if a user clicks off the Find/Replace dialog in order to copy and paste his/her code the dialog box disappears.

* File > Replace In Files
* Click off dialog and onto working file
* Dialog disappears

If there is an alternative method please let me know, otherwise a close box with a scalable field for the Find/Replace dialog would be more appropriate feature.

LKona:
I can still reproduce this issue. I am using Release 1.10 development build 1.10.0-0 (master 04428819f) on a Mac.

Issue #11261, opened 2015-06-13
Search bar ("Ctrl + F")


mohagan35:
It would be great if after pressing "Ctrl + F" that the search bar remains in place after clicking on the script you're working on. 

Sometimes I want to edit code in a script where a particular string occurs. It is time consuming to have to keep pressing "Ctrl + F" and typing in the input search string.

At the very least the previous string you searched for should not be removed from the search bar.

PR #11346, opened 2015-07-04
Resolves issues - 11333 and 11261

gitexperience:
11333 ( https://github.com/adobe/brackets/issues/11333 ) and 11261 ( https://github.com/adobe/brackets/issues/11261 )

Closes the modal bar only when escape is pressed and modal bar is having focus .This removes the problem in find operations as losing the focus of the bar will not close the modal bar until 'escape' key is pressed..

### DIFF
--- a/src/widgets/ModalBar.js
+++ b/src/widgets/ModalBar.js
@@ -239,9 +239,20 @@ define(function (require, exports, module) {
         }
     };
     
+    /**
+     * If autoClose is set, detects when something other than the modal bar is getting focus and
+     * dismisses the modal bar. DOM nodes with "attached-to" jQuery metadata referencing an element
+     * within the ModalBar are allowed to take focus without closing it.
+     */
     ModalBar.prototype._handleFocusChange = function (e) {
         if (this.isLockedOpen && this.isLockedOpen()) {
             return;
+        }
+        
+        var effectiveElem = $(e.target).data("attached-to") || e.target;
+        
+        if (!$.contains(this._$root.get(0), effectiveElem)) {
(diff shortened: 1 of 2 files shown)


abose:
@gitexperience  thank you for contributing to brackets. You could join the [brackets slack channel](https://brackets-slack.herokuapp.com/.https://brackets-slack.herokuapp.com/) and [brackets dev google groups](https://groups.google.com/forum/#!forum/brackets-dev) for discussions with the brackets community.

src/widgets/ModalBar.js
```diff
@@ -239,21 +239,10 @@ define(function (require, exports, module) {
         }
     };
     
-    /**
-     * If autoClose is set, detects when something other than the modal bar is getting focus and
-     * dismisses the modal bar. DOM nodes with "attached-to" jQuery metadata referencing an element
-     * within the ModalBar are allowed to take focus without closing it.
-     */
     ModalBar.prototype._handleFocusChange = function (e) {
```
abose: This is a generic modal bar focus handler and may not be the right place to do this change. 
What you might want to do is to find out the specific modal bar for find/replace [see src/search/findbar.js ] and create a isLockedOpen() function to lock focus handling.

gitexperience:
PR updated,!! now it is localized to find bar ..

abose:
This is now implemented as part of instant search.
@gitexperience thanks for the PR.
Starter bugs- issues and other features in trello can be picked up if you wish to contribute more. :+1:

nlourenco:
Issue #11333 still persists

Issue #11356, opened 2015-07-07
Code folding not working well with multi-line-attributes tag in html


liuxiaole:
Steps to repro:

Write some HTML like below:
```html
<div title="very long content
with 
line breaks">
    <p>abcdefghijk</p>
</div>
```
or
```html
<div data-options="some options"
     data-tips="some tips">
    <p>abcdefghijklmn</p>
</div>
```

Fold the `<div>` tag and then unfold it. Now if you'd like to fold this tag again, you'll find that it doesn't work.

OS: Mac OS X 10.10.3 & Windows 7
BracketsVersion: 1.3.0-16022

thehogfather:
@prafulVaishnav I've got a pretty good idea why this is happening - happy to submit a pr in the next couple of days if no one is assigned.

prafulVaishnav:
@thehogfather Thanks. Currently, one is assigned so Please go ahead and submit a PR.

thehogfather:
@prafulVaishnav you can check out and test PR #11366 as a fix for this issue.

PR #11366, opened 2015-07-09
addresses #11356 xml start tags spanning multiple lines

thehogfather:
collapsing and expanding fold ranges in xml or html elements where the starttag spans more than one line now works correctly.


abose:
@thehogfather Sorry for holding up this pr. Missed it for 1.5. Moved to 1.6.
There seem to be merge conflicts, can you resolve that ?

PR #11375, opened 2015-07-11
Instant Search.

abose:
**Instant Find In Files Search**
This change deals with moving the search logic to Node with instant search results. The overall changes covered are as follows:-
0. The search cache and the search logic itself now live in Node process.
1. Increased search performance and instant search - see the results as you type.
2. Brackets UI is unblocked/responsive while a search is in progress.
3. Preference entries `findInFiles.instantSearch` and `findInFiles.nodeSearch` for enabling.
4. If Node search fails, the Brackets search will be used as fallback until a project switch occurs and Node responds.
5. Ordering of search results made better: Now not only the currently opened file gets precedence (very first) but also the files in working set. 
Order of precedence : Open file -> files in working set -> remaining files.
6. File watchers and document change events captured and propagated to Node.

Pending
- [x]  Fix unit text defects.
- [x]  Clear all debug console.logs.
- [x]  Health logs to log node failures, number of files in search scope.

Contributors: @rroshan1 @prksingh @prafulVaishnav @nethip .


rroshan1:
I believe this awesome "Instant Find In Files Search" is going to be a hit with users and help bring more developers to Brackets!! We have come a long way from the slow search perception to instant search which no other editor provides this well.
Guys, please use this feature and do let us know your feedback (experience + issues).
Also, you can review the code and share your comments.

rroshan1:
Here is a gif of how this feature looks!!
instant_search

mackenza:
@abose I checked out `upstream/abose/instant_search` and gave it a try. It worked great on my Brackets repo and I wanted to see it work on others, so I switched projects and tried again and find-in-files doesn't trigger until I press enter. I went back to brackets and tried again and it's not working there either.

I added `findInFiles.instantSearch: true` and `findInFiles.nodeSearch` to my `brackets.json` but it didn't change.

Am I doing something wrong?

rroshan1:
@mackenza It is intended!! Actually, Node caches the project files at project load, so during that 1 minute or so, if someone launches a find in files search, Brackets fall back to its older search (without instant search and acts on hitting enter). But once caching is complete, it starts with instant search!!

mackenza:
@abose is there some way we could give a visual indication of when the indexing is done?

mackenza:
@abose I have an older laptop I run Linux on (Dell Latitude E6500) and I can't seem to get this working. When I switch projects, I wait for a good long time (recognizing that my slow laptop means slower chaching time) but I never seem to get instant search. I look at `top` on the machine and I would expect to see brackets-node with some cpu trying to cache the project, but I don't see that either.

abose:
@mackenza could you set the preferences entries to true, restart brackets, and then try the search.
What is the project size you are on? [number of files, size in MB]
The first caching of the files will take a bit of time, but search will still function as normal on pressing enter.  
Also can you tell us the file caching time: - to get this, open search window, type something and wait till you see the results- instant search will kick in eventually after cache complete.

abose:
@larz0 Could you check this out. @mackenza was asking for some visual indication till the indexing is complete.

mackenza:
@abose I couldn't get anything working, even Getting Started project. So I did a `git checkout master` and then back to `git checkout abose/instant_search` and now it seems to work like a charm. I am getting results back seconds after I switch projects, even large ones (like I cloned v8). Not sure why my install went dead there but it's back now. Great feature.

mackenza:
@abose I was thinking about the UI and have some ideas but wasn't sure of the difference between `findInFiles.instantSearch: true|false` and `findInFiles.nodeSearch: true|false`? I can infer some things, but the combination is interesting... i.e. could I have nodeSearch false if I have instantSearch true?

Depending on the answer to the above and the relationship of the 2 settings I would think a UI like the following would do the trick:

instant_search

So if you ignore the horrible gimp editing, I am proposing a 2-state toggle button like we have for case and pattern already to indicate instant-search either on or off (default should be on imo) and some sort of green-light vs redlight/greylight indicator whether the current project has finished caching or not.

I am sure @larz0 can make it look good ;)

rroshan1:
@mackenza : No we cannot have instantSearch set to true with nodeSearch set to false. Node Search will automatically be disabled in case Node does not work properly and Brackets will fall back to older search. But if Node works fine, user has the option of going ahead with Instant Search (using Node of course) or disable it to use traditional search.
+1 to the idea of some switch indicator for Instant Search.

mackenza:
@rroshan1 that is what I thought. It seems to me that `findInFiles.nodeSearch` need not be a public setting since it's setting has no meaningful effect to the user. It should be something you can toggle at a dev level for perhaps debugging purposes but not exposed to the user in the default brackets.json or in a UI.

larz0:
@mackenza thanks for the tag :)

Instant Search should be on by default and the odd time when Brackets is indexing we should indicate that by using our spinner with a label mentioning it's for Instant Search.

brackets_cachingforinstantsearch

Love this feature :+1:

rroshan1:
@larz0 Indicator for caching (not exactly indexing) is good. This means Instant Search will be active after caching is complete. 
But, we should have an indicator for instant search as well since we are providing the user the option to disable it. Also, during the caching, the instant search functionality is disabled and it is enabled only after caching is complete. What sort of indicators do you suggest for it?
**Summary**
State of Instant Search :-
- ON -&nbsp; &nbsp; &nbsp; &nbsp; &nbsp;&nbsp;        By default.
- Disabled- &nbsp; During caching OR user has manually disabled it.

abose:
I think for the time being, we can just keep the spinner on indexing as suggested by @larz0 .
Exposing too many options about internal details to the user could be confusing.

But adding a toggle for instant search could be useful in some very large projects.

mackenza:
Disabling Instant Search seems low volume enough that it can stay as preference set in `brackets.json` and not exposed in the UI. I think that is what @larz0 is indicating.

redmunds:
I took this branch for a quick spin. The **performance is impressive**.

I originally thought that I would not like "instant search", and would turn it off, but it's fast enough that it works! On my machine anyway -- it's good to have this a preference so users with slower machines can turn it off.

I notice that the **sort order is incorrect**. This was a huge bottleneck in my "incremental update" branch, so I'm curious how much this will effect performance after it's fixed. It should only delay the initial indexing.

Also, the **sort order for working set items has changed**. In master, the currently viewed file is displayed at the top and all other are sorted as they are in the file tree. In this branch, all of the working set items are listed at the top. I think I like this better, but thought I'd bring it up for discussion.

I also notice that search bar is not dismissed after instant search. This makes sense because the search is updated as you type. I think it also makes sense to **always show the bottom panel** (even when there are no hits for search). I came to this conclusion in my branch so user gets faster feedback, but it also makes sense with instant search because panel can go away and come back as you type.

I also ran into the problem when **searching too early and nothing happens** -- need some kind of feedback or execute search when ready.

**Ctrl-click on disclosure arrows in panel is broken**. It should toggle all entries open/closed. It either does nothing, or it toggles all and then after a short delay toggles everything back to initial state

This branch is very promising. Overall a great job!

nethip:
@redmunds Thanks for your kind words! And really appreciate you checking out this branch and sharing your feedback. The team indeed had put in lot of effort to get it to this state. So kudos to the entire team for doing a fabulous job and for putting in countless hours. In fact we had go back and forth between web threads vs a new node process vs using existing node process. Finally we settled with using existing node process.(which gave us more challenges). We are really excited about this feature and can't wait to see this getting shipped for 1.4. 

Agree with you on the visual feedback when instant search is not yet available/ the search takes more time to get the results.

petetnt:
This is really a killer feature, awesome :+1: 

RE: Waiting for the indexing to complete: would it be possible to search the partial cache while the indexing is still going on? As in that the crawling would send periodical update events at which point the instant search would update itself from that point of the progress. 

After everything is cached, the *Indexing for instant search* spinner would disappear and Instant search would work like it normally does. This way you wouldn't have to fallback to the old Find In Files even if the caching was on-going. (I can see how this makes cases where uses changes his search terms while the indexing is going on more tricky though)

nethip:
> RE: Waiting for the indexing to complete: would it be possible to search the partial cache while the indexing is still going on? As in that the crawling would send periodical update events at which point the instant search would update itself from that point of the progress.

@petetnt This is exactly is what is happening now. Searching and caching are now interleaved. As I when the first 100 results are got, they are sent back to the Brackets. @abose @rroshan1 can fill in more finer technical details about this.

> After everything is cached, the Indexing for instant search spinner would disappear and Instant search would work like it normally does. This way you wouldn't have to fallback to the old Find In Files even if the caching was on-going. 

Agree with the instant search spinner should disappear once the caching is complete.

> (I can see how this makes cases where uses changes his search terms while the indexing is going on more tricky though) 

These were the exactly the challenges with instant search I was mentioning about earlier. The team did lot of collaboration on this front, and made sure we don't have race conditions. e.g. the search results and the search terms should be in sync. Of course we still need to handle some conditions where search 
is taking some time to get the results in which case we need to have visual cues.

abose:
Thanks all for the review,
I have added the spinner as suggested by @larz0 on caching progress.
@redmunds 
The sort order has been changed so that we see the current file, followed by other files in working set(that passes the filters), and then the other results. We did not sort the results as that was one of the larger performance hogs, but yes, we only sort the file list at the start of a search, so this should not affect the instant search when optimizes properly. I Like the current order though, it follows the project tree order more predictable that the other order- which i only came by when reading the code.

About the search too early, if you press enter after a search, the search would be forced until 100 results are available or all files are read. This would not block brackets though. If you press enter 2-3 times during this time, all the searched would be queued. We could pop a toast message when this happens telling that a search is already in progress and not queue the search.

The disclosure button thing is a work in progress which i just learnt from the review comments. Checking it out.

redmunds:
@abose 

>  I Like the current order though, it follows the project tree order more predictable that the other order- which i only came by when reading the code.

No, it's not in project tree order (@TomMalbran and I wish it was). If you been following the discussion in https://github.com/adobe/brackets/pull/11128 you know I've been looking at this.

For example, using the recipe I describe [here](https://github.com/adobe/brackets/pull/11128#issuecomment-116297972), go to end of results, then page backwards, and you'll see that the results are not in file tree order.

Take a look at commits https://github.com/adobe/brackets/commit/607979620cc8ebc0ebe11856217d51b7fb56c021 / https://github.com/adobe/brackets/commit/0e85ed022610746a40f195d4b774fe30e64e43cc and https://github.com/adobe/brackets/commit/d5cf805a0a346da60cf17ebc0a47e58b6df08e8f for some sorting improvements. I also have a local (not yet pushed to branch) fix for sorting ProjectModel cache in ~1.1s if you want to see that (but would also need to fix clearing cache too often).

mackenza:
@abose the visual indication of the spinner for caching the project is working for me. My Brackets project seems to take about 10 secs or so to complete caching on a HP Folio 9470m which has a Core i5-3437U and 4GB Ram on Windows 7 64-bit.

What triggers a cache refresh? I notice when I move away to another project and then back to Brackets, I don't seem to see the spinner anymore so I am assuming it is keeping the cache.

redmunds:
> What triggers a cache refresh?

Open src/project/ProjectModel.js and put breakpoints on `this._allFilesCachePromise`

This gets cleared on every file write. Unfortunately, this happens every time state.json is updated, which is on every file tree click or editor selection change.

abose:
@mackenza in the node based implementation, the file cache is updated on project switch. The OS also does some file access optimizations, so the second time onwards, the file reads are much more faster[<3 secs on brackets src]- so the spinner won't be shown as the caching is already done.

@redmunds 
Assuming that the file list only needs to be sorted once, we should still be able to sort with instant search with some careful first-time flags around. But I think the sorting is not helping the search workflow in that I always thought the sort was pretty random until I read the code. Also having to click to scores of next buttons to reach a search is also not a good workflow. 
Instead of that, I suggest a different workflow; what if we maintain an unsorted search, But- We wire the `Find in` option to instant search, so that whenever someone selects `find in` from a folder in the project tree- we do an instant search with the new search scope?

mackenza:
@abose I think you want to consider the impact of a random search results order. https://github.com/atom/find-and-replace/issues/64 shows it as a potential showstopper (as in switching back to ST3).

I am not sure I see your connection between Find In and Find in Files... maybe you could expand upon your proposed use case?

TomMalbran:
@abose Even if the way the files are sorted seems random, is good to have them sorted, so that files are always added in the same order, and this is important when the results are updated while the panel is open after doing a search. Without sorting I could easily remove all the results from one file and re add some results making the file jump to the end of the list. And I believe that there are more examples that can do that.

Notice that the files are sorted every time the UI is shown. Maybe we don't need to sort the files before searching, but I think that we do need to sort them when displaying them.

abose:
@redmunds @TomMalbran I have added a sort flag to ProjectManager.getAllFiles so that the files will be sorted during the tree traversal itself with file/dir names at each level. 
Please check this out.
The collapse all big is under fix.

abose:
Implemented the collapse also. Since we are doing instant search, once a collapse is done, it will be retained on further instant searches and vice versa.

src/search/FileFilters.js
```diff
@@ -38,7 +38,8 @@ define(function (require, exports, module) {
         Strings            = require("strings"),
         PreferencesManager = require("preferences/PreferencesManager"),
         EditFilterTemplate = require("text!htmlContent/edit-filter-dialog.html"),
-        FilterNameTemplate = require("text!htmlContent/filter-name.html");
+        FilterNameTemplate = require("text!htmlContent/filter-name.html"),
+        FindUtils          = require("search/FindUtils");
```
prafulVaishnav: NIT:Put templates to last.

src/search/FindBar.js
```diff
@@ -242,6 +249,9 @@ define(function (require, exports, module) {
             self.showError(null);
             self._modalBar = null;
             self._closed = true;
+            clearInterval(intervalId);
```
prafulVaishnav: It will throw jshint error. Do window.clearInterval

src/search/FindBar.js
```diff
@@ -258,8 +268,41 @@ define(function (require, exports, module) {
...
+                                    console.log('search exec');
+                                    self.trigger("doFind");
+                                    lastQueriedText = self.getQueryInfo().query;
+                                }
+                            }
+                        }
+                    }
+                };
+                if (intervalId === 0) {
+                    intervalId = setInterval(executeSearchIfNeeded, 50);
```
prafulVaishnav: Same here

src/search/FindBar.js
```diff
@@ -258,8 +268,41 @@ define(function (require, exports, module) {
                 $(e.currentTarget).toggleClass("active");
                 self._updatePrefsFromSearchBar();
                 self.trigger("queryChange");
+                if (self._options.multifile) {  //instant search
+                    self.trigger("doFind");
+                }
             })
             .on("keydown", "#find-what, #replace-with", function (e) {
+                var d = new Date();
+                lastTypedTime = d.getTime();
```
prafulVaishnav: Can you club it as (new Date()).getTime()

src/search/FindBar.js
```diff
@@ -258,8 +268,41 @@ define(function (require, exports, module) {
...
+                    // We only do instant search via node.
+                    if (FindUtils.isNodeSearchDisabled() || FindUtils.isInstantSearchDisabled()) {
+                        // we still keep the intrval timer up as instant search could get enabled/disabled based on node busy state
+                        return;
+                    }
+                    if (self._closed) {
+                        return;
+                    }
+                    d = new Date();
+                    currentTime = d.getTime();
```
prafulVaishnav: Can you club it as (new Date()).getTime()

src/search/FindInFiles.js
```diff
@@ -42,8 +43,19 @@ define(function (require, exports, module) {
         LanguageManager       = require("language/LanguageManager"),
         SearchModel           = require("search/SearchModel").SearchModel,
         PerfUtils             = require("utils/PerfUtils"),
+		NodeDomain            = require("utils/NodeDomain"),
+		FileUtils			  = require("file/FileUtils"),
```
prafulVaishnav: Fix Indentation..

src/search/FindInFiles.js
```diff
@@ -74,16 +86,23 @@ define(function (require, exports, module) {
...
+        DocumentModule.on("documentChange", _documentChangeHandler);
+        FileSystem.on("change", _fileSystemChangeHandler);
+        DocumentManager.on("fileNameChange",  _fileNameChangeHandler);
     }
     
+    function nodeFileCacheComplete() {
+        FindUtils.setInstantSearchDisabled(false);
+        // Node search could be disabled if some error has happened in node. But upon
+        // project change, if we get this message, then it means that node search is working,
+        // we we re enable node search. If a search fails, node search will be switched off eventually.
```
prafulVaishnav: correct typo..

src/search/FindInFiles.js
```diff
@@ -401,6 +423,37 @@ define(function (require, exports, module) {
...
+
+     /**
+     * @private
+     * sends all changed documents that we have tracked to node
+     */
+    function _updateChangedDocs() {
+        var files = MainViewManager.getWorkingSet(MainViewManager.ALL_PANES),
+            i = 0,
+            file = null,
+            key = null;
```
prafulVaishnav: It seems i, files, file is not used.

src/search/node/FindInFilesDomain.js
```diff
@@ -0,0 +1,661 @@
+/*
+ * Copyright (c) 2012 Adobe Systems Incorporated. All rights reserved.
```
prafulVaishnav: Correct the year to 2015..

src/search/node/FindInFilesDomain.js
```diff
@@ -0,0 +1,661 @@
...
+     * @returns {number} number of matches
+     */
+    function countNumMatches(contents, queryExpr) {
+        if (!contents) {
+            return 0;
+        }
+        var matches = contents.match(queryExpr);
+        if (matches && isNaN(matches.length)) {
+            console.log('contents for nan' + contents);
+            console.log(JSON.stringify(matches));
```
prafulVaishnav: matches would be either `null' (if no matches) or array. So isNaN check is not required as it will always be true if matches is not null. You can remove this block.

src/search/node/FindInFilesDomain.js
```diff
@@ -0,0 +1,661 @@
...
+     * Crawls through the files in the project ans stores them in cache. Since that could take a while
+     * we do it in batches so that node wont be blocked.
+     */
+    function fileCrawler() {
+        if (!files || (files && files.length === 0)) {
+            setTimeout(fileCrawler, 1000);
+            return;
+        }
+        var i = 0;
+        for (i = 0; i < 10 && currentCrawlIndex < files.length; i = i + 1) {
```
prafulVaishnav: can you replace i = i + 1 -> i++

src/search/node/FindInFilesDomain.js
```diff
@@ -0,0 +1,661 @@
...
+    
+    /**
+     * Remove the list of given files from the project cache
+     * @param   {Object}   updateObject
+     */
+    function removeFilesFromCache(updateObject) {
+        var fileList = updateObject.fileList || [],
+            filesInSearchScope = updateObject.filesInSearchScope || [],
+            i = 0;
+        for (i = 0; i < fileList.length; i = 1 + 1) {
```
prafulVaishnav: Already defined i = 0. Something wrong in `i = 1 + 1`

src/search/node/FindInFilesDomain.js
```diff
@@ -0,0 +1,661 @@
...
+     * read at this time. We just delete the project cache entry which will trigger a fetch on search.
+     * @param   {Object}   updateObject
+     */
+    function addFilesToCache(updateObject) {
+        var fileList = updateObject.fileList || [],
+            filesInSearchScope = updateObject.filesInSearchScope || [],
+            i = 0,
+            changedFilesAlreadyInList = [],
+            newFiles = [];
+        for (i = 0; i < fileList.length; i = 1 + 1) {
```
prafulVaishnav: Already defined i = 0. Something wrong in `i = 1 + 1`

src/search/SearchModel.js
```diff
@@ -28,7 +28,8 @@ define(function (require, exports, module) {
     
     var FileUtils   = require("file/FileUtils"),
         EventDispatcher = require("utils/EventDispatcher"),
-        FindUtils   = require("search/FindUtils");
+        FindUtils   = require("search/FindUtils"),
+        MainViewManager       = require("view/MainViewManager");
```
prafulVaishnav: Can you correct indentation..

src/search/SearchModel.js
```diff
@@ -199,23 +200,63 @@ define(function (require, exports, module) {
...
+                    return 1;
+                }
+                return FileUtils.comparePaths(key1, key2);
+            });
+        }
+
+        firstFile = firstFile || "";
+
+        // Create a working set path map which indicates if a file in working set is found in file list
+        for (i = 0; i < workingSetFiles.length; i++) {
```
prafulVaishnav: i already defined.

src/search/FindUtils.js
```diff
@@ -352,11 +387,209 @@ define(function (require, exports, module) {
...
+    exports.notifyIndexingStarted           = notifyIndexingStarted;
+    exports.notifyIndexingFinished          = notifyIndexingFinished;
+
+    // events raised by FindUtils
+    exports.SEARCH_FILE_FILTERS_CHANGED              = "fileFiltersChanged";
+    exports.SEARCH_SCOPE_CHANGED                     = "searchScopeChanged";
+    exports.SEARCH_INDEXING_STARTED                  = "searchIndexingStarted";
+    exports.SEARCH_INDEXING_FINISHED                 = "searchIndexingFinished";
+    exports.SEARCH_COLLAPSE_RESULTS                  = "searchCollapseResults";
 });
```
prafulVaishnav: Are we listening to change in the instantSearch enable/disable preference by user?

abose:
instantfilter

Implemented new instant workflow: so that searching within a particular folder reflects live.
This was the `find in` change i mentioned earlier.

src/search/FindUtils.js
```diff
@@ -352,11 +387,209 @@ define(function (require, exports, module) {
...
+     * @return {Array.<*>}
+     */
+    function prioritizeOpenFile(files, firstFile) {
+        var workingSetFiles = MainViewManager.getWorkingSet(MainViewManager.ALL_PANES),
+            workingSetFileFound = {},
+            fileSetWithoutWorkingSet = [],
+            startingWorkingFileSet = [],
+            propertyName = "",
+            i = 0;
+        firstFile = firstFile || "";
```
prafulVaishnav: correct indentation

src/search/FindUtils.js
```diff
@@ -352,11 +387,209 @@ define(function (require, exports, module) {
...
+
+        // Create a working set path map which indicates if a file in working set is found in file list
+        for (i = 0; i < workingSetFiles.length; i++) {
```
prafulVaishnav: i already defined 0

src/search/FindUtils.js
```diff
@@ -352,11 +387,209 @@ define(function (require, exports, module) {
...
+    function isNodeSearchDisabled() {
+        return _prefNodeSearchDisabled() || nodeSearchDisabled;
+    }
+
+    /**
+     * check if a search is progressing in node
+     * @returns {Boolean} true if search is processing in node
+     */
+    function isNodeSearchInProgress() {
+        if (nodeSearchCount === 0) {
```
prafulVaishnav: Can we replace it with return !(nodeSearchCount === 0)

src/search/FindUtils.js
abose: we always check for the preference value in `isInstantSearchDisabled` , so we do not need to listen for preference change in this case.

src/search/FindUtils.js
abose: The line is not part of the var statement, so the indentation is correct in this case.

src/search/FindUtils.js
abose: Not initing it here will throw jslint error.

src/search/FindUtils.js
abose: Linter says (confusing use of !)[http://derickbailey.com/2015/03/25/jshint-confusing-use-of-bang/].

src/search/SearchModel.js
abose: Linter says have to define.

src/search/node/FindInFilesDomain.js
abose: Good catch!. I am guessing that the fileList.length might have been mostly < 2

prafulVaishnav:
@abose Failed due to linting errors.

abose:
@dvorapa Node it is .

dvorapa:
Thank you

PR #11378, opened 2015-07-11
Added scrollTop in extension manager 

darkcoderrises:
Resets extension manager's scroll position to top on filtering


ficristo:
@darkcoderrises in https://github.com/adobe/brackets/pull/11700 was mentioned it was better to create a variable to hold the value of `$(".modal-body", $dlg)`.
Could you take a look?

zaggino:
Replaced by https://github.com/adobe/brackets/pull/12805

PR #11383, opened 2015-07-13
Adding theme name to the health data log

nethip:


nethip:
@prafulVaishnav @abose Could you guys take a quick look at this PR and see if this looks fine?

prafulVaishnav:
LGTM.. Merging..

nethip:
Thanks @prafulVaishnav :)

peterflynn:
@nethip @prafulVaishnav It seems like we should only send the theme name if it's in the Extension Registry, similar to what we do for extensions...  Is there time to do that for 1.4?

nethip:
@peterflynn Unfortunately we are pretty tight on the schedule. We are targeting for a release today itself! (hopefully there are no blockers)

I am wondering what are the use cases in which we would want to figure out if the theme is present in the registry or not and only then log it? I am guessing percentage of people installing themes from outside extension registry is minimal. The idea was to know what are the most used themes inside Brackets irrespective of where the theme is coming from. 

We could hard code to log only if the theme is `light` or `dark`, but that way we would never know about other themes used inside Brackets. 

Am I missing something here?

peterflynn:
@nethip This is a privacy protection we put in place for logging extension names.  It seems like the exact same rationale applies to themes too.  Theme names seem a _little_ less sensitive than extension names, but I don't think we should surprise people by logging the names of themes they didn't intend to share publicly (especially if we post them on a wiki summary for the world to see).  If there's no time to apply this check for theme names, maybe we should disable logging them for now?

peterflynn:
@nethip Also btw, before we ship 1.4 I think we need to update https://github.com/adobe/brackets/wiki/Health-Data with the new list of what's being logged.  We link to this from the UI, and we promised to update it whenever we add things to the Health Report...

nethip:
@peterflynn That is a very good catch! This was what I was missing :smile: . I just spoke to @prafulVaishnav, he says it is relatively simple. Let us target this for 1.4 itself. If it is turning out to be too much of work, we will then disable logging of this for 1.4. Also thanks for pointing out the Health Data URL that need to be updated. I will add to the deployment tasks.

peterflynn:
Sounds good!

PR #11388, opened 2015-07-13
Update descriptions of preferences

sprintr:
Changed `True` to Boolean value `true` in the descriptions. Also changed description of `jslint.options/stupid`.

CC: @abose


abose:
Thanks for the quick fix. :cool: 
@rawat11 I think you asked for one of the changed in this PR `jslint.options/stupid`

PR #11399, opened 2015-07-14
Czech translation for 473f8f0

dvorapa:


abose:
Changes made by @sprintr into nls/cs. Merging.
Thanks @dvorapa .

PR #11417, opened 2015-07-20
Create templates

dkdev179:


abose:
What is this change about?

abose:
Closing change as no-op

PR #11441, opened 2015-07-22
Health Logs related to new find in files workflow.

abose:
Adds,
1. Prefs users has set for instant and node search disable
2. number of files in the project and the in-memory cache size for instant search.

addresses #11375 , #11055

### DIFF
--- a/src/search/node/FindInFilesDomain.js
+++ b/src/search/node/FindInFilesDomain.js
@@ -48,7 +48,8 @@ maxerr: 50, node: true */
         lastSearchedIndex = 0,
         crawlComplete = false,
         crawlEventSent = false,
-        collapseResults = false;
+        collapseResults = false,
+        cacheSize = 0;
     
     /**
      * Copied from StringUtils.js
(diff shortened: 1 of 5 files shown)


src/search/node/FindInFilesDomain.js
```diff
@@ -324,9 +325,13 @@ maxerr: 50, node: true */
...
             return;
         }
-        var i = 0;
+        var i = 0,
+            contents = "";
         for (i = 0; i < 10 && currentCrawlIndex < files.length; i++) {
-            getFileContentsForFile(files[currentCrawlIndex]);
+            contents = getFileContentsForFile(files[currentCrawlIndex]);
+            if (contents) {
+                cacheSize += contents.length;
```
peterflynn: @prafulVaishnav @nethip Fwiw I think this number is going to be inaccurate -- if I'm reading this code correctly, every time you re-index a file that was updated, you're going to add it to the existing cacheSize total even though it's overwriting its old entry.  So if a file is updated often, it could get counted in the cacheSize 100 times instead of just once.

abose:
Actualy the file crawl routine `fileCrawler` is only called once on project switch. File changes and re-index happen via a different route. So, we do not account for changes in project size as the user edits in brackets- the size is computed only on the project switch. So, the data sent should be accurate.

PR #11445, opened 2015-07-22
Typo in path to correspond with src/nls/root/strings.js

dvorapa:


abose:
Thanks @dvorapa for catching this one. 
Minor issues like this would have gone unnoticed.

PR #11452, opened 2015-07-23
Resolving the Unit Test Failures with modified Find In Files

rroshan1:
Resolving the Unit Test Failures with modified Find In Files


abose:
Looks good. Merging.

PR #11456, opened 2015-07-23
After fresh install health data should always return an empty object if not already defined.

abose:
After fresh install health data should always return an empty object if not already defined.


rroshan1:
Looks good. Merging.

PR #11466, opened 2015-07-24
Removing Hardcoded String

prafulVaishnav:


nethip:
Unfortunately, we have to wait for another day for the travis to kick in(it says we have reached the api rate limit for today.). Since this is a small change which we are targetting for 1.4, I am going ahead with the merge.

PR #11476, opened 2015-07-25
Health data for Find in files/instant search

abose:


abose:
Closing as branched out from release. will retarget once release is merged back to master.

PR #11481, opened 2015-07-25
FindBar not working for case insensitive values

gitexperience:
It behaves differently from the very common searching methods. For ex- If you want to search "getResults" and you type "getresults" , it won't find anything . So , this patch actually fixes this.


abose:
@gitexperience it is working as expected when i tried it out.
image
When the circled `Aa` button is not pressed, a case insensitive search is performed.; and when it is pressed, a case sensitive search is performed. I see that it is working as expected.

gitexperience:
Ohh yes! I have disturbed my own local fork and that's why it is creating problem .With a new one it works fine for me too , so no use of that PR ..

PR #11484, opened 2015-07-26
Convert numbers to strings, so StringMatch can match it.

sprintr:
Converts `"number"` values into strings so `StringMatch.stringMatch` can match it. 

@abose This is critical to go with the release.

### DIFF
--- a/src/extensions/default/PrefsCodeHints/main.js
+++ b/src/extensions/default/PrefsCodeHints/main.js
@@ -281,6 +281,13 @@ define(function (require, exports, module) {
                     return null;
                 }
                 
+                // Convert integers to strings, so StringMatch.stringMatch can match it.
+                if (option.type === "number" || option.valueType === "number") {
+                    values = values.map(function (val) {
+                        return val.toString();
+                    });
+                }
+
                 // filter through the values.
                 hints = $.map(values, function (value) {
                     var match = StringMatch.stringMatch(value, query, stringMatcherOptions);
@@ -332,8 +339,7 @@ define(function (require, exports, module) {
             }
             
             // Put quotes around completion.
-            completion = quoteChar + completion;
-            completion = completion + quoteChar;
+            completion = quoteChar + completion + quoteChar;
             
             // Append colon and braces, brackets and quotes.
             if (!ctxInfo.shouldReplace) {


abose:
@sprintr what are the repercussions if we do not have this? what will break/not break. Could you give more details about the problems this pr addresses.
@nethip

sprintr:
`"number"` type preferences when `stringMatch`(ed) will through an `undefined is not a function` and also number type preferences were not matched because of this since `stringMatch` expects string values to match against a query.

Came across this while writing docs for #11200. Have sent them to you via slack.

abose:
OK, that is serious in that it would break brackets when trying to edit preferences file?

sprintr:
Run this in brackets console to see what error get produced.

```js
var StringMatch = require("utils/StringMatch");
StringMatch.stringMatch(1024, "10", {  preferPrefixMatches: true });
```

sprintr:
Yes it broke hints when you press enter. So this fixes that as well.

abose:
Thanks @sprintr . 
Re- GM an RC build is a tough thing to push to the stakeholders as it is an expensive/labour intensive process. Several teams(brackets-dev, Release engineering, LOC testing, international translation vendors) & Lot of manual testing is involved prior to an RC build of brackets.- so just wanted to give reasons to justify that.
@nethip has to make the call. 
If we are trigerring a rew release, I would also like to nominate #11485 - Health logs for find / replace in files, instant search usage.

nethip:
@sprintr How often do the users run into this use case? I did not understand the workflow with the `number` preference. Could you tell me what are the exact steps to repro this (from the user point of view). As @abose mentioned, the release tasks are quite laborious. If the frequency of someone running this issue is very minimal, I would nominate this for the next release.

sprintr:
I don't know any ext using number preference but since brackets supported that even before this feature I believe we should support it here as well. Right now it is broken.

nethip:
@sprintr OK! Could you mention the exact repro steps? Let me speak to the team and take an opinion.

sprintr:
You can copy any number type preference from the wiki page "'Preference code hints for developers". I am on phone so can't give exact steps to repro this

src/extensions/default/PrefsCodeHints/main.js
```diff
@@ -281,6 +281,13 @@ define(function (require, exports, module) {
                     return null;
                 }
                 
+                // Convert integers to strings, so StringMatch.stringMatch can match it.
+                if (option.type === "number" || option.valueType === "number") {
+                    values = values.map(function (val) {
```
nethip: @sprintr use `$.map(values, function (val) {` here.

sprintr: I guess raw ES5 should be faster

nethip: Ignore this. I think the one you have above should work fine.

sprintr:
@nethip Thanks for getting this in. When is the release scheduled?

nethip:
@sprintr No issues! Thanks for jumping in and fixing it. Really appreciate it. We are targeting today for doing the release.

sprintr:
It was nice to get this in time.

PR #11485, opened 2015-07-26
Health logs for find / replace in files, instant search usage.

abose:
Sould be helpful in seeing how people are using the new search features.


abose:
Targeting this to master.

Issue #11477, opened 2015-07-25
Brackets freezes when installing/uninstalling extensions (and other situations) and doesn't respond on Windows 7 64-bit


AlexanderJ2357:
Brackets freezes a lot on windows 7 (I have it on OSX too and haven't noticed freezes there but I don't use OSX as much as windows). It freezes especially when I am installing/uninstalling extensions.

Another situation in which it freezes is when I open the Extension Manager and try to close it by clicking on the Close button while it is loading.

I am afraid that I lose unsaved work when it freezes. I hope it is possible to fix this issue.

hussainb:
+1.
This issue was intermittent before the  build 1.4.0-16380, but after upgrading to build 1.4.0-16380 it happens always, the same happens when pressing Reloading(F5) while developing extensions.
Have to use task manager to kill brackets every time and this is very irritating.

nethip:
@AlexanderJ2357 @hussainb Do you have any extensions installed? Just to rule out if any extension is creating the issue, could you rename the preferences folder, and relaunch brackets?

AlexanderJ2357:
I do have a few of the popular extensions installed like beautify, emmet etc.
I read that a lot more people are experiencing crashing issues with Brackets, especially with large files.
I hope the issues are fixed in the future but for now, I am use Atom and happy with it.

hussainb:
@nethip 
I renamed the user folder under the extensions folder and opened brackets, It created a new empty user folder and then I tried to reload(F5). This time too brackets crashed. I tried to trace errors in the console but when brackets crashes, the developer tools crashes too.
This now doesn't look like an extension issue to me.

The only thing I noticed that if no file is open in the editor (can see the 'CODE THE WEB' logo and project folder open in sidebar) the reload doesn't hang brackets.

System Info
Windows 7 64-bit

nethip:
@hussainb Could you disable our instant search and see if this problem goes away. Use the preferences below to do that. 

```
"findInFiles.instantSearch": false,
"findInFiles.nodeSearch": false
```

Also before reloading, give it a few seconds before Brackets gets completely initialised.

hussainb:
@nethip 
I decided to debug brackets by myself, What I found is that the function `browserReload(href)` in `DocumentCommandHandlers.js` needed a timeout before it updated the href at line no 1638.

The url :
[...]

PR #11505, opened 2015-07-29
Resolve Brackets freezing/crashing on windows on reload.

hussainb:
Workaround to resolve brackets freezing and crashing on some windows OS on installing/uninstalling extensions and on reload. 
Issue Reference: https://github.com/adobe/brackets/issues/11477

### DIFF
--- a/src/document/DocumentCommandHandlers.js
+++ b/src/document/DocumentCommandHandlers.js
@@ -1637,7 +1637,7 @@ define(function (require, exports, module) {
                 // Defer for a more successful reload.
                 setTimeout(function(){
                     window.location.href = href;    
-                },1000)
+                },1000);
                 
             });
         }).fail(function () {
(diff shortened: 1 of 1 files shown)


zaggino:
Confirmed as a solution and merged.

nethip:
Thanks @hussainb for this PR! And thanks @zaggino for reviewing this and merging.

zaggino:
Yes, it's a bit weird fix to be honest, but it worked on two of my pc's so I've fixed the jslint issues and merged it.

petetnt:
Quoting myself from #11702 

```
I think that this particular bugfix should have been released as a version PATCH (as in 1.4.1.) instead of waiting for version 1.5. At worst this bug might have affected **every** Brackets user out there due to it's frequency plus it was a hard crash which makes it even worse. Luckily it doesn't cause data loss.

Brackets hasn't had a patch version yet, but it certainly is something that should be looked into in cases like this.
```

PR #11538, opened 2015-08-03
Allow code-folding for selected text in editor

thehogfather:
This is a small addition to the code folding extension that makes any selected text in the editor foldable (provided the number of lines selected is more than the minimum fold size set in the preference).

The feature can be enabled by setting the preference value `"code-folding.makeSelectionFoldable": true`

cc @abose


nethip:
@thehogfather Wow! That was quick. Thanks for this PR!.

src/extensions/default/CodeFolding/Prefs.js
```diff
@@ -35,7 +36,9 @@ define(function (require, exports, module) {
                            {name: HIDE_FOLD_BUTTONS, description: Strings.DESCRIPTION_CODE_FOLDING_HIDE_UNTIL_MOUSEOVER});
     prefs.definePreference("maxFoldLevel", "number", 2,
                            {name: MAX_FOLD_LEVEL, description: Strings.DESCRIPTION_CODE_FOLDING_MAX_FOLD_LEVEL});
-    
+    prefs.definePreference("makeSelectionsFoldable", "boolean", true,
+                           {name: MAKE_SELECTIONS_FOLDABLE, description: MAKE_SELECTIONS_FOLDABLE});
```
MarcelGerber: Use `Strings.*` instead for the `description` so it's translateable.

abose:
Thanks @thehogfather for the PR.

src/extensions/default/CodeFolding/Prefs.js
thehogfather: @MarcelGerber sure thing! I shall update.

thehogfather:
BTW this is still a work in progress. Amongst other things, I need to sort out issues around restoring folds based on selections. Shall update in the next few days. Case `2` below is  of particular interest:

1. User selects text, folds it, closes the file. When user reopens the file (without modification) the folded selected text region is restored.
2. User selects text, folds it, closes the file. File is modified outside the context of brackets (e.g., git). User opens the file. **What is a reasonable expectation here?**

abose:
@thehogfather I have added a review only tag to the title of the PR, please remove it when you feel this pr is ready to be merged!

petetnt:
@thehogfather I wrestled with a case similar to 2. with my [brackets-persistent-history extension](https://github.com/petetnt/brackets-persistent-history), where I wanted to persist the editing history between project changes *and* outside changes. My original idea was to invalidate the editing history on the changed part by diffing the files (similar to how you could invalidate / uncollapse changed lines) but in the end I went for a simpler solution where I add another history-step to compensate between the state in the cache and the new changes from outside of Brackets.

However the case here is slightly different so it's hard to say what the reasonable expectation would be: overall I wouldn't want Brackets to uncollapse any folds unless absolutely necessary (ie. the folded content has been modified in a way that it doesn't exist (start/end has been removed/modified) anymore), but then again how would you notice that something has changed inside a collapsed fold? One idea would be showing the change in the collapse marker but that's kinda out there too.

thehogfather:
@petetnt quite! 
The case here is also interesting because there really isn't a way to "validate" selection-based folds - any selection is valid - unlike how modifications to a file can render brace-based folds invalid. At the moment what the extension does is to use the starting line of each collapsed region to calculate the 'intended' collapsed region when the file is reopened. If the file has not changed then there is no problem. If the file has changed, then the fold might be invalidated if the start and end lines of the new and cached ranges don't match.

For now I've kept it simple. Selection folds will be restored if and only if they exist in the saved fold preferences and are reported as selection by the ViewStateManager.

It might be worth thinking about rendering selection folds ever-so-slightly differently from other types of folds in the gutter.

thehogfather:
@abose I think this is ready to merge.

nethip:
@abose I think we should target this for 1.5

abose:
It working well in all cases i tested. External modification is always a grey area, But no data is getting corrupted / opening up the folds on external modification seeems to be reasonable- which is how it is working now mostly. 
All unit/integration/main view tests Passing.

Thanks for this feature @thehogfather :+1: 
Merging.

Maetzzen:
Im really happy about this new feature in version 1.5.
But could one tell me if there is a possibility to safe this options?

If i restart my brackets, my html files are folded out and i have to manually fold in the stuff i dont want to see again.
There is some content i dont have to see anymore and it would be more clearly if they are folded in.

Greetings Maetzzen

ezequias:
Could someone tell me if the collapsed text is correctly iconified on my image (that ugly crossed square)?

image

thehogfather:
@ezequias that is correct and would be and issue to raise with the theming provider you are using. @Maetzzen the fold states should persist when you close and reopen brackets. Can you report what operating system you are running?

Maetzzen:
My bracket runs on Windows 8.0

ezequias:
Thanks @thehogfather It's Monokai dark soda.

thehogfather:
@Maetzzen can you confirm if there is any difference if you select a different file in your workspace before closing brackets. So fold some regions in your html file as you did before then switch to a different file and then close and reopen brackets. 

Now check if the folds in your html file are restored or not.

Maetzzen:
@thehogfather
It doesnt matter which kind of file i select.
The files and their segments are minimized as long i keep the brackets opened.
But whenever i close and reopen brackets the segments arent minimized anymore.
In my html element i have a lot of content safed in different tables. And i minimize them to keep an overview but if i reopen brackets the adjustment is resetted. I hope you understand what i mean ^_^'

thehogfather:
Yeah I understand. But I have a theory about why you are experiencing the issue. To disprove or support my theory your project needs to have multiple files. And you need to change the active editor after you've folded a code region. Hence why I asked that you select a different file (before closing brackets and after folding some regions in an intended file).

I could reproduce the issue only if I opened brackets, worked on only a single file and then closed brackets. Hope you understand what I'm trying to say.

petetnt:
@ezequias' issue is reported here: rainje/Monokai-Dark-Soda/issues/6, there's a suggested workaround mentioned too.

thehogfather:
btw @Maetzzen this issue is related to #11775 so perhaps lets carry on discussions there.

Cheers

ezequias:
Thanks @petetnt but I didn't find the monokai main.css but only monokai.css and it does not have the mentioned text.

rainje:
@ezequias Did you update the theme to 1.0.9 ? It should be fixed. Tell me if it isn't.

ezequias:
@rainje yes. Now looks better but I would use a minus sign with the same color of the plus sign to make it perfect. Anyway now is way better than before. Thanks for caring about me.

Pandahisham:
@thehogfather for later verisons please consider a "fold all other" / "focus mode" option , it's great when writing blogs or similar stuff

PR #11558, opened 2015-08-07
Set Bash files syntax highlighting

sprintr:
Syntax Highlighting for [Bash startup files](http://www.gnu.org/software/bash/manual/html_node/Bash-Startup-Files.html)


nethip:
Thanks @sprintr LGTM

PR #11560, opened 2015-08-07
Map filename to language in case of dotfiles

sprintr:
`LanguageManager.getCompoundFileExtension` returns an empty string for dotfiles and extension-less files because of which the Language menu shows `Set as Default for . Files`. Clicking which adds an empty string as an extension in `"language.fileExtensions"`.

This PR will add dotfiles and extension-less files to `"language.fileNames"` instead of `"language.fileExtensions"`. The dot in extension is added dynamically now, will update it in rest of the locales once this gets reviewed.

### DIFF
--- a/src/editor/EditorStatusBar.js
+++ b/src/editor/EditorStatusBar.js
@@ -325,10 +325,16 @@ define(function (require, exports, module) {
         
         languageSelect      = new DropdownButton("", [], function (item, index) {
             var document = EditorManager.getActiveEditor().document,
-                defaultLang = LanguageManager.getLanguageForPath(document.file.fullPath, true);
+                defaultLang = LanguageManager.getLanguageForPath(document.file.fullPath, true),
+                fileExtension = LanguageManager.getCompoundFileExtension(document.file.fullPath),
+                fileName = null;
+
+            if (fileExtension === "") {
+                fileName = document.file._name;
+            }
             
             if (item === LANGUAGE_SET_AS_DEFAULT) {
-                var label = _.escape(StringUtils.format(Strings.STATUSBAR_SET_DEFAULT_LANG, LanguageManager.getCompoundFileExtension(document.file.fullPath)));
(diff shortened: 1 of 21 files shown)


abose:
All unit integration and extension tests are passing.
Can you make the changes to all other locales?
What would be the behavior if the file does not have any exetsnion?

sprintr:
For a file named `foobar` it should show `Set as default for foobar files`. Will add the following lines to preferences

```json
"language.fileNames": {
     "foo": "c"
}
```

abose:
I meant what would happen for files without extensions? like just the filename `foobar`

sprintr:
@abose I am not sure if I understand your concern here. Files named `foobar` will also get the syntax highlighting of the language they are mapped to. You might wanna checkout out this branch to check it.

abose:
got it!  misunderstood the PR before.
Could you modify the locale strings?

sprintr:
@abose No worries. The strings have been modified.

abose:
could you add a unit test for this?

meeksdoteric:
@sprintr With this I'm able to specify a default language for a specific file, but not able to specify a default language for dotfiles or extensionless files. Also I'm new to Brackets and GitHub, so I apologize in advance for lacking knowledge.

I want 'bash' to be the default language for .bash_profile and similar files.

sprintr:
@meeksdoteric for your particular case #11558 should be fine, this one applies to all the extension-less files and dotfiles.

meeksdoteric:
Thanks!
On Thu, Jan 14, 2016 at 3:23 PM Amin Ullah Khan <notifications@github.com>
wrote:

> @meeksdoteric <https://github.com/meeksdoteric> for your particular case
> #11558 <https://github.com/adobe/brackets/pull/11558> should be fine,
> this one applies to all the extension-less files and dotfiles.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/adobe/brackets/pull/11560#issuecomment-171783502>.
>
-- 
E

zaggino:
This requires some unit tests, I still don't understand what it should do :)

PR #11586, opened 2015-08-12
[review only] Add a "Container Panel" instead of having multiple panels on top of each other.

ficristo:
The patch is still a WIP but I wanted some feedbacks: naming, code conventions, anything.
In this version I changed the createBottomPanel to work already with the extensions and to see how far I could get.
In the future I think I need two API, the old one untouched and a new fresh one.
In the toolbar the is a new button to toggle the container. I need an icon...
As for the UI is not my strong point...


ficristo:
I added a new API to create panels and converted the Code Hint and the Find in Files to use it.
There are still some bugs to resolve before it is ready to be reviewed but feedbacks are welcome.

abose:
I have added a reviw only tag, please remove this once this is ready to be merged in.
Could you comment some screenshots or gif's of the workflow so that it would be eazy to track the feature by other reviewrs?

abose:
@ficristo It looks like this is your first contribution to brackets. Thanks for helping us make brackets.
You would have to sign the brackets contributor license agreement at http://dev.brackets.io/brackets-contributor-license-agreement.html .
Also join the brackets dev forum at https://groups.google.com/forum/#!forum/brackets-dev and our slack channel at https://brackets.slack.com/

ficristo:
I think it is ready to be reviewed.
There are still some issues with the UI / UX but I need some help with them.
I need an icon for the toggle of the container and one for clear the search results, and I'm not sure if the UX is good enough.
For the UI of the panels I've reused the bootstrap tab plugin there was already there.

Here some screenshots.
Search before:
search before
Search after:
search after
Code Hint before:
codehint before
Code Hint after:
codehint after

abose:
:cool: change. 
@larz0 @ryanstewart @nethip @MarcelGerber @zaggino

zaggino:
:+1: nice, I always hated how panel stack on top of each other when I have opened search and code hints at the same time

Denisov21:
+1 nice!

nethip:
Awesome! :+1:

larz0:
@ficristo this is great, could you use the same tab style as the Extension Manager tabs? Let me know if you have any questions.

ficristo:
It is not exactly like the extension manager: I reduced the height to 36px, it seemed good enough to me.
container-style

petetnt:
:cool: This is awesome.

ficristo:
@petetnt If the space is small the tabs go on a second line and so on. AFAIK this is the behavior of the bootstrap tab component and can not easily changed (at least I don't know how, but it would be nice if possible).
I needed a title to show on the tab, so the extensions have to use the new API, unfortunately.

petetnt:
@ficristo Thanks for your reply. I think the tabs collapsing to another line is fine.

One idea of enabling old extensions for the new container could be checking for the `.toolbar > .title` value first, then falling back to the `name` field in the extensions `package.json` file and throwing a warning to the console about missing the title. So in worst case a panel-using extension would show up as `petetnt-panel-using-extension` on the tab. Not sure if that's worth the trouble though.

ficristo:
I had some time so I gave a try.
container-old-api-with-titles
Thanks to the new style the result is better of what I remembered.
But it arise a couple of problems: not all extensions use the title class (the Notes extension), not all extensions have a title (the Git one), in this cases I used their id (I didn't try to read the package, and I already had the id).
Plus the icon in the toolbar were used to show/hide the panel (and the icon changed the color based if the extension was shown or not); with the new API the extensions are always shown, but only one is on the foreground. This causes some visual errors: if you click on the toolbar icon (of the extension) the tab is hidden with the panel, but if you click again you see the panel in the container but not the tab. Some times you don't see the extension anymore.
While i don't like to force the extensions to change, I think the last problem is better solved by the extensions' authors, so I think having a new API is the better route going forward.

petetnt:
@ficristo that doesn't look bad at all, but I also understand how not all of those problems can be solved (like syncing the active extensions) without a new API

le717:
:cool: The multiple, stacked panels have often gotten in my way and putting them in tabs is a very nice way to handle them. Good job. :D :+1:

abose:
As this is a major change, There won't be enough testing window to accommodate this for 1.5. pushing this to 1.6 release.

ficristo:
@larz0 Could you provide me an icon to put in the toolbar to open/hide the container?
in the screenshot above `Search after:` there are two `X` one to hide the container and one to clear the results: do you think is better an icon or to simply write `Clear results`?
Thank you.

src/styles/brackets.less
```diff
@@ -2264,3 +2264,59 @@ label input {
...
+
+            // TODO: find a better way to handle the clean of the search results.
+            &.search-results {
+                .close {
+                    display: block;
+                }
+            }
+
+            .table-container {
+                height: calc(~'100% - 25px - 5px - 5px'); // full height minus title and padding height of the titlebar
```
ficristo: I was able to try on OSX and I think this line doesn't work rightly (and it is a bit hackish)...
@larz0 Is there a better way?

ficristo:
I don't work on this anymore.

Issue #10864, opened 2015-04-09
Code folding doesn't play well with View > Line Numbers toggle


peterflynn:
1. Uncheck _View > Line Numbers_
   - Result: code folding triangles are too close to the left edge of the editor -- need more padding
2. Re-check _View > Line Numbers_
   - Result: code folding is on the _left_ of the line numbers now -- should be to the right

This seems relatively edge-casey, to the point where I could envision shipping 1.3 without fixing this.  I suspect most people who have line numbers hidden will also want folding triangles hidden, and they're unlikely to re-enable line numbers afterward (and issue 2 is easy to work around).

VitorLuizC:
<a href="https://gyazo.com/db7cb0136c4c3ba5ca3649b332c179ac"><img src="https://i.gyazo.com/db7cb0136c4c3ba5ca3649b332c179ac.gif" /></a>

petetnt:
Fixed in #12673

zaggino:
Re-opening as #12673 had to be reverted.

ficristo:
I cannot reproduce anymore. Fixed by #10864

PR #11593, opened 2015-08-15
Ensures fold gutter is always added after line numbers 

thehogfather:
This addresses #11577 - where code folding gutter is not displayed nicely when line numbers are disabled. Also addresses #10864 where toggling the display of line numbers while code-folding is enabled causes fold gutters to be rendered before line numbers.


petetnt:
I think that this PR is pretty crucial and absolutely worth revisiting asap

/cc @abose @swmitra

ficristo:
@thehogfather is this not necessary anymore?

thehogfather:
@ficristo actually it is! - apologies. I was cleaning up my git branches and inadvertently deleted the branch. This is my bad. Will create a new PR and it can be reviewed.

Issue #10717, opened 2015-03-07
Find in Files results don't update to reflect external changes unless file is open


peterflynn:
1. Open the Getting Started project
2. Be careful not to open the main.css file
3. Find in Files and search for "h2"
4. Use a different text editor to open main.css, delete the two rules referencing "h2," and save the file

Result: Find in Files panel still lists both results from main.css

Same thing happens if you edit it to add new results.

But if you create, delete, or rename a file, Find in Files _does_ update correctly -- even if the file isn't open.

Happens at least as far back as 0.43, so this is not a new bug.

peterflynn:
The FileSystem change event has a lot of complex edge cases, and `FindInFiles._fileSystemChangeHandler()` isn't handling one of them -- when `entry` is a Directory but `added`/`removed` are non-null empty arrays, it means _any_ immediate child may have changed.  The current code ignores such change events completely.

I think that type of change event only/mainly happens on Mac.

peterflynn:
Would be great to add a unit test for this too, but: [sigh...](https://github.com/adobe/brackets/blob/release-1.2/test/spec/FindInFiles-test.js#L979-L985)

jburtondev:
Hi, i'm new to the Brackets project and would like to work on this issue. Do I have the go ahead to get started?

abose:
Hi @jburtondev Thanks for helping us out. 
You can get started with issues labeled `starter bugs`. 
Once you have a fix on this/ any other issues, please sign the brackets [CLA](http://dev.brackets.io/brackets-contributor-license-agreement.html) and raise a pull request :+1: 

Also join the [brackets-dev](https://groups.google.com/forum/#!forum/brackets-dev) groups to be update with the community.

jburtondev:
@abose Sure great. Thanks for your message.

tallandroid:
Is there an update on the issue ? If not , I would like to pick it up.

nethip:
@tallandroid Sure! We really appreciate it. Let us know if you require any help on this.

tallandroid:
So , I tried to replicate the bug and here is what i did : 
1) Opened Getting Started project in Brackets. 
2) Didn't open main.css, so it didn't come under my Working Files tab. From my understanding of the code, its not one of the documents under document manager but is still a part of file system.
3) Did a search for h2 and results showed up for main.css.
4) Changed the file in a different editor as directed and saved it back.
 
[...]

PR #11602, opened 2015-08-17
Fix: Find in Files results don't update to reflect external changes unless file is open

amrelnaggar:
This addresses #10717 and covers all FileSystem change events

### DIFF
--- a/src/search/FindInFiles.js
+++ b/src/search/FindInFiles.js
@@ -795,7 +795,7 @@ define(function (require, exports, module) {
         
         var addPromise;
         if (entry.isDirectory) {
-            if (!added || !removed || (!added.length && !removed.length)) {
+            if (!added || !removed || (added.length === 0 && removed.length === 0)) {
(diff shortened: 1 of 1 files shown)


tallandroid:
Let me know if its accepted. I will stop working on it if thats the case.

zaggino:
LGTM

src/search/FindInFiles.js
```diff
@@ -795,7 +795,7 @@ define(function (require, exports, module) {
         
         var addPromise;
         if (entry.isDirectory) {
-            if (!added || !removed) {
+            if (!added || !removed || (!added.length && !removed.length)) {
```
zaggino: I'd personally prefer ` || (added.length === 0 && removed.length === 0))`

amrelnaggar:
Made the changes, this was better indeed.

zaggino:
Thanks @amrelnaggar !

Issue #11126, opened 2015-05-16
Add "Replace all" button when doing Find -> Replace


larsenwork:
Clicking "Batch..." in the top of the screen and then "Replace" in the bottom takes too much time :wink:

![screen shot 2015-05-16 at 14 08 12](https://cloud.githubusercontent.com/assets/3859425/7666070/54bda73e-fbd5-11e4-9fde-556c787a1b08.png)

nethip:
tagging @abose

ficristo:
Fixed by #12988

PR #11606, opened 2015-08-19
Add "replace all" button when doing find and replace

amrelnaggar:
For #11126.


ficristo:
I didn't look closely at the code but I wonder if changing the button id could break extensions.
That said this needs at least a test.

amrelnaggar:
Sorry I've been pretty busy.

I didn't think about extensions. Would it be better if I kept the button id as it was and made a new id for Replace all button? Or unit tests will be enough?

ficristo:
@amrelnaggar I looked around on the public available extensions and I think this is safe as is. So only the test is needed.

It is only to understand if this is something that would be contributed back from Adobe: @swmitra @madanbn?

amrelnaggar:
I added the test.

ficristo:
@amrelnaggar could you merge master so to resolve the conflicts?

amrelnaggar:
@ficristo All conflicts have been resolved. 

If there is anything else needs to be done just let me know.

ficristo:
Something went wrong. It seems you merged an old version of master.
You can use `git rebase -i ...` to drop the last two commits and then try to run `git fetch upstream` and merge again `upstream/master`
Alternatively you can open a new fresh PR.

amrelnaggar:
@ficristo I created a new PR #12988. 

Closing this PR.

PR #11610, opened 2015-08-20
Created a grunt task to generate the samples

ficristo:
I added a task to generate automatically the css for the samples.
I think is possible to autogenerate the samples (almost) completely, but simple step first.
Now there is main.css at samples dir level that is copied in every language samples.
Since there are some differences, see fa-ir and zh-tw, I added manually a styles.css to add / overwrite some attribute.


ficristo:
Now it generates the samples from the configurazione in src/nls folders.

valtlait:
+1. Sample projects used to be very inconsistent, and this helps to that.

ficristo:
I don't work on this anymore.
I this this was better than the current situation, but IMHO getting started projects should be removed.

PR #11614, opened 2015-08-21
Added new line and tab as replacement in case of regex search

### LINKED ISSUE #11611: New line replacement
I saw a couple of issues related to this and I think there is a trello card too.
But I really like to see it fixed, so could someone point to me what should be done?
I asked upstream because there wasn't this feature:
https://github.com/codemirror/CodeMirror/issues/3456
(I disagree a little on how it was fixed)
But I realize now that Brackets do not use the search addon (if I am not mistaken).

ficristo:
This should fix #11611.
The idea comes from https://github.com/codemirror/CodeMirror/issues/3456 since Brackets do not use the search addon.

### DIFF
--- a/src/search/FindUtils.js
+++ b/src/search/FindUtils.js
@@ -103,6 +103,43 @@ define(function (require, exports, module) {
     }
 
     /**
+     * Parses a string and replace the \r, \n and \t sequences of characters
+     * with a character based on the regex meaning of these sequences.
+     * \n => Line Feed
+     * \r => Carriage Return
+     * \t => Tab
+     *
+     * @param {string} string - The string to parse.
+     * @return {string} The replaced text.
+     */
+    function parseString(string) {
+        return string.replace(/\\(.)/g, function (match, ch) {
+            if (ch === "n") {
+                return "\n";
+            }
+            if (ch === "r") {
+                return "\r";
+            }
+            if (ch === "t") {
+                return "\t";
+            }
+            return ch;
+        });
+    }
+
+    /**
+     * Parses a string through parseDollars and parseString functions.
+     *
+     * @param {string} string - The string to parse.
+     * @param {Object} match - The match data from the regexp.
+     * @return {string} The replaced text.
+     */
+    function parseRegexp(string, match) {
+        var str = parseDollars(string, match);
+        return parseString(str);
+    }
+
+    /**
      * Does a set of replacements in a single document in memory.
      * @param {!Document} doc The document to do the replacements in.
      * @param {Object} matchInfo The match info for this file, as returned by `_addSearchMatches()`. Might be mutated.
@@ -126,7 +163,7 @@ define(function (require, exports, module) {
         doc.batchOperation(function () {
             matchInfo.matches.reverse().forEach(function (match) {
                 if (match.isChecked) {
-                    doc.replaceRange(isRegexp ? parseDollars(replaceText, match.result) : replaceText, match.start, match.end);
+                    doc.replaceRange(isRegexp ? parseRegexp(replaceText, match.result) : replaceText, match.start, match.end);
                 }
             });
         });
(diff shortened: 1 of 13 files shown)


ficristo:
I found on the Notepadd++ repo this file https://github.com/notepad-plus-plus/notepad-plus-plus/blob/35adb1910ba98897ad3a935fa7d712a986289bd3/PowerEditor/src/ScitillaComponent/FindReplaceDlg.cpp#L53
I don't know anything about the source code of Notepad++ but it seems to have a more complete set of rules for the replacement.
I think for now if would already be good enough to have this PR and, if needed, in future we could add the other rules.
@abose What do you think? Can this be merged?

ShinDarth:
any news about this?

rlindner81:
:+1: please merge. I ran across this problem multiple times already.

drewscott:
I'm also waiting on this to be merged.

abose:
Thanks @ficristo for this change.
So with the change, If someone wants to replace `xyz` with `\n`[interpreted as a new line char] it can be done only in regexp mode, and if the replace is performed in a non-regexp mode, `\n` will be replaced literally[which is the current behavior]. sounds good to me.

can you add unit test cases for this?

src/search/FindUtils.js
```diff
@@ -522,6 +522,28 @@ define(function (require, exports, module) {
...
+            }
+            if (ch === "r") {
+                return "\r";
+            }
+            if (ch === "t") {
+                return "\t";
+            }
+            return ch;
+        });
+    }
```
redmunds: @ficristo This is an awesome improvement!

I think the name `parseString()` is too generic, though. Maybe `regexUnescapeString()` ?

Does this need to detect when the backslash itself escaped so that you can (for example) use "\\n" to insert "\n" (i.e. not a newline)? Either way, that would be a good unit test.

ficristo: While I agree it is a bit generic, I prefer to preserve the name because it matches the one in [CodeMirror](https://github.com/codemirror/CodeMirror/blob/dbb2a5c270f2b2e626ae03157ed3863626268a6d/addon/search/search.js#L79). But it is not a big deal to change it if you really want.
To insert the chars `\` and `n` in regex mode you need to escape it: `\\n`.
I'll look at adding some tests.

abose:
10 find replace `integration tests` and 2 findinfiles integration tests are failing. 
But i guess its failing even without this change for me.
Have to see what change caused the tests to break.

ficristo:
@abose The failing tests are failing in master too.

ficristo:
Now with tests. I still have some trouble with `\r` on FindReplace.

abose:
@ficristo Are all the cases considered? We could omit the fixes for unit tests that were already failing before this PR. Are there any changes pending for merging this PR?

ficristo:
@abose I still don't know how to test `\r` in FindReplace tests.
I tried to add the following code before the test, in the hope to change the new lines, but didn't worked.
```javascript
runs(function () {
    myDocument.replaceRange(defaultContent.replace(/\n/g, "\r\n"), {line: 0, ch: 0});
    myEditor.centerOnCursor = jasmine.createSpy("centering");
});
```

Plus I just saw the `Search -> Replace All in untitled document`. I need to add some test there too?

ficristo:
I think I've solved the `\r` issue.

ficristo:
@abose @redmunds This should be ready to go. Let me know if I need to add more tests.

ficristo:
Mode to 1.9 because, after all, I think having #10346 first would be better.

ficristo:
Since this introduce a way to change the line endings I thought it was better to have something like #13152 first.
But given this feature is really demanded @adobe/brackets-committers what do you think of merging this one anyway?

saurabh95:
I am also in favour in merging #13152, I have tested it myself and code also looks good to me, and I guess the problems mentioned in it(like loosing undo for line endings etc) can be ignored for now.

PR #11618, opened 2015-08-22
[Review Only][WIP] Proof of Concept to show the default extensions

ficristo:
I really want a way to disable JSLint with the UI.
So I tried to show all the default extensions in the extension manager.
extensions-installed-default
The enable/disable buttons do not work yet. I reused the infrastructure already there, but to disable an extension it writes a file in the path of the extension: this is not possible for the default ones. An idea could be to write a json file in the user profile to track which extensions are enabled/disabled, so it will work even for the default extensions.
Is this approach correct?


abose:
Awsome!!
Tagging @larz0 for UI review.

larz0:
It's possible to disable JSLint in preferences right? If we think it's necessary to add a new tab just so that people can disable features then we should be using this wrench icon for the tab:

```
<svg width="22px" height="23px" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
        <path d="M3.351,0.413712646 C3.351,0.413712646 3.582,0.171712646 4.055,0.125712646 C4.516,0.0797126457 9.941,-0.543287354 11.703,1.48371265 C12.914,2.86671265 12.707,3.85771265 12.73,4.91771265 C12.74,5.98871265 13.076,8.65071265 13.42,9.58371265 C13.73,10.4707126 14.619,11.8637126 14.619,11.8637126 L21.174,18.9487126 C21.174,18.9487126 22.533,19.9277126 20.689,21.7827126 C18.857,23.6387126 17.959,22.2207126 17.959,22.2207126 L11.035,14.6177126 C11.035,14.6177126 10.16,13.3617126 8.894,13.0157126 C7.638,12.6707126 5.564,13.1537126 3.951,12.2437126 C2.326,11.3217126 0.195,7.91271265 0.127,7.31371265 C0.047,6.71471265 0,6.09271265 0,6.09271265 C0,6.09271265 0.103,5.93171265 0.254,6.03471265 C0.414,6.13871265 4.758,8.58071265 4.758,8.58071265 C4.758,8.58071265 5.875,8.88071265 7.914,6.81771265 C8.777,3.89171265 8.088,3.44271265 8.088,3.44271265 L3.605,0.666712646 L3.351,0.413712646" fill="#7F8181"></path>
    </g>
</svg>
```

MiguelCastillo:
@larz0 yeah you can, it's just as not as straight forward as a simple on/off flag...  https://github.com/adobe/brackets/wiki/How-to-Use-Brackets#list-of-supported-preferences.  Search for `linting.prefer`.

larz0:
@MiguelCastillo I wonder if it should be in Brackets Preference UI in the future since most of our users don't know that the features are actually extensions.

ficristo:
I wanted a user friendly way to disable JSLint, and I have taken this approach because I thought to be able to reuse the enable/disable infrastructure used by the "normal" extensions.
I do not mind to wontfix this if there are better alternatives.

abose:
+1 I like the idea of disabling default extensions 
Maybe the third tab name as `default extensions`?

valtlait:
@abose Or maybe `Features`, “since most of our users don't know that the features are actually extensions”.

abose:
Can default extensions be shown only when a preference is set? Disabling some default extensions will  break brackets. so maybe restrict it to people who knows what they are doing?

ficristo:
@abose My personal needs were been satisfied instead, more or less, with https://github.com/adobe/brackets/pull/11641.
Now I found less compelling to work on this: while I think that write a json file to track the extensions disabled, what I was talking in https://github.com/adobe/brackets/pull/11618#issue-102546738, is still something good, I'm less sure showing default extensions is good.
When I wrote this PR I wanted something easy to use, adding a preference to me is unwanted.
I thought instead in a whitelist / blacklist to show / hide specific extensions.
But as I said, I am not sure if I'll ever finish this.

valtlait:
Why these features are impemented as extensions if they cannot be disabled without breaking Breckets?

ficristo:
I don't think I'll ever finish this.

PR #11639, opened 2015-08-29
Integrated grunt madge

ficristo:
Used to find circular dependencies.
When (and if) all will be resolved it will be used to avoid regressions.


ingorichter:
This seems like a good idea to avoid circular deps, although we don't have any issues with it (and we successfully resolved them all the time).
This doesn't seem to add a lot of overhead to the build process.
Any other opinions? If not, let's merge it.

petetnt:
What's your opinion on this @ficristo? It LGTM, if you still think it's needed then lets merge it in.

zaggino (approved):

ficristo:
From what I remember, this could add a bit of noise in the build.
If nobody is gonna to fix the warnings (for now I don't think to work on them), maybe is better to wait.
And, FWIW, the latest release seems to require a more up-to-date version of grunt.

ficristo:
I don't work on this anymore.

PR #11640, opened 2015-08-29
Move getInitialQuery from FindUtils to FindBar

ficristo:
So to avoid a circular dependency.
(And IMHO is the right place)


abose:
+1 
@ficristo Thanks for the Change.
Merging.

PR #11664, opened 2015-09-05
Do not run jshint tasks twice

ficristo:
Small cleanup.


nethip:
Thanks @ficristo ! Looks good. Merging.

le717:
The `:all` subtask was present to run all the test suites (`src`, `test`, and `grunt`) all at one time instead of having to run each one individually. With this change, did you confirm running `grunt jshint` runs all the suites like before? Also, how were the JSHint subtasks being run twice as there was only one call to run them (`jshint:all`)?

ficristo:
@le717 If you omit the subtask ("all" in this case) jshint run all the tasks defined.
This change impact who run jshint alone (so they do not need to remember the subtask), and the "watch:all" task that now do not run jshint twice.
I confirm that |grunt test| run the same suites as before.

PR #11674, opened 2015-09-09
Add Node.js v4 stable to Travis config

phillipalexander:
NodeJS v4 was just released [NodeJS v4 announcement](https://nodejs.org/en/blog/release/v4.0.0/). This updates the CI build script to accommodate the release. Note: It should be "4", not "4.0" since Node will surely get a lot of minor updates before it goes LTS and reveal.js should test on the latest of them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: node_js
 sudo: false # use container-based Travis infrastructure
 node_js:
   - "0.10"
+  - "4"
 before_script:
   - npm install -g grunt-cli
   - npm install -g jasmine-node


nethip:
Thanks @phillipalexander for the PR! Most of node stuff that we use inside Brackets is still running on 10.24. This is long due for us to update our node to the latest version, but could not allot some time for that. 

One of the side effects of this change is that we now have two jobs running on Travis. One is with .10 and the other with 4.0. Ideally we should make sure we stick to one version of node across the product(extensions, build server, unit test suite e.t.c).

screen shot 2015-09-09 at 6 36 26 am

zaggino:
This needs to be upgraded too: https://github.com/adobe/brackets-shell/blob/bdc2d3a0db3ec6be83ee853366a157f449b9bc74/Gruntfile.js#L229-L231

And I think we should drop `0.10`.

Denisov21:
Now there is Node v4.2.2 (LTS) https://nodejs.org/en/blog/release/v4.2.2/!

petetnt:
And now there's Node v4.2.3 (LTS) available. Is this still planned for 1.6. or bumped up to 1.7?

zaggino:
First, someone needs to upgrade node in brackets-shell, this by itself doesn't help much.

petetnt:
@zaggino Yeah, I meant upgrading Node in general

Denisov21:
Now there is Node v4.4.3 (LTS) https://nodejs.org/en/blog/release/v4.4.3/!

ficristo:
We have upgraded to node 6.
@phillipalexander could you update this PR to test only node 6?

ficristo:
Sorry, I was eager to test against node 6 so I opened #12755.

Issue #11613, opened 2015-08-21
FindInFilesDomain uses fs.readFileSync


zaggino:
Hi guys, why do we use sync file read operation in the domain? https://github.com/adobe/brackets/blob/master/src/search/node/FindInFilesDomain.js#L200

Also, we should be checking if a file is not binary, with something like this: https://www.npmjs.com/package/isbinaryfile

This is blocking other domain files which often fail to load on Brackets' startup. Related issue https://github.com/zaggino/brackets-git/issues/1149

cc authors @abose @prksingh @rroshan1 @prafulVaishnav

abose:
There are usually large number of files in the project; when we queue large number of async file reads in the queue- it will result in node being unresponsive until all the files got indexed as the event loop was filled with file io related events. To circumvent this, we used synchronous file reads, as this will ensure that only one file op is pending at a time and each file read is interleaved with a `setimmediate` call so that the next file read event happens only after any pending events in queue is processed. It was mostly effective, we used to get frequent node failures immediately with the async file reads. But it is not as widely effective as we wanted it to be. 

What we can do is, 
1. right now 10 files are read as 1 synchronous operation- we can shorten this to just 1.
2. improving on 1 if required, we can set a timeout of 10(0) MilliSecs between each interleaves to yield to other events.
3. Move indexing to a different node process fork - This was our original implementation, but having to manage 2 node processes and sending/synchronising events across 2 process boundaries[brackets-node1-indexnode] was making it difficult to implement. It was also slower for instant search due to the data transfer required across 4 process boundaries.

If 1&2 fixes it would be the simplese fix and just leaves one node process.
 3 is a much more difficult route to take, but as we have stabilised search, the things left would be process management and data transfer optimisations, which would be manageable to do. 

Any other idea on optimizing this is welcome :)

abose:
The binary file check would be straight forward to implement.

zaggino:
Id suggest doing a queue of async reads, so next one starts when previous finishes. This way its non blocking and similarly fast to pure sync approach.

Plus Id put a binary check in place and maybe dont process files over a certain size, 10mb maybe?

abose:
[...]

PR #11675, opened 2015-09-09
Optimizing node for find in files

abose:
- [x] Dont search in files > 16MB 
- [x] only 1 file is indexed at a time instead of 10 in batch.
- [x] File filter has to be applied to indexable files - This will impact the search performance on file filter change[but could be a lesser thig to worry abut], as that will lead to un indexed files.

Tried but not effictive and reverted

```
- [x]  async file reads- this would slow down the indexer considerabily, but might unblock node.
          Reverted to sync file reads as for the brackets project alone, sync file reads complete in ~3 secs instead of 18 secs with async - 6x performance difference. 
- [x] 10 sec delay instead of 5 secs before indexer kick in at first start
         Reverted to 5 secs as this doesn't affect any behavior or solve any problems
```

### DIFF
--- a/src/search/FindInFiles.js
+++ b/src/search/FindInFiles.js
@@ -843,8 +843,7 @@ define(function (require, exports, module) {
                 var files = fileListResult,
                     filter = FileFilters.getActiveFilter();
                 if (filter && filter.patterns.length > 0) {
-                    filter = new RegExp(FileFilters.compile(filter.patterns));
-                    files = files.filter(function (entry) { return !filter.test(entry); });
+                    files = FileFilters.filterFileList(FileFilters.compile(filter.patterns), files);
                 }
                 files = files.filter(function (entry) {
                     return entry.isFile && _isReadableText(entry.fullPath);
--- a/src/search/FindInFiles.js
+++ b/src/search/FindInFiles.js
@@ -840,13 +840,17 @@ define(function (require, exports, module) {
         }
         ProjectManager.getAllFiles(filter, true, true)
             .done(function (fileListResult) {
-                var files = fileListResult
-                    .filter(function (entry) {
-                        return entry.isFile && _isReadableText(entry.fullPath);
-                    })
-                    .map(function (entry) {
-                        return entry.fullPath;
-                    });
+                var files = fileListResult,
+                    filter = FileFilters.getActiveFilter();
+                if (filter && filter.patterns.length > 0) {
+                    filter = new RegExp(FileFilters.compile(filter.patterns));
+                    files = files.filter(function (entry) { return !filter.test(entry); });
+                }
+                files = files.filter(function (entry) {
+                    return entry.isFile && _isReadableText(entry.fullPath);
+                }).map(function (entry) {
+                    return entry.fullPath;
+                });
                 FindUtils.notifyIndexingStarted();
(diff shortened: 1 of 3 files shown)


abose:
addresses part of #11613 
I have done some optimizations on it. Will update filter application soon.  
@zaggino Could you check if this pr improves something?

src/search/node/FindInFilesDomain.js
```diff
@@ -187,8 +188,26 @@ maxerr: 50, node: true */
...
 
+
     /**
-     * Get the contents of a file given the path
+     * Gets the file size in bytes.
+     * @param   {string} fileName The name of the file to get the size
+     * @returns {Number} the file size in bytes
+     */
+    function getFilesizeInBytes(fileName) {
+        var stats = {};
```
zaggino: you don't really need to init with an empty object here, it'll get overridden in try-catch anyway

zaggino:
Hi @abose, this is still indexing `node_modules` even though I have them excluded in search filter, which makes Brackets very unresponsive for me. Features that require node connection like `brackets-git` or `brackets-eslint` are literally freezing.

abose:
@zaggino Changed the file reads to async. Is there any improvement with this change?

zaggino:
well, actually this doesn't make much difference, probably better to stay with the sync mode ... i found a solution to my problem which is this snippet of code added here: https://github.com/adobe/brackets/blob/master/src/search/FindInFiles.js#L851
```
                console.log('before: files', files.length);
                var filter = FileFilters.getActiveFilter();
                if (filter && filter.patterns.length > 0) {
                    filter = new RegExp(FileFilters.compile(filter.patterns));
                    files = files.filter(function (entry) { return !filter.test(entry); });
                }
                console.log('after: files', files.length);
```
which in my project makes this result:
```
before: files 19649
after: files 397
```
it works well even with the 5 seconds timeout, 10 seconds are not required

abose:
Applied file filters before indexing.
I have reverted to sync file reads as for the brackets project alone, sync file reads complete in ~3 secs instead of 18 secs with async - 6x performance difference. Also reverted to 5 sec startup delay.

zaggino:
Looks good to me and solves problems with my projects, hope you don't mind @abose I put a minor tweak commit here, so please have a look. I haven't realized in my code snippet that I can use the filterFileList method which already exists.

One last thing probably, in case user doesn't have his filter configured, we should exclude `bower_components` and `node_modules` here by default, for less handy users, those two folders usually contain tons of files and very rarely someone wants to search them.

abose:
Thanks @zaggino for the changes.

src/search/FindInFiles.js
```diff
@@ -840,13 +840,16 @@ define(function (require, exports, module) {
...
         ProjectManager.getAllFiles(filter, true, true)
             .done(function (fileListResult) {
-                var files = fileListResult
-                    .filter(function (entry) {
-                        return entry.isFile && _isReadableText(entry.fullPath);
-                    })
-                    .map(function (entry) {
-                        return entry.fullPath;
-                    });
+                var files = fileListResult,
```
prafulVaishnav: fileListResult is unused.

prafulVaishnav:
LGTM..

PR #11692, opened 2015-09-12
Issue #11261: keep the search bar open

adrianhartanto0:
This change fixes issue #11261. It will prevent the Search Modal Bar from closing (if open) after a left mouse click on the main editor.This is accomplished by commenting out line 253 to line 257 and added an event listener in window.document.body for the event keydown in line 90. This is necessary to handle the closing of the search bar when ESC key is pressed when the main editor window has focus.

### DIFF
--- a/src/widgets/ModalBar.js
+++ b/src/widgets/ModalBar.js
@@ -87,6 +87,7 @@ define(function (require, exports, module) {
             this._autoClose = true;
             this._$root.on("keydown", this._handleKeydown);
             window.document.body.addEventListener("focusin", this._handleFocusChange, true);
+            window.document.body.addEventListener("keydown",this._handleKeydown,true);
                 
             // Set focus to the first input field, or the first button if there is no input field.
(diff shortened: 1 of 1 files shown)


abose:
I think it is better to close the search bar on click on the editor[current behavior]. once a search is done the context in the bar need not be retained as that is retained in the results panel. 
One modification to the current behavior would be, to `remember` the search phrase if there is no selection when the search bar is invoked again. I would favor that workflow than to stop the popup dismiss on focus change.

Denisov21:
I like this PR!

zaggino:
I'd make this a preference @abose - I'd quite like for search bar to stay open and keep searched strings highlighted while I make changes to the editor.

abose:
+1 for preference.
Would be good if `findbar` rememberes the last searched phrase if there is no selection.

ficristo:
Closing in favour of #11996 which seems a more complete fix.

PR #11693, opened 2015-09-12
Eslint

ficristo:
I added support for ESLint in grunt and removed JSHint.
I need to remove all the JSLint comments from the source files.
While I tried to preserve the old behavior (actually I want some more rules),
I need to fix some rules (the one with 0 in the eslintrc file).
And to decide what to do with the default linter in Brackets.
Personally I am in favor to remove it completely, people can use the extensions instead.
What you all think?


abose:
Why do you prefer eslint over jshint for use with brackets?

petetnt:
@abose see this discussion about replacing JSLint in #11644

abose:
got it!
@ficristo did you do a grunt build to see if there are any build failures due to eslint problems? There would be jslint annotations listed all over the code base. Is it compatible with eslint annotations?

ficristo:
@abose I added ESLint because I thought was the preferred one by the Brackets community. As pointed out by @petetnt I saw in some issues / forum that was the preferred choice. 
If you use JSHint and still want stylistic checks you should use JSCS too.
I run grunt eslint with the set of rules I written and it passes. As I written in the first post there are some rule disabled.
I tried grunt build too and I didn't noticed any problem with JSLint.
I didn't try inline comments, but personally I want to remove all of them.

IanVS:
@ficristo You may want to consider adding `"extends": "eslint:recommended"` to your `.eslintrc`, since as of version `1.0.0`, all rules are disabled by default (see http://eslint.org/docs/user-guide/migrating-to-1.0.0).  Recommended rules are denoted by "`(recommended)`" in http://eslint.org/docs/rules/.

src/config.json
```diff
@@ -20,7 +20,7 @@
         "extension_url": "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default": true,
         "build_timestamp": "",
-        "healthDataServerURL" : "https://healthdev.brackets.io/healthDataLog"
+        "healthDataServerURL": "https://health.brackets.io/healthDataLog"
```
zaggino: this line shouldn't have changed I guess

ficristo:
@IanVS I tried `"extends": "eslint:recommended"` but there are still too many errors, and I think to prefer to list all the rules.

I would like to land the patch as is and file followup to enable (or remove) the current set of rule.
Eventually we also can add other rules.

abose:
Moved to milestone 1.6 so that we could work on comprehensive support of es6 related pr's.

le717:
Just chiming in here to say a while back I [began rewriting](https://github.com/le717/brackets/commits/eslint) the JSHint rules for ESLint (+ incorporating the [style guide wiki page](https://github.com/adobe/brackets/wiki/Brackets-Coding-Conventions)). If it helps any, [here](https://github.com/le717/brackets/blob/cf69b4730a8cdb0415ad2206329c819c3d5eda54/eslint.json) are set of rules I had already implemented. The Brackets source failed heavily against them (partially from incomplete rule porting) but it was working.

zaggino:
hi @le717 , how does `nodeca/indent` differ from eslint's `indent` rule?

petetnt:
@zaggino  ESLint didn't get indendation support until version 0.14 so the plugin was neccessary. It can safely be replaced by native indent though.

le717:
@zaggino I believe @petetnt is correct. IIRC, ESLint did not have native intent support, so I added that plugin to enforce it (as it is not already being done and manually checking it on code reviews when it could be automatic is not the wisest idea).

zaggino:
yes but eslint is now pass 1.0 and has indent support so we should use that one instead of a plugin

le717:
@zaggino Yes, it would be best to use native indent support. :)

ficristo:
Is there a way I can help to move forward this?

abose:
tagging @swmitra

zaggino:
the PR seems good, is there any reason not to merge this?

petetnt:
LGTM :+1:

abose:
Verified build works. Merging.

abose:
Thanks @ficristo 
Should ESlint be prefferd for the lint panel in brackets?

zaggino:
@abose Yes, I believe so

petetnt:
@abose That would require integration of the ESLint (similarly how JSLint is integrated now under https://github.com/adobe/brackets/tree/d939740fd5c368c347ac6a8c7cef2941981aa2fb/src/extensions/default. 

@zaggino has the great ESLint extension so I think he has the most insight on how much work it would require. Personally I would love it :bouquet:

zaggino:
It's easy to get eslint integrated direclty into Brackets core, I can do a PR for this if you want guys.

abose:
+1 for ESLint in default extensions in core.

ficristo:
I agree to switch from JSLint to ESLint as default linter in core.

abose:
one part of https://github.com/adobe/brackets/issues/11644 will also be addressed by the switch.

petetnt:
I created #11984 to track the remaining steps of the ESLint integration

PR #11700, opened 2015-09-14
Extension manager now scrolls to top when search filters are added/removed

nhickok:
In reference to bug #10611:

I took into account the feedback on the previous attempt to fix this bug (#11178) and implemented the suggested changes (i.e. added `$(".modal-body", $dlg).scrollTop(0);` below both calls to `view.filter(...)`). All unit tests pass and the feature works as intended when search filters are added/removed.

### DIFF
--- a/src/extensibility/ExtensionManagerDialog.js
+++ b/src/extensibility/ExtensionManagerDialog.js
@@ -337,6 +337,7 @@ define(function (require, exports, module) {
             $search.val("");
             views.forEach(function (view, index) {
                 view.filter("");
+                $(".modal-body", $dlg).scrollTop(0);
             });
 
             if (!updateSearchDisabled()) {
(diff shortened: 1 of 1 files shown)


ingorichter:
@stowball do you have a minute to verify that this fixes the issue for you?

ficristo:
This seems a duplicate of #11378

NGHGithub:
You're right, they are both the same. I must not have noticed it. I'll close this pull request.

stowball:
Yep, this works as expected. I wonder whether we should now be caching `$(".modal-body", $dlg)` in to its own variable since its called 5 times?

ingorichter:
Yes, assign to it's own variable is a good idea.

mansimarkaur:
Does `$(".modal-body", $dlg)` still need to be assigned to it's own variable?

ficristo:
Yes, do you want to take a look?
It seems to me that @darkcoderrises will not fix it in his PR at #11378, so go ahead if you want.

mansimarkaur:
@stowball I think `$(".modal-body", $dlg)` is being used thrice only in ExtensionManagerDialog.js and not 5 times anywhere. Please correct me if I'm wrong.

PR #11641, opened 2015-08-29
Remember linter collapsed

ficristo:
Remember if a linter was collapsed in the panel.

PR #11721, opened 2015-09-18
Fixed a bug when there is only a linter error and it is collapsed

ficristo:
If you:
1. have multiple linters
2. you collapsed one
3. after, you have only errors relative of the collapsed one

you could end up with the panel error shown without see any error.


ficristo:
@abose this fixes a bug of #11641. Can you double check? Thank you.

zaggino:
tested & confirmed, merged

Issue #11594, opened 2015-08-15
Full screen code editor (Suggestion)


csalmeida:
Hello guys,

After using Brackets since the beginning I noticed that many times I really would like to be coding without any distractions or menus, either from the OS or the text editor it-self.

With this in mind I wonder if it would be possible to add a complete full screen mode to the editor?

**Thank you for the time you devote to this project.**

Best,
csalmeida

petetnt:
Fullscreen API support for CEF was added [quite recently (cef@d820080)](https://bitbucket.org/chromiumembedded/cef/commits/d820080) so it's possible and sounds like a good idea to me :+1:

csalmeida:
Amazing! Hopefully it becomes a reality later on. 
Thank you.

underctrl:
Hello All - Any chance this could be implemented soon? I'd _love_ to be able to put Brackets into fullscreen...

_Many_ applications already do this when you press F11 - Chrome, Firefox, Notepad++, Atom, etc - so it feels very much like a basic missing feature from Brackets...

Thanks!

alekexe:
YEARS later and Bracket still lacks this feature - SAD!

xypd:
Not sure I understand, I've been using it in fullscreen mode for years. I have a split window and hide everything but the code.

This is on a Mac so not sure if these complaints are windows or linux, since I don't seem to see that info.

PR #11732, opened 2015-09-23
Toggle panels and no-Distraction mode 

abose:
Full code view with all other panels/toolbars hidden.
Also minor refactoring of variable names for toggle/hide for sidebar.
shortcut : ctrl+shift+` for toggle panels
shortcut : ctrl+shift+2 for no distraction mode
Also added panel management API's to workspace manager.

@nethip 
- [ ] Add unit tests

### DIFF
--- a/src/extensions/default/NoDistractions/main.js
+++ b/src/extensions/default/NoDistractions/main.js
@@ -36,7 +36,7 @@ define(function (require, exports, module) {
         KeyBindingManager   = brackets.getModule("command/KeyBindingManager");
 
     // Constants
-    var PREFS_PURE_CODE           = "pureCode",
+    var PREFS_PURE_CODE           = "noDistractions",
         CMD_TOGGLE_PURE_CODE      = "view.togglePureCode";
 
     //key biding keys
--- a/src/extensions/default/NoDistractions/main.js
+++ b/src/extensions/default/NoDistractions/main.js
@@ -34,12 +34,15 @@ define(function (require, exports, module) {
         PreferencesManager  = brackets.getModule("preferences/PreferencesManager"),
         ViewUtils           = brackets.getModule("utils/ViewUtils"),
         KeyBindingManager   = brackets.getModule("command/KeyBindingManager"),
+        HealthLogger        = brackets.getModule("utils/HealthLogger"),
         WorkspaceManager    = brackets.getModule("view/WorkspaceManager");
 
     // Constants
     var PREFS_PURE_CODE           = "noDistractions",
         CMD_TOGGLE_PURE_CODE      = "view.togglePureCode",
-        CMD_TOGGLE_PANELS         = "view.togglePanels";
+        CMD_TOGGLE_PANELS         = "view.togglePanels",
+        HEALTH_NO_DISTRACTION     = "noDistractionModeUsed",
+        HEALTH_TOGGLE_PANELS      = "togglePanels";
 
     //key binding keys
     var togglePureCodeKey         = "Ctrl-Shift-2",
@@ -66,6 +69,7 @@ define(function (require, exports, module) {
      */
     function _togglePureCode() {
         PreferencesManager.set(PREFS_PURE_CODE, !PreferencesManager.get(PREFS_PURE_CODE));
+        HealthLogger.setHealthDataLog(HEALTH_NO_DISTRACTION, true);
     }
 
     /**
@@ -110,6 +114,7 @@ define(function (require, exports, module) {
      * The already hidden panel should not be shown in the specific case for better UX.
      */
     function _togglePanels() {
+        var togglePanelCount;
         panelsToggled = !panelsToggled;
         if (panelsToggled) {
             _hidePanelsIfRequired();
@@ -118,6 +123,11 @@ define(function (require, exports, module) {
         } else if (!layoutUpdated) {
             _showPanelsIfRequired();
         }
+
+        //log health data
+        togglePanelCount = HealthLogger.getHealthDataLog(HEALTH_TOGGLE_PANELS);
+        togglePanelCount = (typeof togglePanelCount === 'number') ? togglePanelCount + 1 : 0;
+        HealthLogger.setHealthDataLog(HEALTH_TOGGLE_PANELS, togglePanelCount);
     }
 
     PreferencesManager.definePreference(PREFS_PURE_CODE, "boolean", false, {
--- a/src/extensions/default/NoDistractions/main.js
+++ b/src/extensions/default/NoDistractions/main.js
@@ -38,14 +38,18 @@ define(function (require, exports, module) {
 
     // Constants
     var PREFS_PURE_CODE           = "noDistractions",
-        CMD_TOGGLE_PURE_CODE      = "view.togglePureCode";
+        CMD_TOGGLE_PURE_CODE      = "view.togglePureCode",
+        CMD_TOGGLE_PANELS         = "view.togglePanels";
 
     //key biding keys
-    var togglePureCodeKey         = "Ctrl-Shift-`",
-        togglePureCodeKeyMac      = "Cmd-Shift-`";
+    var togglePureCodeKey         = "Ctrl-Shift-2",
+        togglePureCodeKeyMac      = "Cmd-Shift-2",
+        togglePanelsKey           = "Ctrl-Shift-`",
+        togglePanelsKeyMac        = "Cmd-Shift-`";
 
     //locals
-    var _previouslyOpenPanelIDs = [];
+    var _previouslyOpenPanelIDs = [],
+        panelsToggled = false;
 
     /**
      * @private
@@ -89,6 +93,15 @@ define(function (require, exports, module) {
         _previouslyOpenPanelIDs = [];
     }
 
+    function _togglePanels() {
+        panelsToggled = !panelsToggled;
+        if(panelsToggled) {
+            _hidePanlesIfRequired();
+        } else {
+            _showPanlesIfRequired();
+        }
+    }
+
     PreferencesManager.definePreference(PREFS_PURE_CODE, "boolean", false, {
         description: Strings.DESCRIPTION_PURE_CODING_SURFACE
     });
@@ -111,11 +124,13 @@ define(function (require, exports, module) {
      */
     function initializeCommands() {
         CommandManager.register(Strings.CMD_TOGGLE_PURE_CODE, CMD_TOGGLE_PURE_CODE, _togglePureCode);
+        CommandManager.register(Strings.CMD_TOGGLE_PANELS, CMD_TOGGLE_PANELS, _togglePanels);
 
-        Menus.getMenu(Menus.AppMenuBar.VIEW_MENU).addMenuItem(CMD_TOGGLE_PURE_CODE, "", Menus.AFTER, Commands.VIEW_HIDE_SIDEBAR);
-        //Menus.getMenu(Menus.AppMenuBar.VIEW_MENU).addMenuItem(CMD_TOGGLE_PURE_CODE);
+        Menus.getMenu(Menus.AppMenuBar.VIEW_MENU).addMenuItem(CMD_TOGGLE_PANELS, "", Menus.AFTER, Commands.VIEW_HIDE_SIDEBAR);
+        Menus.getMenu(Menus.AppMenuBar.VIEW_MENU).addMenuItem(CMD_TOGGLE_PURE_CODE, "", Menus.AFTER, CMD_TOGGLE_PANELS);
 
         KeyBindingManager.addBinding(CMD_TOGGLE_PURE_CODE, [ {key: togglePureCodeKey}, {key: togglePureCodeKeyMac, platform: "mac"} ]);
+        KeyBindingManager.addBinding(CMD_TOGGLE_PANELS, [ {key: togglePanelsKey}, {key: togglePanelsKeyMac, platform: "mac"} ]);
     }
 
     initializeCommands();
--- a/src/extensions/default/NoDistractions/main.js
+++ b/src/extensions/default/NoDistractions/main.js
@@ -95,7 +95,7 @@ define(function (require, exports, module) {
 
     function _togglePanels() {
         panelsToggled = !panelsToggled;
-        if(panelsToggled) {
+        if (panelsToggled) {
             _hidePanlesIfRequired();
         } else {
             _showPanlesIfRequired();
--- a/src/extensions/default/NoDistractions/main.js
+++ b/src/extensions/default/NoDistractions/main.js
@@ -33,7 +33,8 @@ define(function (require, exports, module) {
         Strings             = brackets.getModule("strings"),
         PreferencesManager  = brackets.getModule("preferences/PreferencesManager"),
         ViewUtils           = brackets.getModule("utils/ViewUtils"),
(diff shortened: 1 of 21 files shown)


abose:
Should it be hide panels or toggle panels? toggle seems to be troublesome at times.

abose:
@petetnt @ficristo @zaggino

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    function _togglePureCode() {
+        PreferencesManager.set(PREFS_PURE_CODE, !PreferencesManager.get(PREFS_PURE_CODE));
+    }
+
+    /**
+     * hide all open panels
+     */
+    function _hidePanlesIfRequired() {
```
petetnt: Small typo: `Panles` -> `Panels"

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,138 @@
...
+                WorkspaceManager.getPanelForID(panelIDs[i]).hide();
+                _previouslyOpenPanelIDs.push(panelIDs[i]);
+            }
+        }
+    }
+
+    /**
+     * show all open panels that was previously hidden by _hidePanlesIfRequired()
+     */
+    function _showPanlesIfRequired() {
```
petetnt: Small typo: `Panles` -> `Panels"

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,138 @@
...
+        for (i = 0; i < panelIDs.length; i++) {
+            if (WorkspaceManager.getPanelForID(panelIDs[i]).isVisible()) {
+                WorkspaceManager.getPanelForID(panelIDs[i]).hide();
+                _previouslyOpenPanelIDs.push(panelIDs[i]);
+            }
+        }
+    }
+
+    /**
+     * show all open panels that was previously hidden by _hidePanlesIfRequired()
```
petetnt: Same here: Small typo: `Panles` -> `Panels"

petetnt:
I think that `hidePanels` should be sufficient, at least until the panels are grouped together with something like #11586. Panels get currently opened and closed all the time so it would be tons of trouble to keep up with their states at this point IMO.

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,138 @@
...
+        ViewUtils           = brackets.getModule("utils/ViewUtils"),
+        KeyBindingManager   = brackets.getModule("command/KeyBindingManager"),
+        WorkspaceManager    = brackets.getModule("view/WorkspaceManager");
+
+    // Constants
+    var PREFS_PURE_CODE           = "noDistractions",
+        CMD_TOGGLE_PURE_CODE      = "view.togglePureCode",
+        CMD_TOGGLE_PANELS         = "view.togglePanels";
+
+    //key biding keys
```
petetnt: binding

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,138 @@
...
+        } else {
+            _showPanlesIfRequired();
+        }
+    }
+
+    PreferencesManager.definePreference(PREFS_PURE_CODE, "boolean", false, {
+        description: Strings.DESCRIPTION_PURE_CODING_SURFACE
+    });
+
+    PreferencesManager.on("change", PREFS_PURE_CODE, function () {
```
petetnt: `_showPanelsIfRequired, _hidePanelsIfRequired`

abose:
Thanks @petetnt for reviewing.
Made the change to toggle panel only if workspace has not changed after a hide panels command is issued. This looked intuitive to me. 
The documentation for the preference was intentionally worded as `Code view only`. The string describes what `no distractions` mode in simple terms.

src/utils/Resizer.js
```diff
@@ -61,12 +62,13 @@ define(function (require, exports, module) {
     // Load dependent modules
     var AppInit                 = require("utils/AppInit"),
         EventDispatcher         = require("utils/EventDispatcher"),
+        viewUtils               = require("utils/ViewUtils"),
```
ficristo: nit: ViewUtils, in ProperCase

ficristo:
In the `View` menu, `Toggle Panels` menu doesn't show the shortcut.

@abose I briefly tested the patch and seems fine.
I agree with @petetnt for the panels.

abose:
@zaggino There seems to be a bug. If you hide the git panel, the blue active icon for the git extension still remains in the main-toolbar. So added events when panels are shown or hidden so that extensions can listen to panel close events.

zaggino:
thanks for the info @abose

petetnt:
Tested this out again against https://github.com/adobe/brackets/commit/51c023279576b78d8f61886991a50c10609b72df , love the No distractions mode. 

There's a small issue with toggle panels though (other than the shortcut not showing up/working)

1. Open foo.js
2. Create a validation error (I am using `brackets-eslint`)
3. Click Toggle Panels -> nothing happens (expected: panel closes)
4. Click Toggle Panels -> panel closes.

That issue that affects `brackets-git` affects other Extensions too, such as `brackets-todo`. Is there a way to (at least temporarily / during 1.6.) to throw a warning if panel was closed but the button state remained + link to say Brackets-Wiki where those events would be explained). That would provide a easy way for extension makers that use panels to know that they should update their extensions).

ryanstewart:
Looks good to me. And agree, that the no distraction mode is a nice touch. 

Do we have an idea of how often people will want to jump into this? Main question is wether we can set an easier shortcut. 

I also like that it doesn't go full screen the way Sublime Text does, but any thought about whether we need that?

petetnt:
@ryanstewart there was an issue about "true" fullscreen some while ago (#11594) and when checking out that one I saw that CEF got updated to support fullscreen quite recently so it could be implemented relatively easily (needs a `brackets-shell` update though) separate from this feature. :+1:

ryanstewart:
@petetnt ahh, awesome. And I'm definitely not advocating for that, just wanted to throw it out there. Sounds like it's been covered though!

abose:
`ctrl+shift+~` falls in the same vertical last line in the key board, so i thought it would be easily accesible to dismiss panels with one hand.

abose:
@petetnt i have created https://github.com/adobe/brackets/issues/11924
This is an issue in extract extension too.

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,155 @@
+/*
+ * Copyright (c) 2013 Adobe Systems Incorporated. All rights reserved.
```
nethip: Change copyright from 2013 to 2015

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,155 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+/*jslint vars: true, plusplus: true, devel: true, nomen: true, regexp: true, indent: 4, maxerr: 50 */
```
nethip: Please review if all of these params to jslint are valid for this file.

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,155 @@
...
+    var _previouslyOpenPanelIDs = [],
+        panelsToggled = false,
+        layoutUpdated = false;
+
+    /**
+     * @private
+     *
+     * Updates the command checked status based on the preference name given.
+     *
+     * @param {string} name Name of preference that has changed
```
nethip: Should we update the documentation as I don't see any {string} parameter.

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,155 @@
...
+     */
+    function _updateCheckedState() {
+        CommandManager.get(CMD_TOGGLE_PURE_CODE).setChecked(PreferencesManager.get(PREFS_PURE_CODE));
```
nethip: A null check would be better here.

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    function _togglePureCode() {
+        PreferencesManager.set(PREFS_PURE_CODE, !PreferencesManager.get(PREFS_PURE_CODE));
+    }
+
+    /**
+     * hide all open panels
```
nethip: Since you are following the adding "@private" to the function documentation, let us be consistent with labeling all the  private functions with "@private"

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,155 @@
...
+     */
+    function _hidePanelsIfRequired() {
+        var panelIDs = WorkspaceManager.getAllPanelIDs(), i = 0;
+        _previouslyOpenPanelIDs = [];
+        for (i = 0; i < panelIDs.length; i++) {
+            if (WorkspaceManager.getPanelForID(panelIDs[i]).isVisible()) {
```
nethip: It is better to have a `null` check here.

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,155 @@
...
+        PreferencesManager.set(PREFS_PURE_CODE, !PreferencesManager.get(PREFS_PURE_CODE));
+    }
+
+    /**
+     * hide all open panels
+     */
+    function _hidePanelsIfRequired() {
+        var panelIDs = WorkspaceManager.getAllPanelIDs(), i = 0;
+        _previouslyOpenPanelIDs = [];
+        for (i = 0; i < panelIDs.length; i++) {
```
nethip: `forEach`?

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,155 @@
...
+                _previouslyOpenPanelIDs.push(panelIDs[i]);
+            }
+        }
+    }
+
+    /**
+     * show all open panels that was previously hidden by _hidePanelsIfRequired()
+     */
+    function _showPanelsIfRequired() {
+        var panelIDs = _previouslyOpenPanelIDs, i = 0;
```
nethip: No need of an extra variable here. You could further simplify this with `forEach`

src/extensions/default/NoDistractions/main.js
```diff
@@ -0,0 +1,155 @@
...
+        for (i = 0; i < panelIDs.length; i++) {
+            if (WorkspaceManager.getPanelForID(panelIDs[i])) {
```
nethip: Store the result from `getPanelForID` instead of calling the same function below.

src/nls/root/strings.js
```diff
@@ -768,5 +771,6 @@ define({
     "DESCRIPTION_OPEN_PREFS_IN_SPLIT_VIEW"           : "false to disable opening preferences file in split view",
     "DESCRIPTION_OPEN_USER_PREFS_IN_SECOND_PANE"     : "false to open user preferences file in left/top pane",
     "DEFAULT_PREFERENCES_JSON_HEADER_COMMENT"        : "/*\n * This is a read-only file with the preferences supported\n * by {APP_NAME}.\n * Use this file as a reference to modify your preferences\n * file \"brackets.json\" opened in the other pane.\n * For more information on how to use preferences inside\n * {APP_NAME}, refer to the web page at https://github.com/adobe/brackets/wiki/How-to-Use-Brackets#preferences\n */",
-    "DEFAULT_PREFERENCES_JSON_DEFAULT"               : "Default"
+    "DEFAULT_PREFERENCES_JSON_DEFAULT"               : "Default",
+    "DESCRIPTION_PURE_CODING_SURFACE"                : "true to enable Code view only mode, Every other UI element in brackets will be hidden"
```
nethip: can we rephrase this to `"true to enable code only mode and hide all other UI elements in {APP_NAME}"`

src/extensions/default/NoDistractions/main.js
abose: As we just got the panel ID's from the above statement, this should never be null at this line. It would be better that a NULL access exception be raised here in that case, else will be hard to debug if things fails- which it should not :crystal_ball: .

src/view/WorkspaceManager.js
```diff
@@ -191,6 +201,7 @@ define(function (require, exports, module) {
      */
     Panel.prototype.hide = function () {
         Resizer.hide(this.$panel[0]);
+        exports.trigger(EVENT_WORKSPACE_PANEL_HIDDEN, this.panelID);
```
nethip: Instead of having two events `EVENT_WORKSPACE_PANEL_HIDDEN` and `EVENT_WORKSPACE_PANEL_SHOWN`, can we have just one event and pass in the param whether the panel is hidden or shown?

abose: I generally prefer named stateless events. That is usually much easier to deal with.

src/extensions/default/NoDistractions/main.js
nethip: We will never know whether we are given a null panel or an actual panel as that entirely depends on the API that is getting called. A check is surely of no harm here.

src/view/WorkspaceManager.js
nethip: I am not sure if having a new event is easier to deal with. With the param passing approach we have one event less. I don't see a compelling reason why we would want two events when one can suffice the requirement.

nethip:
Couple of quick comments.

* We need to add unit tests for the new *No Distraction* feature. I saw some changes in the unit tests but I am not sure if they are going to cover the new workflows
* We need to add health logs for the new functionality. Also make sure to update our Health Data wiki with this new data
* It is better to have `null` checks at places where we calling APIs, as we would never know if the API returns an actual value or `null`. If you want to catch errors here, use a null check along with `console.warn/console.error`. That should raise errors in the console, in case of failures
* Please update the documentation, wherever required
* Make sure we update our API docs with the changes/additions

Overall it looks good :+1: It is good to merge, provided the review comments are addressed.

src/view/WorkspaceManager.js
abose: In the case, say if we raise a panel visibilitychanged event, then we would need to write an intermediate function to route that event to `PanelShown` and `panelHidden` functions. That can be avoided if we can directly tag a specific `trigger` to an `action`. Maybe just a choice of writing style :)

abose:
@nethip thanks for reviewing. Made the changes.
Will add unit tests soon.

nethip:
@abose No problem :) I took the liberty to add unit tests checkbox to your first comment. Also I created a waffle card for tracking the unit tests for this feature I am going ahead merging this.

Thanks for this feature @abose !

seehup:
Comment on pull request

seehup:
Comment on pull request

Issue #11726, opened 2015-09-21
Move PathUtils to Brackets core code?


ficristo:
I wondered what was PathUtils and found that was coming from
https://github.com/jblas/path-utils
It does not seem used outside Brackets, at least there isn't a package.json or bower.json.
Seems written by and Adobe guy, so, can be relicensed (I think this is a must but not sure) and imported?
My main motivation is very simple: I do not like global variables.

petetnt:
:+1: for this.

Semi-offtopic, but I think that that regexp is amazing / like the comment says, scary:

`/^(((([^:\/#\?]+:)?(?:(\/\/)((?:(([^:@\/#\?]+)(?:\:([^:@\/#\?]+))?)@)?(([^:\/#\?\]\[]+|\[[^\/\]@#?]+\])(?:\:([0-9]+))?))?)?)?((\/?(?:[^\/\?#]+\/+)*)([^\?#\.]*(?:\.[^\?#\.]+)*(\.[^\?#\.]+)|[^\?#]*)))?(\?[^#]+)?)(#.*)?/,`

abose:
:+1: 
Pathutils seems to be a submodule. As we are already shipping with the lib, I don't think licensing would be an issue if we are to change the folder where the library is placed in code- as long as the copyright notices are left untampered.

le717:
We would just need to be sure to continue exposing the global variable with a deprecation warning (like we do with the global CodeMirror) in order to prevent sudden extension breaking.

I am trying to think where `PathUtils` would be relocated to. I'm thinking either `src/file` or `src/utils`. The former already has `FIleUtils` there, but once the old file system is (eventually) removed, it will be the only module in that folder, but the latter might work better (since it is a utils module).

zaggino:
Guys, this brings out a question for me ... why do we use git submodules instead of bower/npm modules? For codemirror, mustache, ...

busykai:
Submodules allow you keep forks of the code and make modifications to them without publishing anywhere. (Take `CodeMirror` or `jslint` as an example.) I actually quite disagree that #11734 had to be merged. There's no single technical reason for that: `path-utils` being submodule has nothing to do with `PathUtils` being global? If you don't like globals, then the only thing needed to be done was to configure path for requirejs.

Also, commits in #11734 do not specify where it's been taken from, at which version/sha it the snapshot has been taken.

zaggino:
[...]

PR #11734, opened 2015-09-23
Move PathUtils to Brackets core code

ficristo:
Fix for #11726.


zaggino:
Reviewed, looks good to me.

ficristo:
I think, based on the discussion on #11726, this needs to be reverted.
It can always be relanded, and we can even add better comments.

peterflynn:
@zaggino Just catching up on 1.5 PRs... is this a breaking change for any extensions that were using PathUtils?  (The path has changed).  Did you check the ExtensionGrabber tree to see if any / how many extensions wold be broken by this?

peterflynn:
Ah nm, I see that this was reverted a day later.

Issue #11448, opened 2015-07-23
Can I force load a file in a specific split window?


pliablepixels:
Here is the issue I am facing:
I am using Brackets for a large project and use split view (vertical) very often. The problem I am facing is that when I first open a file from the file tree, it gets associated to the split window that is active. That is good.

However, after that, that file gets permanently associated to that window pane and I cannot really open the file in the other window pane unless I go to the sidebar, delete that file from the "Left" or "Right" pane and then open it in the required pane.  The problem here is that  the sidebar that shows Left and Right is vertically quite small, so its often impossible to know which file is pre-associated to which pane.

What I'd love to do is irrespective of which pane it may have opened in before, if I got to the file tree and click (or double click) on a file, it should open in the pane I want it to open in, without having to remove it from the associated pane in 'Left' or 'Right'

Is there a way to do this?

thanks

petetnt:
Duplicate of #11308

MarcelGerber:
Tip: You can also drag files between panes. That's not exactly the solution to what you're asking for, but I guess it'll help anyway.

pliablepixels:
Marcel yes that is what I do today. The problem is the left and right pane
are not resizeable so I often use the file tree - without knowing it's
already associated to a pane which is not visible in the work are pane which
is a pain :)

There is a plugin that does make it resizeable but it's very buggy.

Issue #11345, opened 2015-07-04
Splitscreen UX


Wikunia:
It would be awesome to right click on a file and choose the name of the splitted screen to display the content in that screen. Or a drag & drop feature. At the moment I don't know how to see the content of an unopened file on the "second" screen.

petetnt:
Similar to #11308

prafulVaishnav:
@Wikunia Single left click on the file in project tree will show the content of the file in the focused screen. You can use (View > Vertical/Horizontal Split) and drag & drop the file to desired split view. If second screen is in focus then left click on file in project tree will be opened in that screen.

Issue #11746, opened 2015-09-26
feature: close split view automatically


zaggino:
steps:
1) open preferences file
2) split view opens
3) click to close the `brackets.json` in the right pane, nothing else is remaining in the right pane

i believe, that the split view should now close automatically, given it was also automatically opened

petetnt:
Makes sense in the `preferences` scenario, I think that the preferences split view should be more isolated anyway. Auto-closing the panel when it's possible to modify the views while the preferences are open is a bit tricky too. Consider something like this:

1. Open preferences file
2. Split view opens
3. Drag another file to right panel
4. Close `brackets.json` -> Right view should stay open right?

But what if

4. Drag the file back to left panel
5. Close `brackets.json` - nothing stays in the right panel but it shouldn't auto close as it has been "*dirty*" - right?

zaggino:
Well, I'd implement it in a way that if split-view was automatically opened (forced on me) it also should automatically close. If it was opened manually, it should never auto-close. I don't really like split view and how it's forcing itself anytime I want to change something in preferences.

petetnt:
This was added in #11749 right?

PR #11749, opened 2015-09-27
[New feature] Add flip-view and close buttons to pane-headers

petetnt:
This new feature improves the Splitscreen UX by adding buttons to panel headers that allow user to quickly flip the current view to another pane. It's not a complete solution, but another alternative to dragging the files over to another panel which can be a bit cumbersome, as evidenced in #11308, #11448, #11345 and others.

When Split View is enabled, arrow buttons are added to the panel that indicate flipping 

image

Clicking such button flips the view to the other panel, revealing the last used file in the pane it was flipped from

image

If there's no files open in the new other pane, the button obviously isn't there

image

This of course works in vertical split mode too:

image

The first commit doesn't include unit tests because Jasmine Spec Runner won't work for whatever reason: while I am looking at that issue, any feedback, improvements or questions are welcome.

~~edit: Build failed because the API rate limit exceeded~~

edit: 2
The panes now contain a `close buttons` too as similar to what @zaggino suggested.

image

Clicking the `x` closes the file, or prompts for confirmation if a file is dirty. If pane contains no files, the `close button` collapses the views by setting the layout scheme to `1x1`


zaggino:
Love it, gave me an idea -> `X` mark in the corner to close split view quickly (one click instead of two), would solve my https://github.com/adobe/brackets/issues/11746 problem.

src/view/Pane.js
```diff
@@ -159,17 +159,33 @@ define(function (require, exports, module) {
         EventDispatcher     = require("utils/EventDispatcher"),
         FileSystem          = require("filesystem/FileSystem"),
         InMemoryFile        = require("document/InMemoryFile"),
+        //WorkingSetView      = require("project/WorkingSetView"),
```
le717: Forget to remove a line?

src/styles/brackets.less
```diff
@@ -486,11 +486,58 @@ a, img {
...
+            padding-right: 4px;
+            padding-bottom: 2px;
+            padding-left: 4px;
+            margin-left: 0;
+            margin-bottom: 0;
+            .sprite-icon(0, 0, 13px, 13px, "images/flip-view-icons.svg");
+            background-origin: content-box;
+            -webkit-transform: translateZ(0); // forces GPU mode for better filter rendering on retina
+            transform: translateZ(0); // future proofing
+            -webkit-filter: drop-shadow(0 1px 0 rgba(0,0,0,0.36));
```
le717: `filter` has a standardized version. It'd probably be best to add that too like you did with `transform`.

petetnt: Thanks @le717, fixed those :+1: 

RE: the filter declaration, there are some other components using the plain vendor-prefixed versions of ones that have standardized ones available, might be worth checking out.

petetnt:
@zaggino yeah that is an great idea to me!

petetnt:
Added an unit test

petetnt:
The panes now contain a `close buttons` too as similar to what @zaggino suggested.

image

Clicking the `x` closes the file, or prompts for confirmation if a file is dirty. If pane contains no files, the `close button` collapses the views by setting the layout scheme to `1x1`

zaggino:
few comments:
1) i like how the close highlights when hovered, can you make the same with the arrows? right now arrow icon doesn't highlight but some weird border around it highlights
2) the icons doesn't have to be there all the time, maybe then can appear only when I hover over the title area? (i mean the `.pane-header` element)
3) when i click close icon on the last file in the right panel, it doesn't close but i have to click again to close the empty panel?
4) if i have one file in the left panel and one file in the right panel, when i close the left file and then close the left pane, the right file disappears (it works fine closing the right panel)

petetnt:
Thanks for the comments @zaggino. I'll look into those asap, esp. 4), not sure what's up with that :hammer: It seems that [`getCurrentlyViewedFile()`](https://github.com/petetnt/brackets/blob/petetnt/flip-view-buttons/src/view/Pane.js#L1190) returns the file in the active pane and just clicking on the `x` doesn't active the pane itself -> wrong file closes.

I was putting some behavior for 2) and 3) behind some prefs: `hover`, `always` and `off` for the panel buttons (default being `hover`) and something like `CLOSE_PANE_AFTER_LAST` for the panes:

zaggino:
Ah right, I didn't read the code before testing this so I didn't knew about the various preferences. But show on `hover` seems to be reasonable default.

petetnt:
@zaggino You didn't miss anything, the prefs haven't been implemented yet :)

zaggino:
Ah right, keep the work coming then, it'd be great to see this in 1.6 :)

petetnt:
Made additions and modifications according to @zaggino 's comments

- Unified hover styles for the flip view buttons
- Added preferences for showing the buttons and collapsing the pane after last file is closed
- Fixed bug where close button would close the wrong file if the panel was not active

src/view/Pane.js
```diff
@@ -161,15 +161,44 @@ define(function (require, exports, module) {
...
+     * Internal pane id
+     * @const
+     * @private
+     */
+    var SECOND_PANE         = "second-pane";
+
+    // Define showPaneHeaderButtons, which controls when to show close and flip-view buttons
+    // on the header. Possible values "hover", "always" and "never"
+    PreferencesManager.definePreference("pane.showPaneHeaderButtons", "string", "hover", {
+        description: Strings.DESCRIPTION_SHOW_PANE_HEADER_BUTTONS
```
sprintr: You seems to have forgotten to add `"hover" "always"  "never"` to values array here so it can be shown in the hints.

petetnt: Thanks @sprintr , fixed that.

abose:
@petetnt I can't see the close button when i tested this out. Do I need to do something more to enable close butoon?

petetnt:
@abose good catch, the close button is there but you cannot see it when you are using a `light` theme such as Brackets default. Didn't test it out on a lighter theme after adding the close button, will fix asap

petetnt:
@abose It's now fixed :hammer:

abose:
works now :sunglasses: 
image
When a pane is empty, the flip button seems to be confused.

petetnt:
@abose fixed that regression (nitpicky CSS inheriting :fire_engine:).

abose:
Thanks for this feature @petetnt .
All unit/integration tests passing. LGTM.
When this is ok to be merged, Please remove the [review] tag from the title.

petetnt:
Ready to be merged as far as I am concerned unless others have something else to add.

abose:
Thanks @zaggino @sprintr @le717 for reviewing.
Merging.

petetnt:
:fireworks: :tada: :+1:

zaggino:
cool stuff, nice work @petetnt ;)

arctwelve:
@petetnt Congrats on the new feature.

PR #11751, opened 2015-09-27
Allow CSS Code Hints in PHP

### LINKED ISSUE #11629: Code hint/completion regardless of extension

sprintr:
Allow CSS Code Hints in style tags for PHP files. Fixes #11629

CC: @abose

### DIFF
--- a/src/language/CSSUtils.js
+++ b/src/language/CSSUtils.js
@@ -83,23 +83,34 @@ define(function (require, exports, module) {
     
     /**
      * @private
+     * Returns state of a context
+     * @param {{editor:{CodeMirror}, pos:{ch:{string}, line:{number}}, token:{object}}} ctx
+     * @return {{tokenize:function, state:string, stateArg:string, context:Object}}
+     */
+    function _getContextState(ctx) {
+        if (!ctx || !ctx.token) {
+            return null;
+        }
+        var state = ctx.token.state.localState || ctx.token.state;
+        if (!state.context && ctx.token.state.html.localState) {
+            state = ctx.token.state.html.localState;
+        }
+        return state;
+    }
+
+    /**
+     * @private
      * Checks if the current cursor position is inside the property name context
      * @param {editor:{CodeMirror}, pos:{ch:{string}, line:{number}}, token:{object}} context
      * @return {boolean} true if the context is in property name
      */
     function _isInPropName(ctx) {
-        var state,
+        var state = _getContextState(ctx),
             lastToken;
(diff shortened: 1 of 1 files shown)


src/language/CSSUtils.js
```diff
@@ -83,19 +83,31 @@ define(function (require, exports, module) {
     
     /**
      * @private
+     * Returns state of a token
+     * @param {{editor:{CodeMirror}, pos:{ch:{string}, line:{number}}, token:{object}}} ctx
+     * @return {{tokenize:function, state:string, stateArg:string, context:Object}}
+     */
+    function _getState(ctx) {
```
ficristo: Can you rename it like _getTokenState or _getLocalState? _getState seems too common to me.
Can you add an assert for ctx?

src/language/CSSUtils.js
```diff
@@ -83,19 +83,31 @@ define(function (require, exports, module) {
...
      * @param {editor:{CodeMirror}, pos:{ch:{string}, line:{number}}, token:{object}} context
      * @return {boolean} true if the context is in property name
      */
     function _isInPropName(ctx) {
-        var state,
+        var state = _getState(ctx),
             lastToken;
-        if (!ctx || !ctx.token || !ctx.token.state || ctx.token.type === "comment") {
+        if (!state || !ctx || !ctx.token || ctx.token.type === "comment") {
             return false;
```
ficristo: Here, and twice below, you should move the check at the end of the `if` (after you are sure ctx is valid)
or simply move the assignment after the first `if`.

abose:
Hi guys, As we will be releasing 1.5 by Thursday, this change has to be pulled in by Tuesday or we could push this to 1.6.

sprintr:
@abose Cool! Will push the changes in a while.

src/language/CSSUtils.js
sprintr: `getContextState` makes more sense here.

abose:
Thanks @sprintr .
Verified, all unit tests passing. Merging :+1:

PR #11786, opened 2015-10-07
remove predefined values from cubic-bezier()

myakura:
looks like https://github.com/adobe/brackets/commit/1295cfe09162a08b1341dec71ff8e169b5526f9f#diff-0 unintentionally introduced these predef values. so taking them out.

### DIFF
--- a/src/extensions/default/CSSCodeHints/CSSProperties.json
+++ b/src/extensions/default/CSSCodeHints/CSSProperties.json
@@ -10,7 +10,7 @@
     "animation-iteration-count":   {"values": ["infinite"]},
     "animation-name":              {"values": ["none"]},
     "animation-play-state":        {"values": ["paused", "running"]},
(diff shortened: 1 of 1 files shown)


abose:
Thanks @myakura for contributing to brackets. 
The [Contributor License Agreement](http://dev.brackets.io/brackets-contributor-license-agreement.html) should be signed before we can accept the change.

nethip:
@redmunds Would you mind reviewing this change?

petetnt:
I think those values might be intentional: the values are equivalent to `ease-in-out`, which is often used* as the default value for `cubic-bezier` curves because it nicely shows what it's meant for. Other default would be equal to `linear (0,0,1,1)`, which isn't that nice IMO.

* citation needed, might just be my own opinion.

nethip:
Thanks @petetnt for quickly jumping in and trying to analyze this :+1:

redmunds:
@nethip Yes, @petetnt is correct about the original reason for using predefined values. But, I think that @MarcelGerber fixed it so that `cubic-bezier()` will now give the same result, so predefined values may no longer be necessary.

myakura:
@abose thx. signed!

abose:
Merged for 1.6. Thanks @myakura

PR #11818, opened 2015-10-13
Release

rubenandre:
I like this code editor but  I see myself limited because only the for some languages like html, css, js. For I would like to use this code editor for PHP and Ruby on rails, but when I write the keyword php or tuby on rails, does not identify me. So I have to change editor.

Is there a plugin that I can install to be able to also program the brackets in other languages?


abose:
@ruben718  This PR is about merging release brach with master, Which is usually done after the 1.5 release. Could you specify why this PR was raised?

abose:
Closing as this would be normally be done by the team after r1.5 is drafted.

PR #11834, opened 2015-10-17
Czech translation for v1.5

dvorapa:


abose:
1.6 ?

dvorapa:
I know I'm late but this is translation for v1.5 strings. For v1.6 I'll make another PR after conventional Google Groups request...

abose:
Thanks @dvorapa .  
1.6 strings could be another PR i guess, that would avoid rework and delays in string translations.
Merging :+1:

Issue #11609, opened 2015-08-20
Right-clicking and creating new files or folders doesn't resolve paths for .. and /


humphd:
It's possible to right-click in the file tree and create a file (or filder) named `../something` or just `..`, and the path isn't resolved, but rather the `..` and `/` characters are treated as literals to be used in the filename itself.

![screenshot 82](https://cloud.githubusercontent.com/assets/177243/9372501/1a83a41a-4694-11e5-806a-83b65455868a.png)

First discovered by @Pomax in https://github.com/humphd/brackets/issues/434, and now I'm filing upstream, since this is really a bug in Brackets.

tallandroid:
Works for me on mac. I created both a file and a folder with prefix to be .. in a subfolder. Both got created in the parent folder. However , the UI shows this 

![image](https://cloud.githubusercontent.com/assets/425227/9381349/522d0292-46eb-11e5-855e-1b70a71a136f.png)

Can you confirm if the directory structure is as expected in the file system. In that case , its a bug with the File Tree UI.

petetnt:
Nice find! I like how it *sort of* works how you would expect it to regarding creating files with relative paths and it also *sort of* works how you would expect it to work if they were just filenames.

Another case that brings really weird behavior is naving a file `/foo.js`. It's created as `foo.js`, but stays as `/foo.js` in the editor until refreshing.

I'd say the expected behavior would be showing a "filename cannot contain error", something similar to this Windows example:

![image](https://cloud.githubusercontent.com/assets/7641760/9386376/e942b0c8-4763-11e5-85bf-0efd0dde1601.png)

humphd:
@petetnt, I had the same thought, to show a "filename cannot contain..." error of some kind.  I don't think you want people using the filename as a way to bypass pathing.

petetnt:
@peterflynn noted on the mailing list (Brackets-Dev) that error/validation is there already, but it's not in a tooltip form but a modal error that appears when there are invalid characters. The "/" case is just failing: 

![image](https://cloud.githubusercontent.com/assets/7641760/10785129/aa291ed0-7d6a-11e5-8ec7-7b4b88307c3e.png)

tallandroid:
Submitted a pull request : https://github.com/adobe/brackets/pull/11862

As @peterflynn  pointed out , the validation is failing for this case. 

The reason being, the validation checks for the basename of the file which strips out the `/../` from the fullPath. So the validation code gets the basename which is perfectly valid. 
[...]

PR #11862, opened 2015-10-28
Fix to prevent moving files by renaming files using FileTreeView

tallandroid:
1) Added / to be an invalid file name character for mac platform
2) Return baseName to include ../ if the fullpath contains ../


petetnt:
`../` is an valid expression on Windows too. So the function could be re-factored as:

```
    function getBaseName(fullPath) {
        var lastSlash = fullPath.lastIndexOf("/");
        var preFinalSlash = fullPath.substring(0, lastSlash).lastIndexOf("/");

        if (lastSlash === fullPath.length - 1) {  // directory: exclude trailing "/" too
            return fullPath.slice(fullPath.lastIndexOf("/", fullPath.length - 2) + 1, -1);
        } else if (fullPath.substring(preFinalSlash, lastSlash + 1) === "/../") {
            return fullPath.slice(preFinalSlash + 1);
        }
        return fullPath.slice(lastSlash + 1);
    }
``´

However, it's only a partial fix for all the issues listed in #11609. It solves the main one, but the others still remain.

Some tests cases that should fail:

```
../helloworld.js
hello/world.js
/helloworld.js
helloworld.js/
../hello/../world.js
```

tallandroid:
The fix covers the cases like : 
../helloworld.js
../hello/../world.js

hello/world.js and helloworld.js/ are already covered by appshell rename validation logic. /helloworld.js is already covered.

petetnt:
@tallandroid yeah, this PR does fix the original issue that #11609 so maybe the other things I mentioned are out of the scope for this PR (albeit related to my original issue https://github.com/adobe/brackets/issues/11609#issuecomment-133038635).

Which platform are you on? At least on Windows the validation fails with `/`:s. Check something like this:

image

image

:arrow_right: folder `foo` doesn't exist, it throws an error

image

Let's create a folder 

image

and try that again

image

it actually created the other file too (like one could except if relative paths would be valid `filename`:s

image

(The foo/bar.js file opens the correct file too, so there's that).

Similar things can be done with things like creating folders by naming a file with ending slash et cetera and it can be abused to infinity. That's why I think it's important that the fix would apply to all the cases regarding `/`'s in filenames.

tallandroid:
I get your point. I think it makes more sense to weed out validation against base name completely. That is the safest path to go. I have updated the PR. I would appreciate if you can validate it on windows platform.

I verified the use cases on mac platform.

petetnt:
Works great for renaming old files on Windows :+1: 

However, creating a new file and renaming that instantly skips `_renameItem` all together so at that part exploiting `/`'s is still possible:

Check out this if-clause https://github.com/adobe/brackets/blob/d939740fd5c368c347ac6a8c7cef2941981aa2fb/src/project/ProjectModel.js#L959 and the method `createAtPath` that it uses to create the file: https://github.com/adobe/brackets/blob/d939740fd5c368c347ac6a8c7cef2941981aa2fb/src/project/ProjectModel.js#L992

tallandroid:
I intentionally left out createAtPath workflow as I believe its out of the scope of this PR. I completely understand the workflow mentioned above but I am skeptical of this being called a bug or not. Couple of reasons : 
1) The file gets created at the path intended and gets shown correctly in Working Files section.
2) FileTreeView shows it as a separate file, but on refresh it gets pushed to the right path, so I am assuming that is what is expected. What is required is refreshing the FileTreeView on create like the way we do it for rename.

I would prefer discussing that over the group once and once confirmed , I myself am willing to make that change too :)

petetnt:
Yeah, I am okay with this PR focusing on the renaming files part too :+1: 

LGTM

tallandroid:
Ping.  I am not sure abt the turnaround time.

abose:
Sorry for the late response .
@tallandroid @petetnt What is left to be done in this PR?

petetnt:
@abose I think it's ready to be merged, the other related issues can be fixed and tracked separately.

tallandroid:
@abose is this done ? Let me know if you are waiting on something.

abose:
Thanks @tallandroid .
Will merge once 1.6 is branched out to release.

tallandroid:
Ping ?

zaggino:
Reviewed and merged (as it should have been a while ago).

petetnt:
Thanks @zaggino and especially @tallandroid for your patience!

Issue #11735, opened 2015-09-24
Feature Request: Keep search bar open while editing


spongessuck:
Right now, if you click in the edit panel, the search bar closes. It would be nice if the search bar would remain open while editing, or at least for it to remember the last thing you searched for. I often search around for matches to do some minor edits and move on to the next match, but I have to re-open the search and type my pattern again.

kidwm:
and keep the highlight effects too.

fabio-farolfi:
every time I have to reopen the search and type my pattern again.
I hope for a solution as it is terribly annoying.

abose:
Search bar will now remember the last searched phrase. closing as fixed for 1.6.

phanect:
Hi @abose, 

> Search bar will now remember the last searched phrase

It doesn't help when I want to:

1. Copy phrase from opening file and paste it as phrase to find
2. Copy another phrase from opening file and paste it as phrase to replace

Phrase to find is discarded when you copy another phrase.
So I appreciate if search bar is kept open, or Brackets remember last *entered* phrase, not searched phrase.

Likewise42:
Just commenting to say that I would also like the search bar to stay open until I manually close it.

PR #11869, opened 2015-10-29
Remember last searched phrase & instant search paste fix.

abose:
fix: instant search doesn't work after ctrl+v on textfield
Remember last searched phrase #11735

### DIFF
--- a/src/search/FindBar.js
+++ b/src/search/FindBar.js
@@ -53,7 +53,8 @@ define(function (require, exports, module) {
     var lastTypedTime = 0,
         currentTime = 0,
         intervalId = 0,
-        lastQueriedText = "";
+        lastQueriedText = "",
+        lastKeyCode;
 
     /**
      * @constructor
@@ -278,6 +279,7 @@ define(function (require, exports, module) {
             })
             .on("keydown", "#find-what, #replace-with", function (e) {
                 lastTypedTime = new Date().getTime();
+                lastKeyCode = e.keyCode;
                 var executeSearchIfNeeded = function () {
                     // We only do instant search via node.
                     if (FindUtils.isNodeSearchDisabled() || FindUtils.isInstantSearchDisabled()) {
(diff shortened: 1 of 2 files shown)


petetnt:
Should the `lastQueriedText` persist over Find In Files only or during all the findbar events? (Normal CTRL+F, Search and Replace, Replace in Files...). Consider something like following:

1. Find In Files -> "Hello world"
2. Focus editor 
3. Reopen Find Bar -> it says "Hello world" (expected)
4. Delete "Hello World", replace with "Hot dogs"
5. Focus editor
6. Reopen Find Bar -> it says "Hello world" again
  -> expected: 

a) "Hot dogs" (persisting searches)
b) Nothing? (the current default behavior (without Find in Files/Instant search))

I would be partial for all searches persisting (a))

There's another variation of the previous where it's even more mixed up:

1. Repeat 1-3.
2. Replace with something that exists
3. Focus editor
4. Find the next match with F3
5. Open Find bar -> The last phrase persists
6. Focus editor
7. Open Find bar -> it's "Hello world" again

:+1: for this feature otherwise

abose:
Thanks @petetnt for reviewing. 
I have modified this to save last typed text irrecpective of success of last executed query.

petetnt:
Tested against https://github.com/adobe/brackets/commit/30716f4d765b87617cfc28cf1d0cae4106f0c3ef again and it fixes everything I mentioned in https://github.com/adobe/brackets/pull/11869#issuecomment-152453483 :+1: LGTM as it is.

Here's some food for though: should it remember the position of your search too? Consider something:

1. Make a document

```
bar
foo 1
foo 2
bar
foo 3 
foo 4
```

Search for `foo`, cycle to the third `foo` and focus the editor on the `bar` -> Find-bar closes. Reopen Find bar. Is the expected highlighted one the closest one (the current implementation) or the 3rd one, the one you were in when you focused the editor. I think the latter would make more sense *in most cases* when the last searched phrase persists too.

abose:
I prefer the current implementation, the search progressing closest to the current cursor. That might be individual preference.

rroshan1:
I have verified the changes. Looks good to me.
Point to Note:-
<b>On reopening the "Find in Files" bar, any active selection would be given preference over the last remembered search.</b> :+1: 
Merging the PR.

petetnt:
:cool:

abose:
Thanks all :)

Issue #11763, opened 2015-09-30
Dilato: Core: The new created file under the Working Files can't be renamed/showed in File Tree/Showed in Explorer with no message, if it isn't saved.


Zhanxin:
Steps:
1. Install and launch the Brackets
2. Click File menu and click New, then user would create a file under the Working Files
3. Right click on the file, click Rename/Show in File Tree/Show in Explorer

Result: There is no response after click these, if the file is not saved
Expect: There is an error message pop up to prompt user to save the file
![noresponse](https://cloud.githubusercontent.com/assets/2245179/10196484/abf4409c-67c5-11e5-8724-a8c345d41705.png)

manoharprabhu:
Hi,
I am currently working on this bug, and I am having trouble phrasing the error message that needs to be shown.
Right now, it looks like this:
![image](https://cloud.githubusercontent.com/assets/2051859/10811249/e6c3e7b0-7e2f-11e5-9c3f-273a6a4492b3.png)

Is this fine?

abose:
@manoharprabhu Thanks For contributing to brackets.
Please raise a Pull request with the fix. Community members would help you in reviewing strings used in brackets once you raised the PR :+1:

mansimarkaur:
I want to start contributing to brackets. Is this issue still open for working? If yes, please help me get started.

ficristo:
If @manoharprabhu will not work on this anymore, yes.
Personally I'd prefer to simply disable the menu items than showing a message, maybe you can check other code editor like Notepad++ or Sublime and see what they do.

manoharprabhu:
@mansimarkaur can go ahead and contribute the fix for this issue.
Sorry, I got caught up in a lot of stuff and couldn't continue contributing.

mussakhojayeva:
is this issue still valid?
i would like to work on fixing it, if it is possible

ficristo:
Maybe I closed the related PR a bit hastily...
@petetnt was it ready to be merged but lacked only tests?
Maybe we could merge as is and add some tests in a different PR.
WDYT @petetnt?

petetnt:
@ficristo sounds good to me

ficristo:
@petetnt could you double check and eventually merge?
@mussakhojayeva are you willing to write tests for this feature? (I need to wait for the merge before)

petetnt:
See the follow up in #13134, fixed in #12806

PR #11873, opened 2015-10-29
Fix for #11763

manoharprabhu:
The new created file cannot be renamed/showed in file tree/showed in
explorer with no messages if it isn't saved.

The file is checked if it is an InMemoryFile before attempting to perform
the above mentioned operations on it. If it is indeed not saved, then a
message is shown to the user to save the document to proceed. Upon
clicking 'Ok', the save dialog box opens up allowing the user to save the
file.

### DIFF
--- a/src/command/DefaultMenus.js
+++ b/src/command/DefaultMenus.js
@@ -34,7 +34,41 @@ define(function (require, exports, module) {
     var AppInit         = require("utils/AppInit"),
         Commands        = require("command/Commands"),
         Menus           = require("command/Menus"),
-        Strings         = require("strings");
+        Strings         = require("strings"),
+        MainViewManager = require("view/MainViewManager"),
+        CommandManager  = require("command/CommandManager");
+    
+    /**
+     * @private
+     * Enable or disable the supplied menu items in context menu.
+     *
+     * @param {boolean} enabled Indicates whether the items are enabled or disabled.
+     * @param {Array} Array of command IDs of the menu items on which the action is to be performed.
+     */
+    function _setContextMenuItemsVisibility(enabled, items) {
+        items.forEach(function (item) {
+            CommandManager.get(item).setEnabled(enabled);
+        });
+    }
+    
+    /**
+     * @private
+     * Compute the visibility of context menu items when working on files that have not yet been saved.
+     */
+    function _setMenuItemsVisibility() {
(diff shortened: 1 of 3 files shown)


abose:
tagging @petetnt

petetnt:
The fix itself LGTM :+1: Not sure if `Rename file` should focus on the file after saving (it's a bit redundant so I think not). 

The phrasing of the error message is a bit harder though like @manoharprabhu you mention in https://github.com/adobe/brackets/issues/11763#issuecomment-152085268.

I might go for something like

> **The file needs to be saved before continuing**
> You'll be prompted to save this file.
> [ OK ]

Could you add an unit test for this @manoharprabhu? Otherwise nice work :100:

busykai:
Wouldn't a better solution be disabling the context menu item if the file is not yet saved? The rename action should not be permitted in the first place. (Instead of permitting it and then showing an error -- it's disruptive.)

petetnt:
@busykai I was thinking about that too (especially with the `rename` in mind). When I was testing this out (I didn't even know that you could create a new file without creating it to the tree before this issue!) I noticed  that it's not exactly obvious what separated *that* file from the others. So in that sense prompting for save did make sense for in-memory files and the disabled options *could* be confusing.  

Then again, those options don't make much sense for files that aren't saved, so in that regard you are absolutely correct @busykai

manoharprabhu:
That sounds reasonable. So is it fine to disable those 3 items in context menu for unsaved files?

abose:
+1 for disabling

manoharprabhu:
Do I have to implement this as an extension?
I can see a extensions like 'CloseOthers' manipulating the context menu.
I found out that I have to listen to the 'beforeContextMenuOpen' event on the WORKING_SET_CONTEXT_MENU and PROJECT_MENU to disable the three items if the file has not been saved yet.

abose:
As this is one of the core workflows, i am inclined to do the change directly in core than in an extension.

manoharprabhu:
@abose , I have registered the event handlers in DefaultMenus.js file. I also had to import the MainViewManager and CommandManager modules to make the change work.
Is there a better place to register the event handlers for disabling the menu items?

abose:
DefaultMenus.js looks fine to make the change.

manoharprabhu:
Hey @abose . I was quite busy over the past couple of weeks. The latest pull request contains required the changes to disable the context menu items for  unsaved files.
I am getting around to writing unit tests for these.

PR #11885, opened 2015-11-01
Add UrlCodeHints for poster-attribute

### LINKED ISSUE #11884: File selector dropdown doesn't appear when linking a "poster" attribute.
The Video tag uses a poster attribute that allows you to link a picture.  There is no link drop-down that will let you easily navigate your project folders like it would appear with other attributes like `src` or `href`.

petetnt:
Fixes #11884


abose:
Thanks @petetnt 
Unit tests passing.
Merging.

PR #11894, opened 2015-11-05
Ensure that .cm-error gets applied last

### LINKED ISSUE #11890: HTML tag highlighting is sometimes misleading
If you have an HTML tag `<div>` closed by `</div?>` or a `</div*>` where * denotes any special character (anything other than an alpha-numeric) , the opening and closing tags are both highlighted which is misleading. So the parser assumes that the `<div>` is closed by `</div?>` even though this is not true and this is hard to spot when debugging.

petetnt:
This simple PR ensures that `.cm-error` gets applied last. Fixes #11890

Before:

image

After

image


abose:
:cool:  merging. Thanks @petetnt  .

petetnt:
:fireworks:

PR #11942, opened 2015-11-20
HTML Special character update

GBratsos:
Updated HTML special character >.


valtlait:
In HTML `>` is not special character although `<` does.

ficristo:
Hi @GBratsos thank you for your contribution.
You need to submit the contributor license agreement. See http://dev.brackets.io/brackets-contributor-license-agreement.html
Did the `>` caused you any trouble?
There is a mixed usage of `>` and `&gt;` on the samples so if you really think this is worthwhile I'll merge this.

ficristo:
Since we didn't hear back from I close this.

PR #11949, opened 2015-11-23
Type Inference in hint list.

swmitra:
Show type data in the hint list. This is work in progress as we need more ideas about the type symbols which are just text in this PR.
types

A snapshot of the rendered types from this implementation.

### DIFF
--- a/src/extensions/default/JavaScriptCodeHints/styles/brackets-js-hints.css
+++ b/src/extensions/default/JavaScriptCodeHints/styles/brackets-js-hints.css
@@ -22,14 +22,44 @@
  */
 
 .brackets-js-hints-type {
-    width: 30px;
-    color: cornflowerblue;
+    font-size: 10px;
+    width:25px;
+    color: white;
     font-weight: 900;
-    padding-left: 5px;
     text-align: center;
     float: left;
-    box-sizing: border-box;
+    box-sizing: border-box;  
+    font-style:normal;
+    padding:0px 0px 0px 6px;
+    text-shadow:
+    -1px -1px 0 #8d8d8d,
+    1px -1px 0 #8d8d8d,
+    -1px 1px 0 #8d8d8d,
+    1px 1px 0 #8d8d8d; 
 }
+
+.highlight .brackets-js-hints-type {
+    font-size: 13px;
+    text-shadow:
+    -1px -1px 0 #6495ed,
+    1px -1px 0 #6495ed,
+    -1px 1px 0 #6495ed,
+    1px 1px 0 #6495ed; 
+}
+
+.brackets-js-hints-type-details {
+    color: #d3d3d3;
+    font-weight: 100;
+    font-style: italic!important;
+    padding-left: 10px;
+    float:right;
+    margin-left:5px;
+} 
+
+.highlight .brackets-js-hints-type-details {
+    color: #0c50af;
+}
+
 
 .brackets-js-hints.priority-high {
     color: #486c00; /* green */
--- a/src/extensions/default/JavaScriptCodeHints/styles/brackets-js-hints.css
+++ b/src/extensions/default/JavaScriptCodeHints/styles/brackets-js-hints.css
@@ -159,7 +159,7 @@ span.brackets-js-hints {
 }
 
 .dark .jshint-description {
-    color: #696969;
+    color: #ccc;
 }
 
 .jshint-jsdoc {
@@ -175,7 +175,10 @@ span.brackets-js-hints {
     clear: left;
     max-height: 3em;
     overflow: hidden;
-    opacity:0.7;
+}
+
+.dark .jshint-jsdoc {
+    color: #ccc;
 }
 
 .jshint-link {
--- a/src/extensions/default/JavaScriptCodeHints/styles/brackets-js-hints.css
+++ b/src/extensions/default/JavaScriptCodeHints/styles/brackets-js-hints.css
@@ -38,6 +38,14 @@
     1px 1px 0 #8d8d8d; 
 }
 
+.dark .brackets-js-hints-type {
+    text-shadow:
+    -1px -1px 0 #696969,
+    1px -1px 0 #696969,
+    -1px 1px 0 #696969,
+    1px 1px 0 #696969; 
+}
+
 .highlight .brackets-js-hints-type {
     font-size: 13px;
     text-shadow:
@@ -51,13 +59,20 @@
     color: #d3d3d3;
     font-weight: 100;
     font-style: italic!important;
-    padding-left: 10px;
     float:right;
-    margin-left:5px;
+    padding-left: 10px;
+}
+
+.dark .brackets-js-hints-type-details {
+    color: #696969;
 } 
 
 .highlight .brackets-js-hints-type-details {
-    color: #0c50af;
+    color: #6495ed;
+}
+
+.brackets-js-hints-type-details.keyword {
+    color: #6495ed;
 }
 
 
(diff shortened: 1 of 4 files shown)


swmitra:
Tagging @nethip @abose @ryanstewart

sprintr:
I believe it would be nice to show the data types as the output of `typeof(instance)` such as `"string"`, `"number"`, `"function".

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures

swmitra:
Thanks @sprintr. I will try that once and post a snapshot of the rendering. Our goal is to have the type data without being too loud. We have to compare different displays in order to conclude.

swmitra:
@sprintr With some CSS changes and new literals , this is how it looks. 

types1

I must confess , to me it looks cluttered.

petetnt:
@swmitra maybe aligning the literal labels to right would help (plus maybe more padding to the right side). I think that the labels the OP look pretty nice too, but then again the literals are more clearer (in a way). Maybe the labels could come from a `labelProvider` for a customization option :beers: 

What about mixed types? For example `[0, "string", {object: "foo"}]`

abose:
+1 for right align.
+1 for short symbols as in the initial implementaion. Maybe a hybrid- on hover/arrow key on current selected entity, we could expand the short symbol to detailed string?

ficristo:
FWIW, I like the second screenshot with right alignment.
I looked a bit in VisualStudioCode which has the inference type too: I like it would inference the return type of a function too.

swmitra:
@petetnt @ficristo @abose @sprintr  A snap with right aligned symbols and more padding.
types2

swmitra:
@petetnt I will write a label provider which can be overridden by extensions. 
autocomplete plus works in the same way I guess. 
Also I have to infer a mixed type array , will update the PR with the check.

ficristo:
@swmitra Can you try to put the type on the right of the variable names?

abose:
:sweat_smile: I too meant right align after the variable names!

sprintr:
This looks good. I am not sure if short names will look good in right alignment.
+1 for right alignment.

swmitra:
@sprintr @petetnt @ficristo @nethip Was trying with different styles. This one
types3
 stands out with a small css tweak.

swmitra:
@larz0 @ryanstewart Please provide your inputs.

swmitra:
Some more tuning with the type details ( font-style and weight change ) 
types4

ficristo:
Just throwing a couple of ideas:
- the icon and the type should gray and only the one selected ones should be blue so the focus is on the autocomplete (or simply found some way to make the autocomplete a bit more outstanding)
- the icons seem ot use another style but they are interesting, maybe make them use less space
- I don't like much the italic version but the word `keyword` look better there. In the previous screen it has a too much different style

swmitra:
Thanks @ficristo . Something like this - 
types5

ficristo:
As usual my english need some improvements...
I like the smaller icons but I meant that the left padding should equal to the right: so reduce the right padding. Then the icon will be centered and it will use less space.
The keywords should remain in blue so they stand out more.

And then I stop here.
@swmitra Thank you for your the effort.

sprintr:
I think the function argument list can be omitted from the code hints (see `var3` in the above screenshot) since brackets already supports it. This will make code hints small and tidy.

2015-11-24 19_40_48- main js html - brackets

swmitra:
Thanks a lot @sprintr for your prompt feedback. I agree that it doesn't exactly look tidy , but from a developer point of view , I think this adds more flexibility even before I choose the guess. In it's current form, the hinting module allows the parameters hinting only after choosing the guess from the list. But in the approach above we get to see  the param type information upfront. 

I know it's debatable , but that's why communities are best. In a day the entire ux got changed and became considerably more informative.

swmitra:
Snapshots under default dark theme and light theme. 
types6dark
types6

sprintr:
@swmitra Maybe we could take a leaf out of pref code hints where lengthy descriptions are shown under the hint. The type could come on the right and the function signature could come under the hint.

swmitra:
@sprintr I was actually planning to use that for available docs for the particular definition when highlighted.

abose:
@swmitra Docs in hints is brilliant! 
Don't know about the space utilization though, it would mean that other variables might be pushed down which would call for a not so helpful code hints. a mock Screenshot would help.

Denisov21:
:+1: for short symbols as in the initial implementaion
Wonderful project!

abose:
Tagging @larz0 for ui review.

swmitra:
Uploading new screenshots for brackets light and dark theme.
latest_light
latest_dark

'<>' in Red indicates an external link for more details. Fetched from JS doc.

swmitra:
@sprintr @petetnt @ryanstewart @abose @ficristo Feedback Please :speech_balloon:

src/extensions/default/JavaScriptCodeHints/styles/brackets-js-hints.css
```diff
@@ -21,6 +21,200 @@
...
+
+.dark .jshint-description {
+    color: #696969;
+}
+
+.jshint-jsdoc {
+    display: none;
+    padding-left: 28px !important;
+    padding-right: 10px !important;
+    color: grey;
```
sprintr: It doesn't look very visible. We use `#ccc` in pref hints so it will be good to use it here as well.

sprintr:
Hints with lengthier signature are breaking the layout of individual hints.

2015-12-09 16_00_14- main js html - brackets
2015-12-09 16_08_51- main js html - brackets

swmitra:
@sprintr It is intentional to make the type details appear in next line if it is beyond a certain character length.

abose:
It will look better if the signature is also right aligned in the second line.

sprintr:
Maybe we could remove the 5px padding on the signature then

sprintr:
```js
// 1
function Get() {
    var h1 = document.
}
```
```js
// 2
function Get() {
    var h1 = document.crea
}
```

Brackets now throws an error and doesn't give any code hints for script 1 (from dot up to 3 chars) whereas script 2 works just fine.

swmitra:
@sprintr  Fixed now. jQuery was complaining as it was thinking that js doc spans innerHtml was malformed ( as it contains html tags sometimes ). Fixed that by adding the doc as text instead of html.

sprintr:
@swmitra Great. works fine now.

swmitra:
@sprintr Replaced '<>' with a boxed '?'. Looked better to me. Please have a look and let me know.

sprintr:
@swmitra Looks better to me too. `<>` had an html like feeling to it.

abose:
Thanks @swmitra +  @petetnt @ficristo @sprintr @Denisov21 For this feature :fireworks: 
All unit tests/ extension tests for code hints passing.
Tagging @larz0 for UI Review.
Please log issues if any found for the feature.

Merging to give better testing window before shipping.

nethip:
@swmitra Great work! This is awesome.

Small recommendation. I think the hints list looks cleaner without the box icons on the left. Also we could do something better with replacing current icon "?". Let us go without the box icons for now. We could then incorporate the changes(or icons) once we hear from @larz0 .

ex1st:
[Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) and [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) have been forgotten.

PR #11957, opened 2015-11-25
fix hardcoded port in MultiBrowser LiveDevelopment and minor fix in PerfUtil

konradlipner:
If performance data gathering is disabled, markStart will return undefined instead of id. Calls like updateMeasurement and finalizeMeasurement will fail on id.id.


kolipka:
Travis says:
Running "cla-check-pull" task
Unexpected response from api.github.com
statusCode: 403
API rate limit exceeded for 52.0.240.122. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.

abose:
We have a limit on the number of Travis builds that can go through before exceeding the Github API rate limit. It seems that dev activity was high during the time, and the rate limit exceeded. We usually retrigger the build after some time. So no worries in the case.

src/LiveDevelopment/MultiBrowserImpl/transports/node/NodeSocketTransportDomain.js
```diff
@@ -58,7 +58,7 @@
     var _clients = {};
     
     // This must match the port declared in NodeSocketTransport.js.
-    // TODO: randomize this?
+    // Default socket port number
     var SOCKET_PORT = 8123;
```
ingorichter: Since there is a pref for the socket port, this should be read from the prefs too.

kolipka: Are you sure? This is part of NodeDomain, can I read prefs there? Anyway: starting server sends port number, this here will be overwritten.

PR #11909, opened 2015-11-12
Proper word wrap intents. fix for #11633

abose:
fixes https://github.com/adobe/brackets/issues/11633
@petetnt @swmitra @wesleifreitas

PR #11964, opened 2015-12-01
Port missing piece of the indent-wrap hack from Codemirror. Fixes #11963

### LINKED ISSUE #11963: Tab indentations aren't rendered
I'm using the latest and greatest master (and the same happened for a build from 2 weeks ago), that documents with tabs used for indentation are all rendered flat against the left edge of the document.

Upon inspecting, it appears that the inline `text-indent` and `padding-left` styles are the cause. If you add an overriding style of: 

```css
.CodeMirror-line {
	padding-left: 0 !important;
	text-indent: 0 !important;
}
```

then the indentation works as expected. Files which use spaces for indentation are unaffected by default nor by the override.

I have tried with and without extensions; the same thing happens.

So, what's the point of the crazy text  `text-indent`/`padding-left` inline-styles, and can they safely be removed completely? Because clearly, they break things now.

petetnt:
In #11909 we added a `lineRender` hack from https://codemirror.net/demo/indentwrap.html. This hack alone works fine for files using spaces as whitespace. However, the PR was missing this additional hack from the pages stylesheet:

```
      .CodeMirror pre > * { text-indent: 0px; }
```

This PR adds that, thus fixing  #11963

@abose @stowball @rroshan1 

edit: Steps to repro:
1. Create new file foo.html
2. Set indentation to Tab size: 4 from Brackets
3. Write `<h1>Foo</h1>`
4. Try to indent it with by pressing tab
   -> Text doesn't move at all

After PR:
1. Create new file foo.html
2. Set indentation to Tab size: 4 from Brackets
3. Write `<h1>Foo</h1>`
4. Try to indent it with by pressing tab
   -> Tabs work as indented and the `indent-wrap` hack still works too

### DIFF
--- a/src/styles/brackets_codemirror_override.less
+++ b/src/styles/brackets_codemirror_override.less
@@ -24,6 +24,10 @@
 
 .CodeMirror pre {
     padding: 0 @code-padding 0 0;
+
+    > * {
+        text-indent: 0px;
+    }
 }
 
 .show-line-padding .CodeMirror pre {


abose:
I tried the above steps, but I couldn't repro the issue in my machine. Tabs work fine with the above steps. Am I missing some steps?

petetnt:
@abose Hmm, there shouldn't be any other steps: just make sure that your Brackets indentation mode is set to `tabs`. The width of `.cm-tab` is `0px` so,  negative `text-indent` eats away the `padding-left`, resulting in zero indentation. 

Unless there are some difference between platforms how the tabs are handled. I am running on Windows 8.1. at the moment.

abose:
`"useTabChar"` preference was not set. Now I can repro this.
Verified that works. Merging.
Thanks @petetnt

petetnt:
:cool:

Issue #11984, opened 2015-12-08
Switching to ESLint


petetnt:
Branched/continued from #11644, #11693

This issue was created to track down moving to ESLint separately from the existing PR (and the general ES6 support)
- [x] Update Brackets core code to ESLint (done in #11693)
- [x] Integrate ESLint to Brackets core @zaggino (https://github.com/adobe/brackets/pull/11693#issuecomment-162841573)
- [ ] Decide what to do with the existing linter (JSLint)

The integration part should be easy but what about the third box? Possible options include:
- Make ESLint the preferred linter and let the user use JSLint if the user decides to do so (confusing to those who have just used JSLint in the past)
- Make ESLint the preferred linter and remove JSLint altogether (spinning it to non-core extension) (similar issues as above)
- Just let them exist side by side and the let the user configure the settings (most confusing as files get linted twice?)
- Provide UI to select the preferred linter on first launch of Brackets 1.x (most user friendly, needs UI work etc.)

@abose @ficristo @MarcelGerber

ficristo:
I've a biased opinion since I don't like particularly JSLint, but if we want it still in core it should be at the least updated.

abose:
Modified 3. We could default to ESLint only and provide a preference to enable JSLint also.

zaggino:
Submitted a PR https://github.com/adobe/brackets/pull/11988

tclzcja:
Since the brackets-shell ticket has been merged, does that mean ESLint will be formally included in Brackets 1.8 release? Thank you!

ficristo:
@tclzcja no, that ticket is only about to use ESLint to lint brackets-shell code.
You will have still to use an extension for 1.8 for your code.

petetnt:
What's the status of this?

zaggino:
I can setup a PR for "Integrate ESLint to Brackets core" if desired.

petetnt:
Shipped in 1.11.

PR #11988, opened 2015-12-09
add ESLint as a default Brackets extension

zaggino:
ref: https://github.com/adobe/brackets/issues/11984

mostly cut out from my extension https://github.com/zaggino/brackets-eslint


zaggino:
@petetnt @ficristo @abose I'll need help here with testing, right now I'm using this branch as my primary and everything seems to work fine for me but that might not be the case for all types of projects.

package.json
```diff
@@ -12,6 +12,10 @@
         "branch": "",
         "SHA": ""
     },
+    "dependencies": {
```
abose: Instead of doing this as part of the build process, can the node module be directly packaged or submoduled to?
Right now, Cloning the git repo makes a working copy of brackets. This will mean that we have to do a npm install for brackets to work right?

zaggino: Yeah, we could submodule it. Shouldn't be a problem.

zaggino: Wait, it might be a problem because ESLint has a lot of dependencies which are not present in its repository.

abose: We would anyway pull all the dependencies when building, so using `prepackaged node modules` or `submoduled with dependencies` in source would have the same effect. Maybe we could trim all the dev dependencies from eslint?

zaggino: Yeah, I'm not exactly sure what you mean here...

petetnt:
I've been using this branch today and most things seem to work as expected. However, I am using `eslint-plugin-react` and when I launch Brackets, I get the following error:

```
Provider ESLint (async) failed: Error: Cannot find module 'eslint-plugin-react'
```

Despite this error, `eslint` and `eslint-plugin-react` work as expected after dismissing the error (by, for example, saving a `.jsx` file)

zaggino:
@petetnt https://github.com/adobe/brackets/commit/634160c0efcd1ba2b07a67034836f0a1b9545c9c should fix that issue.

package.json
zaggino: `ESLint` has a lot of runtime dependencies like `espree` or `escope` (and more). Some of those dependencies might be written in ES6 and compiled to ES5 only when published to npm - that's why submoduling doesn't make much sense here. Only thing that comes to my mind is creating a repo for installed ESLint and submoduling that, but this is quite weird. What's the problem with npm?

abose: I also meant NPM, But for eslint, could remove dev/testing dependencies from package.json if the install size is too big.
It is possible to put the node_modules folder in the extension directory right?

zaggino: Yes, it's possible to move the node_modules to the extension directory, do you want it to be committed there?

abose: Since we would have to ship the node modules anyway, it would be ok to put it in the extension directory. Also will avoid the NPM setups needed otherwise.

petetnt:
https://github.com/adobe/brackets/commit/634160c0efcd1ba2b07a67034836f0a1b9545c9c fixed it, thanks!

src/extensions/default/ESLint/domain.js
```diff
@@ -0,0 +1,132 @@
...
+  var noop = function () {};
+  var globalPackagesAvailable = false;
+
+  function _setProjectRoot(projectRoot) {
+    var opts = { useEslintrc: true };
+    var configPath;
+    var rulesDirPath;
+    var ignorePath;
+
+    if (projectRoot) {
```
petetnt: This method could be simplified a bit by using [fs.accessSync](https://nodejs.org/api/fs.html#fs_fs_accesssync_path_mode):

``` js
    if (projectRoot) {
        configPath = projectRoot + '.eslintrc';
        fs.accessSync(configPath, function (err) {
            if (err || !fs.statSync(configPath).isFile()) {
                opts.rules = require('eslint/conf/eslint.json').rules;
            }
        });

        rulesDirPath = projectRoot + '.eslintrules';
        fs.accessSync(rulesDirPath, function (err) {
            if (!err && fs.statSync(rulesDirPath).isDirectory()) {
                opts.rulePaths = [rulesDirPath];
            }
        });

        ignorePath = projectRoot + '.eslintignore';
        fs.accessSync(ignorePath, function (err) {
            if (!err && fs.statSync(ignorePath).isFile()) {
                opts.ignore = true;
                opts.ignorePath = ignorePath;
            }
        });
    }
```

This way there's no need for `noop` either. The `fs.accessSync` can be passed an optional mode value too if one just wants to check if the file is readable

```
fs.accessSync("foo", fs.F_OK, function (err) {/*...*/});
```

zaggino: node in brackets is 0.10, no `fs.access` there: https://nodejs.org/docs/latest-v0.10.x/api/fs.html

zaggino: btw you're using `accessSync` and a callback function? weird pattern

petetnt: @zaggino Figures, I remembered that Brackets was running on 0.12. My bad :hamster:

petetnt: @zaggino hah, wow, I wrote the thing in as async with a promise, noticed that it probably should be sync instead and then I don't know what happened. Waaaay too little coffee for morning coding. :coffee:

ficristo:
ESLint supports multiple ways to define en `.eslintrc` file:
http://eslint.org/docs/user-guide/configuring#configuration-file-formats
Does this PR supports all of them?

Style nit: I guess it should indented by 4 spaces.

zaggino:
@ficristo made a fix in https://github.com/adobe/brackets/commit/7917a2d41f71e76abd6b787f571a6ed945624339 to consider all config file names

@ficristo fixed indentation to 4 spaces in https://github.com/adobe/brackets/commit/d9faa73d5f24ae268da8258e9758b6d4173cf010

@abose against my better judgement, committed the actual ESLint into the extension folder so `npm install` is not required after cloning the repository ...

let me guys know if you have any more issues with this

zaggino:
btw let me say again, we should really use `npm install` and not commit this to git, I get this error when switching branches on windows:
```
$ git checkout zaggino/eslint
fatal: cannot create directory at 'src/extensions/default/ESLint/node_modules/eslint/node_modules/handlebars/node_modules/uglify-js/node_modules/yargs/node_modules/cliui/node_modules/center-align/node_modules/align-text/node_modules/kind-of/node_modules/is-buffer/test': Filename too long
```

zaggino:
had to remove ESLint's code because of the issue above with long file names ... we should just use `npm install`

zaggino:
added this with npm3, no longer have the long file path problem ... `node_modules` occupy about 10MB of disk space this way ... that's 10MB more when cloning a repo which could be avoided by using npm

ficristo:
As shown by travis grunt eslint doesn't pass.
If I click on an error the cursor goes on the right row but on the wrong colum. (Should go on the previous one)

zaggino:
grunt eslint doesn't pass because this pr turns on indentation checks for 4 characters and there're files outside of this PR not respecting this

abose:
But the NPM install would also pull the same amount of code right. So if brackets have to be working, the node modules anyways need to be in place.
On second thought, wouldn't this add 10+mb to the installer size?

zaggino:
Yes, that's true. Installer size would definitely grow.

ficristo:
I don't see the point of the `noop` function, can you remove it?
Just leaving the comments `// no action required` should be enough.

It shouldn't matter much but maybe use `regex.test` instead of `entry.match(/^\.eslintrc(\.(js|yaml|yml|json))?$/);` since you aren't using the results.

It is safe this `process.chdir(projectRoot);` ? I assume it is since running in a node domain.

zaggino:
`process.chdir(projectRoot);` has been in the extension for months, never caused any problems ... so I assume it's safe, new version of eslint will support cwd parameter (https://github.com/eslint/eslint/commit/eb1ef887f382ae4e90b9a08b7d069507b3c18d32) so this won't be required.

ficristo:
I don't know if happened with JSLint too, but if I create a new project without a `.eslintrc` and then I add one I need to restart Brackets to make it take effect.

zaggino:
@ficristo fixed in https://github.com/adobe/brackets/commit/2b57d05bfcb500cfd06392e7b61a925547c71d08

zaggino:
This died due to lack of feedback.
Latest version currently available here https://github.com/zaggino/brackets-eslint

PR #11992, opened 2015-12-11
Adding preference to enable/disable type details in js hinting

swmitra:
- Adding preference **jscodehints.typedetails** to enable/disable type details in js hinting. Default value is true. 
  . 
- Styling changes to make the addition UI more subtle ( snapshots attached for light and dark theme for reference )

latest_light
latest_dark


abose:
Looks Better now :+1: 
Merging.
Thanks @swmitra

PR #12002, opened 2015-12-14
Reintroduce JSLint as a prefered linter

ficristo:
and disable temporary the `no-trailing-spaces` and `eol-last` eslint rules
This is based on the discussion in: https://github.com/adobe/brackets/pull/11998

Actually the ESLint checks are ATM weaker than JSLint so there shouldn't be any problems.


abose:
Thanks @ficristo .

PR #12022, opened 2015-12-23
ESLint: enabled some rules

ficristo:
Waiting for https://github.com/adobe/brackets/pull/11988.

Rules enabled:
- no-bitwise
- new-cap
- no-empty
- strict

Now the node env is specified only in the node scripts instead of globally.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,15 +1,12 @@
 {
-    "env": {
-        "node": true
-    },
     "rules": {
-        "no-bitwise": 0,
+        "no-bitwise": 2,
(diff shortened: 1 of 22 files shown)


zaggino:
This now conflicts with `master` @ficristo

ficristo:
The `strict` rule was enabled in https://github.com/adobe/brackets/pull/12718
I enabled the `no-bitwise` rule in https://github.com/adobe/brackets/pull/12719
For the `new-cap` and `no-empty` rules I didn't fint an equivalent in JSLint, so I close this.

PR #12024, opened 2015-12-24
jasmine.sh: add some log to say which platforms are support

ficristo:
I have some problem to run all the test suite from Brackets so I tried to run it from the command line only to found that OSX is the only platform supported.
This PR simply add some log to say it.


abose:
LGTM Merging.
Thanks @ficristo

PR #12025, opened 2015-12-24
Fix for relaunch corruption after activating no distraction mode

swmitra:
This PR is a fix for "first-pane" visibility corruption seen on relaunch after activating no-distractions mode.


swmitra:
Tagging @nethip .

nethip:
Thanks @swmitra

PR #12035, opened 2015-12-30
Update React to 0.14.7 and Immutable to 3.7.6.

petetnt:
This PR updates React to the newest version of `0.14.5` and Immutable to go with it.

Updating React at this point is pretty crucial (IMO) if React is continued to be used: in 0.14 there were a lot of breaking changes that are still relatively easy to fix: biggest thing was how DOM-elements are being handled in React. For example: DOM operations like `render` are now being handled by `ReactDOM`-module, `getDOMNode` has been deprecated for the components, `this.refs.componentref` returns the DOM node directly instead of the component and so on.

This PR also updates the tests involved: it fixes instances where TestUtils were used to find non-component (plain HTML) DOM Nodes and it also fixes all instances where `props` were being accessed instead of the DOM Nodes itself, which caused the following warnings: `Warning: ReactDOMComponent: Do not access .props of a DOM node; instead, recreate the props as`render`did originally or read the DOM properties/attributes directly from this node (e.g., this.refs.box.className).`

The PR has been done in backwards compatible fashion to the extent it is possible: extensions that are using deprecated methods and features should update themself before React on Brackets is updated to 0.15 some time in the future.

### DIFF
--- a/src/project/FileTreeView.js
+++ b/src/project/FileTreeView.js
@@ -147,9 +147,9 @@ define(function (require, exports, module) {
             if (e.keyCode !== KeyEvent.DOM_VK_LEFT &&
                     e.keyCode !== KeyEvent.DOM_VK_RIGHT) {
                 // update the width of the input field
-                var domNode = this.refs.name,
-                    newWidth = _measureText(domNode.value);
-                $(domNode).width(newWidth);
+                var node = this.refs.name,
+                    newWidth = _measureText(node.value);
+                $(node).width(newWidth);
             }
         },
 
@@ -184,7 +184,7 @@ define(function (require, exports, module) {
 
             var node = this.refs.name;
             node.setSelectionRange(0, _getName(fullname, extension).length);
-            ViewUtils.scrollElementIntoView($("#project-files-container"), $(ReactDOM.findDOMNode(node)), true);
+            ViewUtils.scrollElementIntoView($("#project-files-container"), $(node), true);
         },
 
         render: function () {
@@ -557,7 +557,7 @@ define(function (require, exports, module) {
 
             var node = this.refs.name;
             node.setSelectionRange(0, fullname.length);
-            ViewUtils.scrollElementIntoView($("#project-files-container"), $(ReactDOM.findDOMNode(node)), true);
+            ViewUtils.scrollElementIntoView($("#project-files-container"), $(node), true);
         },
 
         render: function () {
--- a/src/project/FileTreeView.js
+++ b/src/project/FileTreeView.js
@@ -33,6 +33,7 @@ define(function (require, exports, module) {
     "use strict";
 
     var React             = require("thirdparty/react"),
+        ReactDOM          = require("thirdparty/react-dom"),
         Classnames        = require("thirdparty/classnames"),
         Immutable         = require("thirdparty/immutable"),
         _                 = require("thirdparty/lodash"),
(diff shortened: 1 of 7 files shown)


MiguelCastillo:
Great to upgrade. Always a lil scary due to tainted historical experience when the tree was first implemented. I will pull down this branch and help you test it.

petetnt:
@MiguelCastillo Yup. Like I said, the this particular update is quite important as 0.13 -> 0.14 did contain tons of breaking changes. However the changes themselves are rather trivial to fix.

That said this PR is (again) a bit outdated as React is already on 0.14.7. But as this PR fixes the critical things for an update path, upgrading to 0.14.x or 0.15 should be trivial. I'll update this PR later when I got some spare time.

MiguelCastillo:
I am about to pull down your branch.  It might be good to update to 0.14.7. :)

petetnt:
@MiguelCastillo Bumped React to 0.14.7, Immutable to 3.7.6. All tests should still pass

src/project/FileTreeView.js
```diff
@@ -181,9 +182,9 @@ define(function (require, exports, module) {
             var fullname = this.props.name,
                 extension = LanguageManager.getCompoundFileExtension(fullname);
 
-            var node = this.refs.name.getDOMNode();
+            var node = this.refs.name;
             node.setSelectionRange(0, _getName(fullname, extension).length);
-            ViewUtils.scrollElementIntoView($("#project-files-container"), $(node), true);
+            ViewUtils.scrollElementIntoView($("#project-files-container"), $(ReactDOM.findDOMNode(node)), true);
```
MiguelCastillo: New behavior in 0.14 is that `this.refs.name` is the actual dom node.  So `ReactDOM.findDOMNode` is redundant and no longer needed.  Yay :)

src/project/FileTreeView.js
```diff
@@ -141,12 +142,12 @@ define(function (require, exports, module) {
...
          */
         handleInput: function (e) {
-            this.props.actions.setRenameValue(this.refs.name.getDOMNode().value.trim());
+            this.props.actions.setRenameValue(this.refs.name.value.trim());
 
             if (e.keyCode !== KeyEvent.DOM_VK_LEFT &&
                     e.keyCode !== KeyEvent.DOM_VK_RIGHT) {
                 // update the width of the input field
-                var domNode = this.refs.name.getDOMNode(),
+                var domNode = this.refs.name,
```
MiguelCastillo: This is old code... But do you think we can rename `domNode` to `node` to be consistent with the rest of the usage of `this.ref.XXX`?

MiguelCastillo:
The code is running ok right now.  I will continue to run it for the rest of the day to see if I can catch any odd behavior.  If all goes well, we can merge soon.

MiguelCastillo:
In case there is an interest in changes between .14 and .13.  https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html

src/project/FileTreeView.js
petetnt: Done

src/project/FileTreeView.js
petetnt: Good catch: this was a brainfart due to it needing being needed @ L365 on `componentDidUpdate` because it targets `this`. Fixed.

MiguelCastillo:
PR is looking beautiful.  Thanks for updating!!

MiguelCastillo:
Looks good to me. Merging.

petetnt:
:+1: neat. Thanks Miguel.

PR #12041, opened 2016-01-02
Use the global RequireJS config for ExtensionLoader. Fixes #1087

### LINKED ISSUE #1087: Global RequireJS config paths should be passed to extensions
To support RequireJS plugins in extensions, the path must be configured for each extension's require context.

Currently we must hardcode the path to a RequireJS extension in 2 places (1) global brackets require and (2) ExtensionLoader require context creation.

Ideally there would be a way to query the global require context and pass the config options to extensions.

petetnt:
Brackets has a long running (minor) issue #1087, where the global RequireJS path configuration is not being re-used but instead repeated in `ExtensionLoader.js`. 

This PR fixes the issue by creating a simple wrapper API around the semi-private (see [James Burkes comment here](https://groups.google.com/forum/#!topic/requirejs/Hf-qNmM0ceI)) `requirejs.s.contexts._.config` and later converts them to absolute paths in `ExtensionLoader.js`.

This same method can also be used to retrieve global modules with other modules that aren't run in the same scope as the rest of Brackets, such as WebWorkers and Node instances. For more information see my comment at: https://github.com/adobe/brackets/issues/12006#issuecomment-166005753

### DIFF
--- a/src/utils/ExtensionLoader.js
+++ b/src/utils/ExtensionLoader.js
@@ -42,7 +42,7 @@ define(function (require, exports, module) {
         Async          = require("utils/Async"),
         ExtensionUtils = require("utils/ExtensionUtils"),
         UrlParams      = require("utils/UrlParams").UrlParams,
-        PathUtils              = require("thirdparty/path-utils/path-utils");
+        PathUtils      = require("thirdparty/path-utils/path-utils");
 
     // default async initExtension timeout
     var INIT_EXTENSION_TIMEOUT = 10000;
--- a/src/utils/ExtensionLoader.js
+++ b/src/utils/ExtensionLoader.js
@@ -41,7 +41,8 @@ define(function (require, exports, module) {
         FileUtils      = require("file/FileUtils"),
         Async          = require("utils/Async"),
         ExtensionUtils = require("utils/ExtensionUtils"),
-        UrlParams      = require("utils/UrlParams").UrlParams;
+        UrlParams      = require("utils/UrlParams").UrlParams,
+        PathUtils              = require("thirdparty/path-utils/path-utils");
 
     // default async initExtension timeout
     var INIT_EXTENSION_TIMEOUT = 10000;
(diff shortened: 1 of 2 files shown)


ficristo:
I find a bit weak to rely on the semi-private API and the discussion linked above seems quite old.
I'm not against per se but is this the only way?

petetnt:
@ficristo It's the only way if you want to use the same configuration over different domains (for example extensions, web workers and node domains): it's a limitation of `RequireJS`.  The semi-private API part I agree with, but then again it hasn't changed in over 5 years: if it does, there will be a similar public API available I am sure. If not, you can always just reduplicate the configuration.

It isn't a very meaningful change as of now: everything will work as they used to. It's a blocker for something similar to #12006 (my `npm` concept) though and overall a low hanging fruit that falls under "nice to have".

ficristo:
Thank you for the clarification.

src/utils/ExtensionLoader.js
```diff
@@ -65,10 +65,13 @@ define(function (require, exports, module) {
...
-            "text" : srcPath + "/thirdparty/text/text",
-            "i18n" : srcPath + "/thirdparty/i18n/i18n"
-        };
+    // Retrieve the global paths
+    var globalPaths = brackets.getGlobalRequireJSConfig().paths;
+
+    // Convert the relative paths to absolute
+    Object.keys(globalPaths).forEach(function (key) {
+        globalPaths[key] = PathUtils.makePathAbsolute(srcPath + "/" + globalPaths[key]);
+    });
```
ficristo: I hope not, but if we'll find problems with this approach we'll have to dupe again this configuration:
https://github.com/adobe/brackets/blob/bf9b07d8e7b238a0f7f5392c8460ad34b63ee12c/src/main.js#L29
right?

src/utils/Global.js
```diff
@@ -125,5 +125,10 @@ define(function (require, exports, module) {
     // only be able to load modules that have already been loaded once.
     global.brackets.getModule = require;
 
+    // API for retrieving the global RequireJS config
+    global.brackets.getGlobalRequireJSConfig = function () {
```
ficristo: Are you going to expose this to extensions? Otherwise can you prefix with _ and add a comment that is for internal use only?

src/utils/ExtensionLoader.js
petetnt: @ficristo, yup. There shouldn't be any problems with the approach however.

petetnt:
Updated this PR against the latest master (non-global PathUtils), prefixed the API and made a note about internal use only.

src/utils/ExtensionLoader.js
```diff
@@ -41,7 +41,8 @@ define(function (require, exports, module) {
         FileUtils      = require("file/FileUtils"),
         Async          = require("utils/Async"),
         ExtensionUtils = require("utils/ExtensionUtils"),
-        UrlParams      = require("utils/UrlParams").UrlParams;
+        UrlParams      = require("utils/UrlParams").UrlParams,
+        PathUtils              = require("thirdparty/path-utils/path-utils");
```
ficristo: Small indentation issue here.

ficristo:
Do you have tryed it with an installer?

petetnt:
Fixed the indentation issue and added a missing piece from the original fix that had gone missing after rebasing. 

Haven't tried it with installer, but extensions seem to load and install just fine.

ficristo:
I tryed with a custom build installer on Linux (Ubuntu) and the TODO extension worked.
It works on Windows 10 too (I didn't try the installer here).
LGTM, thank you.

PR #12047, opened 2016-01-04
Simulate indentation of the project tree using some divs

### LINKED ISSUE #10574: Clicking to left of filename can collapse parent folder

ficristo:
instead of using combinations of padding/margin.

Fixes #10574

### DIFF
--- a/src/project/FileTreeView.js
+++ b/src/project/FileTreeView.js
@@ -59,6 +59,8 @@ define(function (require, exports, module) {
     var CLICK_RENAME_MINIMUM  = 500,
         RIGHT_MOUSE_BUTTON    = 2,
         LEFT_MOUSE_BUTTON     = 0;
+    
+    var INDENTATION_WIDTH     = 18;
 
     /**
      * @private
@@ -106,6 +108,40 @@ define(function (require, exports, module) {
         var width = measuringElement.width();
         measuringElement.remove();
         return width;
+    }
+
+    /**
+     * @private
+     *
+     * Create an appropriate div based thickness to indent the tree correctly.
+     *
+     * @param {int} depth - The depthness of the current node.
+     * @return {ReactComponent} - The resulting div.
+     */
+    function _createThickness(depth) {
+        return DOM.div({
+            className: "thickness",
+            style: {
+                width: (INDENTATION_WIDTH * depth) + "px"
+            }
+        });
+    }
+
+    /**
+     * @private
+     *
+     * Create the arrow icons used for the folders indented correctly.
+     *
+     * @param {int} depth - The depthness of the current node.
+     * @return {ReactComponent} - The resulting ins.
+     */
+    function _createAlignedIns(depth) {
+        return DOM.ins({
+            className: "jstree-icon",
+            style: {
+                marginLeft: (INDENTATION_WIDTH * depth) + "px"
+            }
+        }, " ");
     }
 
     /**
@@ -457,6 +493,19 @@ define(function (require, exports, module) {
                 'context-node': this.props.entry.get("context")
             });
 
+            var liArgs = [
+                {
+                    className: this.getClasses("jstree-leaf"),
+                    onClick: this.handleClick,
+                    onMouseDown: this.handleMouseDown,
+                    onDoubleClick: this.handleDoubleClick
+                },
+                DOM.ins({
+                    className: "jstree-icon"
+                }, " "),
+            ];
+            var thickness = _createThickness(this.props.depth);
+
             if (this.props.entry.get("rename")) {
                 nameDisplay = fileRenameInput({
                     actions: this.props.actions,
@@ -464,25 +513,20 @@ define(function (require, exports, module) {
                     name: this.props.name,
                     parentPath: this.props.parentPath
                 });
+
+                liArgs.push(thickness);
             } else {
                 // Need to flatten the argument list because getIcons returns an array
                 var aArgs = _.flatten([{
                     href: "#",
                     className: fileClasses
-                }, this.getIcons(), name, extension]);
+                }, thickness, this.getIcons(), name, extension]);
                 nameDisplay = DOM.a.apply(DOM.a, aArgs);
             }
 
-            return DOM.li({
-                className: this.getClasses("jstree-leaf"),
-                onClick: this.handleClick,
-                onMouseDown: this.handleMouseDown,
-                onDoubleClick: this.handleDoubleClick
-            },
-                DOM.ins({
-                    className: "jstree-icon"
-                }, " "),
-                nameDisplay);
+            liArgs.push(nameDisplay);
+
+            return DOM.li.apply(DOM.li, liArgs);
         }
     }));
 
@@ -668,6 +712,7 @@ define(function (require, exports, module) {
             if (isOpen && children) {
                 nodeClass = "open";
                 childNodes = directoryContents({
+                    depth: this.props.depth + 1,
                     parentPath: this.myPath(),
                     contents: children,
                     extensions: this.props.extensions,
@@ -698,11 +743,13 @@ define(function (require, exports, module) {
                 });
             }
 
+            var thickness = _createThickness(this.props.depth);
+
             // Need to flatten the arguments because getIcons returns an array
             var aArgs = _.flatten([{
                 href: "#",
                 className: directoryClasses
-            }, this.getIcons()]);
+            }, thickness, this.getIcons()]);
             if (!entry.get("rename")) {
                 aArgs.push(this.props.name);
             }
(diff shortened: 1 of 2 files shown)


ficristo:
@petetnt @busykai @nethip 
Could you take a look?

src/project/FileTreeView.js
```diff
@@ -109,6 +111,40 @@ define(function (require, exports, module) {
...
     /**
+     * @private
+     *
+     * Create an appropriate div based thickness to indent the tree correctly.
+     *
+     * @param {int} depth - The depthness of the current node.
+     * @return {ReactComponent} - The resulting div.
+     */
+    function _createThickness(depth) {
+        return DOM.div({
```
dibosh: Does this `div` have defined `display` as `inline-block`?

src/project/FileTreeView.js
```diff
@@ -109,6 +111,40 @@ define(function (require, exports, module) {
...
+            className: "thickness",
```
dibosh: Rather than making it dependant on some css class, wouldn't it be better if we can define all the styles inside `style` property?

src/project/FileTreeView.js
dibosh: I just saw it is defined in the `thickness` class, may be we can just define it inside the `style` prop?

ficristo:
@dibosh Thank you to have looked at it.
I think is better to use a class so extension authors can override it.
I use the style only because I need it to be dynamic.
If there is an agreement to use only the style prop is fine by me.

ficristo:
@MarcelGerber if you have a bit of time could you take a look?

src/project/FileTreeView.js
```diff
@@ -109,6 +111,40 @@ define(function (require, exports, module) {
     }
 
     /**
+     * @private
+     *
+     * Create an appropriate div based thickness to indent the tree correctly.
+     *
+     * @param {int} depth - The depthness of the current node.
```
MarcelGerber: Nit, but `depthness` is no word ;)

MarcelGerber:
LGTM, but I'd like to have someone else have a look at the React code (even though it looks straighforward) since I don't know nothing about React.

ficristo:
@MarcelGerber  Thank you.
Maybe @busykai or @petetnt could you take a look?

src/project/FileTreeView.js
```diff
@@ -457,32 +493,40 @@ define(function (require, exports, module) {
...
+                    className: this.getClasses("jstree-leaf"),
+                    onClick: this.handleClick,
+                    onMouseDown: this.handleMouseDown,
+                    onDoubleClick: this.handleDoubleClick
+                },
+                DOM.ins({
+                    className: "jstree-icon"
+                }, " "),
+            ];
+            var thickness = _createThickness(this.props.depth);
```
petetnt: Nit: Consecutive `var`s are usually either separated with a dot or new line

src/project/FileTreeView.js
```diff
@@ -59,6 +59,8 @@ define(function (require, exports, module) {
     var CLICK_RENAME_MINIMUM  = 500,
         RIGHT_MOUSE_BUTTON    = 2,
         LEFT_MOUSE_BUTTON     = 0;
+    
+    var INDENTATION_WIDTH     = 18;
```
petetnt: Should this be configurable? It wasn't with the CSS but could be a nice to have when it's in code.

src/project/FileTreeView.js
petetnt: I think I agree with @dibosh's suggestion of just using inline-styles.

src/project/FileTreeView.js
```diff
@@ -109,6 +111,40 @@ define(function (require, exports, module) {
...
+     * @param {int} depth - The depthness of the current node.
+     * @return {ReactComponent} - The resulting ins.
+     */
+    function _createAlignedIns(depth) {
+        return DOM.ins({
+            className: "jstree-icon",
+            style: {
+                marginLeft: (INDENTATION_WIDTH * depth) + "px"
+            }
+        }, " ");
```
petetnt: No need for the second argument (see https://jsfiddle.net/qLLs0ejk/)

src/project/FileTreeView.js
```diff
@@ -457,32 +493,40 @@ define(function (require, exports, module) {
...
+            var liArgs = [
+                {
+                    className: this.getClasses("jstree-leaf"),
+                    onClick: this.handleClick,
+                    onMouseDown: this.handleMouseDown,
+                    onDoubleClick: this.handleDoubleClick
+                },
+                DOM.ins({
+                    className: "jstree-icon"
+                }, " "),
```
petetnt: No need for the second argument (see https://jsfiddle.net/qLLs0ejk/)

src/styles/jsTreeTheme.less
```diff
@@ -46,6 +46,7 @@ li.jstree-closed > ul { display:none; }
 .jstree-brackets .jstree-open > ins { background-position:-72px 0; }
 .jstree-brackets .jstree-closed > ins { background-position:-54px 0; }
 .jstree-brackets .jstree-leaf > ins { background-position:-36px 0; }
+.jstree-brackets .thickness { display: inline-block; }
```
petetnt: See the comment on using inline-styles instead.

petetnt:
Made a few comments :+1:

src/project/FileTreeView.js
ficristo: @petetnt If really needed I prefer do it in a different PR.

ficristo:
I've updated the PR.

src/project/FileTreeView.js
petetnt: That's fine by me :)

src/project/FileTreeView.js
```diff
@@ -114,35 +114,36 @@ define(function (require, exports, module) {
...
+     * @param {int} depth The depth of the current node.
+     * @return {ReactComponent} The resulting div.
      */
     function _createThickness(depth) {
         return DOM.div({
             className: "thickness",
             style: {
-                width: (INDENTATION_WIDTH * depth) + "px"
+                display: "inline-block",
+                width: INDENTATION_WIDTH * depth
```
petetnt: Unitless value here, does React automatically convert it to `px`?

ficristo:
@petetnt I tryed to remove the unit and it worked...
Now I looked at the [docs](https://facebook.github.io/react/tips/style-props-value-px.html) and if you don't specify the unit most of the styles are automatically converted to use `px`.

MarcelGerber:
@ficristo @petetnt I suppose this is ready to go?

petetnt:
@MarcelGerber LGTM, but has merge conflicts after some recent changes.

ficristo:
I made some changes so to fix the rename case too. So it needs another look.

src/project/FileTreeView.js
MarcelGerber: As it turns out in #11717, the indentation should only be 10px, and now that it's merged, that's the behaviour on master, too. I think it looks way cleaner.

MarcelGerber:
Renaming a nested folder doesn't really work.
image

ficristo:
@MarcelGerber I fixed that issue, but now when you rename a folder the element with the children (`ul`) move a little. I wasn't able to find the cause, and I think it is a minor issue.

There is a `renameBehavior` function that should be prevent the directories to collapse, but seems to not work, that's why I've added some more `stopPropagation`.
I don't know enough of react to really look at this problem.

MarcelGerber:
You probably didn't see this comment:
As it turns out in #11717, the indentation (and as such, `INDENTATION_WIDTH`) should only be 10px, and now that it's merged, that's the behaviour on master, too. I think it looks way cleaner.

MarcelGerber:
If you can somehow manage to not apply [`.jstree-rename-input`](https://github.com/adobe/brackets/blob/b1a24ba2665c62ebfd3031f8cf1aa25c233177ed/src/styles/jsTreeTheme.less#L155-L156)'s `top`/`left` rules to the folders, but only to files, we can get rid of the jumping box.

ficristo:
I noticed another small issue were selecting a folder would result in a blue border.
So I reworked the code for render the folder to be more similar to the one of the files: in the renaming case the html layout is a bit different in the sense the input is not anymore in a link tag.
And then hopefully fix the bouncing behaviour: I applyed a style both to folders and files to which doesn't change anything visibily.

MarcelGerber:
LGTM. Going for it.

ficristo:
Thank you @MarcelGerber @petetnt @dibosh for review

PR #12053, opened 2016-01-05
Wrap indent-key to parentheses in .eslintrc

petetnt:
The key `indent` was missing parentheses in `.eslintrc`, rendering the file invalid JSON. 

This PR fixes that simple error.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -43,7 +43,7 @@
         "new-parens": 2,
         "no-new-object": 2,
         "no-invalid-this": 0,
-        indent: [0, 4],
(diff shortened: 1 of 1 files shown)


.eslintrc
```diff
@@ -43,7 +43,7 @@
         "new-parens": 2,
         "no-new-object": 2,
         "no-invalid-this": 0,
-        indent: [0, 4],
+        "indent": [0, 4],
```
MiguelCastillo: Awesome - thank you!!!

PR #12065, opened 2016-01-08
Fix FlipView / WorkingSet / SplitViewSameDoc integration issues

petetnt:
This PR is a continuation to #12060 and is similarly targeted to @swmitra's `BugFixesFor1.6Beta` branch instead of the master.

This PR fixes following issues:
#### Dirty doc working set flip issue

> Issue 3( Minor): 
> 
> As I have mentioned earlier in my same doc split PR , the file which was not present in working set , gets added to the working set after flip. The fix is not trivial in this case , I tried commenting the line where we are triggering working set event but looks like that results in stale mrulist entry. 

We check in [Pane.js#L585](https://github.com/petetnt/brackets/blob/ddcbcca8f05c89806651103f37433769fb305b08/src/view/Pane.js#L585) if the current file being flipped is actually on the working set: if it isn't, we destroy the view and return as resolved instead of adding it to the other `Pane`s `_viewList` 
#### Flip the behavior of `focused-editor-after-flip`

As discussed in https://github.com/adobe/brackets/pull/12060#issuecomment-169936958, this commit essentially flips the following commit around: https://github.com/petetnt/brackets/commit/8bc2845d06bf3efa3b3d34ad860895b336f09ec4 

This way the focused editor after flipping any view is the same one that was focused before the flip.
#### Prevent splicing the wrong file when the file isn't on the working set

> https://github.com/adobe/brackets/pull/12060#issuecomment-169934931

This issue was caused by the same problem than the previous one.
#### Unify handling of flipping and drag & dropping across working set

> https://github.com/adobe/brackets/pull/12060#issuecomment-169935671

By closing duplicate entries before flipping the view (if it already is open on the other pane) we make sure that `D&D` and `FlipView` scenarios handle the same way: currently the pane the file gets dragged/flipped to stays open_[1]_. This also fixes issues where clicking the `FlipView` button duplicates the current view instead of just moving it to another one (when the same file had previously been opened and closed in the other pane). 

/ping @swmitra :memo: 

This PR together with #12060 still needs some extensive unit tests, but I think those are easier to add in bulk when we get the baseline functionality fixed.

_[1]_ This would need some kind of merge functionality as discussed in  https://github.com/adobe/brackets/pull/12060#issuecomment-169935671, but I don't think it's possible to land in 1.6.

### DIFF
--- a/src/view/Pane.js
+++ b/src/view/Pane.js
@@ -56,8 +56,8 @@
   *
   * Pane Object Events:
   *
-  *  - viewListChange - Whenever there is a file change to a file in the working set.  These 2 events: `DocumentManger.pathRemove`
-  *  and `DocumentManger.fileNameChange` will cause a `viewListChange` event so the WorkingSetView can update.
+  *  - viewListChange - Whenever there is a file change to a file in the working set.  These 2 events: `DocumentManager.pathRemove`
+  *  and `DocumentManager.fileNameChange` will cause a `viewListChange` event so the WorkingSetView can update.
   *
   *  - currentViewChange - Whenever the current view changes.
   *             (e, newView:View, oldView:View)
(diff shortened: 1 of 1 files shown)


swmitra:
@petetnt Sorry for late reply. I have tested most of the scenarios along with the reported scenarios for the issues. It looks fine to me. Great work as always! :+1: 

Merging this PR with the bug fixes branch swmitra/BugFixesFor1.6Beta

Will continue with some more testing.

petetnt:
:+1:

PR #12076, opened 2016-01-12
Update README.md

nethip:
As filing an issue with Brackets is now the only way for users to submit translations suggestions (Fr and JP), updated README to reflect that.

@ryanstewart


abose:
LGTM. merging.

PR #12092, opened 2016-01-17
Update year to 2016 in LICENSE and about-dialog.html

prayagverma:


LICENSE
```diff
@@ -1,19 +1,19 @@
-Copyright (c) 2012-2015 Adobe Systems Incorporated. All rights reserved.
- 
+Copyright (c) 2012-2016 Adobe Systems Incorporated. All rights reserved.
+
 Permission is hereby granted, free of charge, to any person obtaining a
```
vanjing: Permission is hereby granted, with fee and with charge when using and copy of this software with restriction and associated documentation files (the "Software"),

pra85:
@abose I have updated this pull request to follow the same format as https://github.com/adobe/brackets-shell/pull/545 .

For the files which have only a single year mentioned in their License header (like https://github.com/adobe/brackets/blob/master/src/LiveDevelopment/Agents/ConsoleAgent.js ) should I make their year format as  `Mentioned Year - present` ?

Also should I do a single commit with these updates or should I break them up by years mentioned in them (mainly asking how to break them up in stages)

abose:
`Mentioned Year - present` would be fine in those cases.
It would be good to do all in a single PR, but with different commits- maybe folder wise in the source tree.
Also, note that 3'rd party code licenses may not be altered. Only change `**ADOBE licenses**` to the new format.

pra85:
@abose Updated all the **`Adobe Systems Incorporated.`** licenses and committed them folder wise.

abose:
Verified all licences. 
Merging as confirmed with @ryanstewart 
Thanks @pra85 for making this change.

PR #12110, opened 2016-01-20
Remove basePadding from indentwrap hack

petetnt:
As noted by @redmunds on Slack there's an slightly ankward gap between the gutters and the text

image

This is due to the indent-wrap hack in https://github.com/adobe/brackets/pull/11909: the texts are pushed 4 pixels back by the hack. This PR compensates for that by pulling the editor back -4 pixels:

image

### DIFF
--- a/src/editor/Editor.js
+++ b/src/editor/Editor.js
@@ -1038,10 +1038,10 @@ define(function (require, exports, module) {
         });
         // For word wrap. Code adapted from https://codemirror.net/demo/indentwrap.html#
         this._codeMirror.on("renderLine", function (cm, line, elt) {
-            var charWidth = self._codeMirror.defaultCharWidth(), basePadding = 4;
+            var charWidth = self._codeMirror.defaultCharWidth();
             var off = CodeMirror.countColumn(line.text, null, cm.getOption("tabSize")) * charWidth;
             elt.style.textIndent = "-" + off + "px";
-            elt.style.paddingLeft = (basePadding + off) + "px";
+            elt.style.paddingLeft = off + "px";
         });
     };
 


abose:
tagging @swmitra

redmunds:
Thanks for this fix. This is better, but I now see a gap between gutter and Active Line highlighting:

brackets-1 6-12110

wyattbiker:
Same issue. Here is my original question and also temporary fix for now.

https://groups.google.com/forum/?hl=en#!topic/brackets-dev/6ri52cF2FB8

petetnt:
@redmunds Which theme is that? Looks like the regular theme, but grayer? Cannot replicate on my machine.

However, I pushed in alternative (but similar) solution in https://github.com/adobe/brackets/commit/b461a1a20c1fafd9a5428c236129391d9a72c878 if you want to give it a try.

lkcampbell:
@petetnt has anyone figured out when this gap appeared and why is started happening?

I only noticed it fairly recently. If we understood what introduced the problem we could address it where it was introduced instead of hacking the style sheet.

lkcampbell:
@petetnt, never mind, I see what the problem is. I didn't realize that ugly Code Mirror word wrap hack made it into the core code. We really should have made that a preference.

petetnt:
FWIW I think the padding has been working pretty well (sans the 4px space issue this PR is trying to account for). I do think that the hack is pretty ugly too, but indent wrapping is much more crucial IMHO.

petetnt:
In #12184 there's another case where the gutter shows up (and screws the fold marks too) with this fix. The thing is that I can't replicate it on Windows 10, 8.1 nor Mac OS, all with same version of Brackets Shell and same commit.

abose:
I too cant replicate the `gap issue` .

ficristo:
LGTM, even I can not replicate the gap issue.

redmunds:
@abose Can we get this fix in for 1.7 ?

petetnt:
@redmunds do you have any idea what causes the gutter thing yet? I have tried this on 10+ computers and I still haven't been able to replicate the issue you were having above 😕

src/styles/brackets_codemirror_override.less
```diff
@@ -27,6 +27,7 @@
 
     > * {
         text-indent: 0px;
+        margin-left: -4px;
```
busykai: @petetnt could this be done as part of [the hack itself](https://github.com/adobe/brackets/pull/11909/files#diff-3b7e955135e4a77b20767ed40e551395R990) as opposed to stylesheet? this way the value of the shift (`4px`) will be shared and not duplicated.

src/styles/brackets_codemirror_override.less
```diff
@@ -122,7 +123,7 @@ div.CodeMirror-cursors {
 
 .CodeMirror-lines {
     padding: @code-padding 0;
-
+
```
busykai: please undo this change.

abose:
+1 this has been a long pending issue that's a must fix. Tagging @swmitra 
also +1 It will be neater to club the hacks together in one place as @busykai pointed out.

src/styles/brackets_codemirror_override.less
RaymondLim: @petetnt Not sure how to fix it in the hack itself, but I found out that just removing `padding: 0 @code-padding 0 0;` on line 26 also fixes the issue.

petetnt:
Made some changes to this, 3ff302e now contains the compensation in the padding hack itself. Not sure if it's the most optimal solution either but seems to work for me.

petetnt:
Build failed to the `cla-check` so disregard that.

RaymondLim:
@petetnt I confirmed that your fix is good with the default "show line numbers" on. If you turn off line numbers, then all lines are shifted to the left and your change does not fix in that case. So I believe we also need to adjust `@code-padding` in brackets_codemirror_override.less file to get it right.

petetnt:
Thanks for the feedback @RaymondLim , I'll get that fixed ASAP 👍

petetnt:
Somewhat related: #12630

petetnt:
@RaymondLim adjusting `@code-padding` to 0 showed just similar issues for me (albeit upside down from the screenshots in this thread 🍔. 

However, removing the `basePadding` variable altogether from the hack seemed to fix everything, as it is not really needed for the hack itself (Brackets pads itself with `@code-padding` from top of the editor and the left side of the `pre`s).

Does it work for you / others?

RaymondLim:
@petetnt Sorry, I was wrong to say that I don't have any other changes. Actually, I do have the following change in Editor.js that I believe is from your previous commit.

```
        // For word wrap. Code adapted from https://codemirror.net/demo/indentwrap.html#
        this._codeMirror.on("renderLine", function (cm, line, elt) {
            var charWidth = self._codeMirror.defaultCharWidth(), basePadding = 4;
            var off = CodeMirror.countColumn(line.text, null, cm.getOption("tabSize")) * charWidth;
            elt.style.textIndent = "-" + off + "px";
            elt.style.paddingLeft = (basePadding + off) + "px";
            $(elt).children().each(function (i, child) {
                child.style.marginLeft = -basePadding + "px";
            });
        });
```

petetnt:
Yeah, but that is not needed anymore. Together with #12630 all of the issues should be resolved. (The PR's solve different issues)

ficristo:
LGTM, if we'll find more issue we can look at them in another PR.

Issue #12129, opened 2016-01-22
Core: Sidebar can be shown in no distraction mode.


chunyangwang:
Steps: 
1. Open index.html
2. Click View -> No Distractions
3. Click View -> Show Sidebar

Expect result:
Sidebar can not be shown.

Actual result:
Sidebar was shown as well as No Distractions was checked.

![1-22-2016 3-40-44 pm](https://cloud.githubusercontent.com/assets/16833083/12505430/daf02cfc-c121-11e5-8763-9135801bc1d6.png)

bhldev:
Pull open

nethip:
@bhldev  We essentially are collapsing the project tree. You can still pull open the project tree in no distraction mode. Could you double check on this and update this thread?

bhldev:
This functions as follows currently: if the sidebar is on and no distraction mode is turned on, the sidebar gets turned off. If No distraction mode is on and the sidebar gets turned on, the sidebar appears still. Please update with intended functionality so I can either retract pull or modify pull to match UX experience. Thanks.

abose:
Allowing the user to open the sidebar was conscious decision we made . This is to help with users to see the project contents even in no distraction mode for some time if he choose. Closing .

PR #12148, opened 2016-01-25
Feature : Remove Show Sidebar on No Distraction Mode

brian-lim-42:
Removes "Show Sidebar" and "Hide Sidebar" from the View Menu when in No Distraction mode to enhance user experience (Issue #12129)

### DIFF
--- a/src/extensions/default/NoDistractions/main.js
+++ b/src/extensions/default/NoDistractions/main.js
@@ -136,10 +136,12 @@ define(function (require, exports, module) {
 
     PreferencesManager.on("change", PREFS_PURE_CODE, function () {
         if (PreferencesManager.get(PREFS_PURE_CODE)) {
+            Menus.getMenu(Menus.AppMenuBar.VIEW_MENU).removeMenuItem(Commands.VIEW_HIDE_SIDEBAR);
             ViewUtils.hideMainToolBar();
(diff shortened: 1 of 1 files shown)


petetnt:
Instead of explicitly removing the menu items, they most likely should just enabled/disabled with [`commandManager#setEnabled`](https://github.com/adobe/brackets/blob/2a832d0059c9b5b67fb868a8672764fe728ac7f7/src/command/CommandManager.js#L120) 

However, I am not entirely sure if #12129 is an *entirely* valid bug in the first place: opening the sidebar for a while to find a while seems somewhat valid use case, but it does bring some weird inter-op with the No-Distractions concept in general so I would be okay with disabling them too.

@abose opinions?

abose:
@bhldev  Thank you for contributing to Brackets. You will need to Accept the [Brackets CLA](http://dev.brackets.io/brackets-contributor-license-agreement.html) before we can merge this pull request.

abose:
Allowing the user to open the sidebar was conscious decision we made as @petetnt said. This is to help with users to see the project contents even in no distraction mode for some time if he choose. Closing this PR as that was a design choice.

@bhldev If you are interested in contributing to features in brackets, please take a look at our trello board at https://trello.com/b/LCDud1Nd/brackets to see features requested by the community.

Issue #11488, opened 2015-07-27
Resizer module uses depreated function


le717:
(Split from https://github.com/adobe/brackets/pull/11174#issuecomment-119032811)

`src/utils/Resizer.js` uses `PreferencesManager.convertPreferences`. which is deprecated, so the error is being thrown in core.  I have no clue how that preference should be updated to the current system (prefixed? Global preferences file), and I am unsure if deprecation errors should be thrown in core (I think historically that is a "no").

ficristo:
/cc @peterflynn @abose
Do you know how to fix this?

abose:
also see related use in `src/project/ProjectManager.js`.

sprintr:
I removed all the code using deprecated code and Brackets kept on working. I seem to have lost that branch.

abose:
Was there a lot of places using deprecated API in core?

sprintr:
15-20 places maybe. I'll submit a PR tonight.

abose:
:+1: Thanks @sprintr

sprintr:
@swmitra This is fixed and should be closed.

PR #12182, opened 2016-02-01
Remove deprecated code using the old preferences system

sprintr:
Removed all the occurrences of `PreferencesManager.convertPreferences`. I have also removed its definition but that is a separate commit just in case we wanted to revert it.

Follow up of #11488 and #11174

/CC: @abose @MarcelGerber @redmunds @peterflynn @le717


le717:
I had put up a PR to drop the old preferences system entirely in #9752, if you wanted to go that far. Since the conversion code from old to new is being removed (assuming I am reading the diff correctly), we would still have the old system lying around unused.

You may want to considering dropping [`setValueAndSave`](https://github.com/le717/brackets/commit/72291f99eceea1502a8efef09aba58c6cf4f2a11) too.

My thought is, while the old preferences system has been deprecated for ages, `convertPreferences` only got a deprecation warning in v1.4 (#11174). If the definition was to be removed, has enough time lapsed for extension authors to update? IIRC, last time we looked at this, all affected extension authors had been contacted but only the actively maintained extensions were updated. Honestly, if _by now_ if the extensions have not been updated, then there's no need to keep dragging out feet holding out.

sprintr:
@le717 The purpose of this PR was to make sure that no warnings are thrown from the core modules. I agree with you on the removal of deprecated code since it has been a quite a while since when it was deprecated. If they have not updated until now, they may not be updated ever.

I also went ahead and removed the old system from the unit tests too.

MiguelCastillo:
@sprintr @le717 It sounds reasonable to remove the whole thing. Do you guys think they should two separate PRs?

zaggino:
This now has conflicts, @sprintr would you mind merging master to your branch so I can pull this?

Issue #12187, opened 2016-02-02
Use a filewatcher library


ficristo:
Right now we use an outdated version of fsevents and fsevents_win to watch files.
I think using a library better maintained would be a nice alternative option.
I tried once to use [chokidar](https://github.com/paulmillr/chokidar) but I've had some performance problem: maybe because it watches file recursively and Brackets doesn't. I don't know.
Another library could be https://github.com/facebook/watchman/tree/master/node

ficristo:
Or this https://github.com/amasad/sane that can use watchman under the hood but it's not necessary.

petetnt:
Gaze is (yet another) option: https://github.com/shama/gaze

MiguelCastillo:
I haven't done much research on this in a few months, but this https://github.com/paulmillr/chokidar was the better library I have seen around.  It is used by watchify (browserify's watching plugin).

MiguelCastillo:
@ficristo I have not noticed performance issues...  But perhaps that's due to the fact that it is part of build process where it is really hard to see where performance is lost.  How did you go about testing this performance issues you saw? Can you share you test setup so that I can run them?  I am really curious :)

MiguelCastillo:
@ficristo is the motivation purely about switching to a more actively maintained library?

ficristo:
No, there aren't specific problems.
I only think using up to date library is (almost) always a plus.

MiguelCastillo:
@ficristo sounds good to me. this will require quite a bit testing; you got different platforms that behave differently. If my memory serves me right, @peterflynn was one of the main dev involved in the current  implementation. Perhaps he could chime in when he gets a sec.

You mentioned a really key point.  chokidar will recursively add watchers and brackets does not.  This is concerning. :)

Curious to know how does this works in Ubuntu... Have you tried it?

ficristo:
No, I didn't. I'm not really Linux friendly...

MiguelCastillo:
Let's continue further discussion in the actual PR https://github.com/adobe/brackets/pull/12190#issuecomment-183676768.

ficristo:
Another couple of interesting libs.
https://github.com/atom/node-pathwatcher
https://github.com/bevry/watchr

ficristo:
This PR https://github.com/adobe/brackets/pull/12647 is now using chokidar.

ficristo:
[...]

PR #12190, opened 2016-02-03
[PoC] Use chokidar instead of the fsevents/fsevents_win combo

ficristo:
This change still cause significant cpu usage.
Only looked at the Activity Monitor on OSX or the Task Manager on Windows.
See #12187


MiguelCastillo:
@ficristo Ha! build failed because 

```
data: {"message":"API rate limit exceeded for 52.22.60.255. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://developer.github.com/v3/#rate-limiting"}
```

Maybe trigger the build again and see what happens.

ficristo:
I had totally missed the `depth` parameter. Now it's a bit better.

ficristo:
Actually it seems to me that the file watcher is recursive on Windows and OSX.
So I removed the `depth` parameter and `added` the ignore one.

ficristo:
@petetnt Could you try your extension https://github.com/petetnt/corrupt-brackets-history with this PR?
I was able to run it once without errors, but running again it failed.
It's not urgent, but I'm curios if this could at least help.
Be aware that, expecially on startup, this is CPU intensive.

petetnt:
@ficristo Tested it, first corruption occurred on the 116th try :sob: 

Otherwise I had no issues.

MiguelCastillo:
@ficristo let me take a step back.  File watching is a **_very_** critical path piece functionality that we cannot regress on.  So, it is important that we understand the impact of this proposed changes.

Do you think you can help us understand:

1. Differences in behavior between what we have and the new library.
2. Performance differences.
3. Platform support.

These three questions are very important before we continue going down this path.  Thank you so much for understanding why I am being very cautious about these changes :)

ficristo:
@MiguelCastillo This was an intersting thing to explore for me. I don't mind if it takes it's time.
The situations for what I've understood is as follows.

Currently:
- OSX: an old version of fsevents, customized to work for 32 bit OS (recursive)
- Windows: a custom version of fsevents (recursive)
- Linux: native node api (non recursive)

With chokidar:
- OSX: a more up to date version of fsevents, 64 bit OS (recursive)
- Windows: native node api (recursive)
- Linux: native node api (recursive)

This probably will require an update to CEF where only OSX 64 bit is supported.
The initial CPU usage is a bit high with chokidar, probably too much. There is some effort upstream to improve it.
That's what I learned so far.

MiguelCastillo:
@ficristo Thats an awesome set of information. I think it makes sense to gather some metrics to understand how we will be affected.  Updating CEF is a PITA.  We should explore if we have any plans to update soon.

ficristo:
I've tried to add the preferences for the paths to ignore but I seems to have encountered some circular reference problems...

zaggino:
@ficristo I've build shell using https://github.com/adobe/brackets-shell/pull/543 and used this PR as a base to run tests. I can see some fs watching logged to the console but running the tests fails:

image

ficristo:
@zaggino I fixed the tests.
I'm still not familiar enough with the Brackets code for file watching so if you want to make any changes go ahead.

zaggino:
Thanks @ficristo, will continue testing this today.

ficristo:
Replaced by https://github.com/adobe/brackets/pull/12647.

PR #12230, opened 2016-02-19
node crash fix: error on live preview server creation.

abose:
handle error message on live preview server creation. 
Else node crashes if the port is already in use.


abose:
Merging as PR relates to error handling that has been validated internally in the team.

PR #12238, opened 2016-02-22
Add gyp and gypi file extensions to python language

ficristo:
The documentation here https://gyp.gsrc.io/docs/LanguageSpecification.md#Overview
says `The .gyp file syntax is a Python data structure.`


abose:
LGTM 
MErging.
Thanks @ficristo

PR #12243, opened 2016-02-25
Fix12131: release refresh 1.6: added one more shortcut ctrl+shift+1  to toggle panels

abose:
default toggle panel shortcut was ctrl+shift+` as it is present in one vertical line in the keyboard. However, we later learnt from IQE team than non-English keyboards does not have the ` char. So added one more shortcut ctrl+shift+1 which will be preferred.

Long term solution for locale specific shortcuts maybe to have a locale filter just like the os filter we have while registering shortcuts. As we have already shipped with ctrl+shift+` removing that didn't look like a good idea, so maintained both.

### DIFF
--- a/src/extensions/default/NoDistractions/main.js
+++ b/src/extensions/default/NoDistractions/main.js
@@ -47,8 +47,10 @@ define(function (require, exports, module) {
     //key binding keys
     var togglePureCodeKey         = "Ctrl-Shift-2",
         togglePureCodeKeyMac      = "Cmd-Shift-2",
-        togglePanelsKey           = "Ctrl-Shift-`",
-        togglePanelsKeyMac        = "Cmd-Shift-`";
+        togglePanelsKey           = "Ctrl-Shift-1",
+        togglePanelsKeyMac        = "Cmd-Shift-1",
+        togglePanelsKey_EN        = "Ctrl-Shift-`",
+        togglePanelsKeyMac_EN     = "Cmd-Shift-`";
 
     //locals
     var _previouslyOpenPanelIDs = [],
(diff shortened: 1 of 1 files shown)


abose:
@swmitra @petetnt

src/extensions/default/NoDistractions/main.js
```diff
@@ -160,7 +162,11 @@ define(function (require, exports, module) {
         Menus.getMenu(Menus.AppMenuBar.VIEW_MENU).addMenuItem(CMD_TOGGLE_PURE_CODE, "", Menus.AFTER, CMD_TOGGLE_PANELS);
 
         KeyBindingManager.addBinding(CMD_TOGGLE_PURE_CODE, [ {key: togglePureCodeKey}, {key: togglePureCodeKeyMac, platform: "mac"} ]);
+
+        //default toggle panel shortcut was ctrl+shift+` as it is present in one vertical line in the keyboard. However, we later learnt
```
petetnt: Nitpick: This comment could be something like "The default "Toggle Panels" shortcut was CTRL+SHIFT+` as it is present in one vertical line in the standard US keyboard. However, we later learned from the IQE team that some non-US keyboards do not have an dedicated `-key, so an additional shortcut "CTRL+SHIFT+1" was added"

petetnt:
Confirmed to work on non-US keyboard under Windows 10, LGMT with small language nitpick :+1:

swmitra:
LGTM :+1:

abose:
Thanks @petetnt looks like the PR got merged for the doc change :sweat_smile: .

petetnt:
No probs :joy_cat:

PR #12248, opened 2016-02-26
When flipping views, if the doc is already open on the other view, show it without closing the original pane

petetnt:
In https://github.com/adobe/brackets/commit/d72753b6b8bcc575f6e3cbbb96b4795a037a7740 there was an hotfix that prevented flipping files that were already open in the other pane (due to issues with merging views that have inline widgets open).

This PR changes it so that flipping an view that is already on the other pane just shows the other file in the other pane, being somewhat inline with the general `split-view-same-doc` concept. PR has an test included.

Tagging @swmitra

### DIFF
--- a/src/view/Pane.js
+++ b/src/view/Pane.js
@@ -212,6 +212,23 @@ define(function (require, exports, module) {
     }
 
     /**
+     * Ensures that the given pane is focused after other focus related events occur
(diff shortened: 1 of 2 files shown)


swmitra (approved):
Looks good to me and tested to be working fine.

PR #12353, opened 2016-04-08
Fix of not picking file modifications.

mfatekho:
Inapproriate use of win api in fs_even_node module. In fsevents_win.cpp there are a thread that monitors file system events queue and post messages to uv library. Proposed solution is to use extra bool variable in THREADINFO structure to control thread exit state
@busykai


src/filesystem/impls/appshell/node/node_modules/fsevents_win/fsevents_win.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014 - present Adobe Systems Incorporated. All rights reserved.
```
busykai: This change, the base of your commit, seems to be a left-over of the intermediate work. Could you please rebase the commit on top of `adobe/master`?

src/filesystem/impls/appshell/node/node_modules/fsevents_win/fsevents_win.cpp
```diff
@@ -1,5 +1,5 @@
...
+ * Copyright (c) 2014 Adobe Systems Incorporated. All rights reserved.
```
busykai: Please make sure that the copyright stays intact. We shouldn't be changing it and I didn't know the format was updated, sorry.

busykai:
@peterflynn, @bchintx, @abose: we were able to reproduce the filewatchers going offline. we expect this is supposed to fix #7638 and related issues. The issue occurs when some other tools interferes with the same file(s) causing "spurious" messages which result in [this condition](https://github.com/mfatekho/brackets/blob/d6ade93f0ca248dfa62598a6a505430e9185dbb5/src/filesystem/impls/appshell/node/node_modules/fsevents_win/fsevents_win.cpp#L356) resolves to `FALSE` and causes the filewatchers to quit abruptly.

The fix consists of keeping an internal flag of whether the filewatchers should quit instead of relying on specific parameters of the "terminate" message.

src/filesystem/impls/appshell/node/node_modules/fsevents_win/fsevents_win.cpp
```diff
@@ -337,7 +339,7 @@ DWORD WINAPI NodeFSEvents::Run(LPVOID lpData)
 	{
 		// iterate watching for directory changes until we're signalled to stop
 		BOOL bContinue = TRUE;
```
busykai: This variable should be taken off entirely. Since you removed the code at lines 386-390, `bContinue` never changes the value and having it around is excessive and not needed.

evgen48: According to changes now bContinue reflects status from winapi call at line 346  ReadDirectoryChangesW, in some situations that function may fail and return false according to MSDN, that leads to  looper exit. Do you propose to skip  error handling for that function and continue watching thread as well?

busykai: Nevermind, I simply missed that line. You're right, the logic's good.

busykai:
LGTM. Will merge soon.

swmitra:
LGTM :+1:

MiguelCastillo:
This looks good to me.  Totally got PTSD from old C++ days... :D

ingorichter:
Looks like a simple fix to get rid of the problem.

src/filesystem/impls/appshell/node/node_modules/fsevents_win/fsevents_win.cpp
bchintx: @mfatekho I think there's still a logic error here.  If the `ReadDirectoryChangesW()` call below fails, then `bContinue` is assigned the value of `FALSE`.  As a result, on the next iteration of the loop, the `while` condition below in line 342 will *still* be false, and we'll still get kicked out of the loop.

I'd partially agree with @busykai 's [comment](https://github.com/adobe/brackets/pull/12353/files#r59280730) above, and propose removing `bContinue` from the `while` loop condition expression in line 342.  (Although, per the above discussion, don't entirely remove the `bContinue` variable b/c it'll still be needed below to capture the result of `ReadDirectoryChangesW()`.

bchintx:
Please don't merge quite yet.  Please see my [comment inline above](https://github.com/adobe/brackets/pull/12353/files#r59806934), as I think there's still a logic error that will prevent the fix from working as expected.

ingorichter:
@bchintx I was thinking about getting rid of `bContinue`, but you said it way better than I could have done this. ;-)

src/filesystem/impls/appshell/node/node_modules/fsevents_win/fsevents_win.cpp
evgen48: Hi @bchintx there are several points with removing `bContinue` and leaving the rest untouched.
1. on line 247 there are `FILE_FLAG_OVERLAPPED` is set meaning  `ReadDirectoryChangesSW` is working in async mode. 
2. MSDN says that for asynchronous mode returning value is a result of task submission, false means we run out of some resources, handles, stack, ram, etc. 
3. Submitting new requests, e.g. calling `ReadDirectoryChangesSW` in that situation (if that ever happen i believe in low resource condition there will be lots of problems) makes situation even worse: I'm not sure whether that function will issue `SwitchToThread` call, if not we will have one cpu core load spot until low resource condition go away.
4. Further reading of possible error codes reported by `ReadDirectoryChangesW()` makes me think they all are not recoverable, so no point to keep that loop

I just wanted to say that if we simply ignore that error we have to think about possible short loop that burns CPU, and at least add `SwitchToThread()`, `Sleep(0)`, etc, if we decided to not abort watcher thread.

busykai: Let's just get rid of `bContinue` it bothers me (both its name and its use in the `while` condition). Just like @bchintx pointed out, let's wrap the `ReadDirectory...` call in an if, this way @evgen48's concern of not handling the error condition will be addressed too.

mfatekho:
Hi guys,
Refactored code a little due to your recommendations. As you suggested, I have removed bContinue variable and wrapped ReadDirectoryChangesW to if.

src/filesystem/impls/appshell/node/node_modules/fsevents_win/fsevents_win.cpp
```diff
@@ -338,18 +338,16 @@ DWORD WINAPI NodeFSEvents::Run(LPVOID lpData)
...
 				FILE_NOTIFY_CHANGE_FILE_NAME | FILE_NOTIFY_CHANGE_DIR_NAME | FILE_NOTIFY_CHANGE_ATTRIBUTES | FILE_NOTIFY_CHANGE_SIZE
 				| FILE_NOTIFY_CHANGE_LAST_WRITE | FILE_NOTIFY_CHANGE_LAST_ACCESS | FILE_NOTIFY_CHANGE_CREATION | FILE_NOTIFY_CHANGE_SECURITY,
-				&dwBytesReturned, lpThreadInfo->m_lpOverlapped, NULL);
-
-			if (bContinue)
-			{
+				&dwBytesReturned, lpThreadInfo->m_lpOverlapped, NULL)) {
+				break;
+			}
 				// wait for a signal
```
busykai: Lines 351-358 can be indented left now.

src/filesystem/impls/appshell/node/node_modules/fsevents_win/fsevents_win.cpp
```diff
@@ -385,7 +383,7 @@ DWORD WINAPI NodeFSEvents::Run(LPVOID lpData)
 						}
 					}
 				}
-			}
+
```
busykai: This line simply needs to be removed.

ingorichter:
Poking @bchintx to have another look.

bchintx:
Looks good to me.  Merging!

busykai:
Yay! Thanks @mfatekho and @evgen48 for root-causing and fixing this!

MegaArman:
Just trying out brackets for the first time today.  Here's my trouble scenario:
-have some files open 
-git checkout in my terminal
-the files are not reloaded automatically nor am I prompted right away to reload them.
-Write some code
-Hit ctrl-s and now I am given a dialogue window if I want to overwrite what's on disc. This is an issue because now I have to figure out if my changes to the file are actually OK
 
In the Geany editor under Edit->Preferences->Files-> Disc check timeout, you can set how often the disk is checked and it will prompt you to reload automatically

Is there something like that for Brackets?

PR #12372, opened 2016-04-19
Update Polish translation (checked and tested)

jurkian:
@abose as I promised, I've updated the Polish translation and tested it live on Brackets to make sure the user experience is good enough. I've also compared the other Polish translations and took the best parts of them, so I think it's the most up to date translation available as for now.

I didn't touch the Brackets preferences comments, because I think preferences file is not for a typical user and is best understood in English. In Polish lots of these elements would look quite weird if translated.

@winek could you please have a look at this?


abose:
@jurkian Thank you for Helping us with the translation. :fireworks: 
The review for this PR is already complete at https://github.com/adobe/brackets/pull/11198. 
Verified the change.
Merging.

PR #12386, opened 2016-04-27
ImagePreview URL protocol validation

swmitra:
This PR is aimed to restrict/validate non-empty protocol for image preview URL. 

Ping @petetnt @abose @nethip for review.

### DIFF
--- a/src/extensions/default/QuickView/main.js
+++ b/src/extensions/default/QuickView/main.js
@@ -60,6 +60,9 @@ define(function (require, exports, module) {
 
     var styleLanguages = ["css", "text/x-less", "sass", "text/x-scss", "stylus"];
 
+    // List of protocols which we will support for image preview urls
+    var validProtocols = ["data:", "http:", "https:", "ftp:", "file:"];
+
     prefs = PreferencesManager.getExtensionPrefs("quickview");
     prefs.definePreference("enabled", "boolean", true, {
         description: Strings.DESCRIPTION_QUICK_VIEW_ENABLED
@@ -474,7 +477,8 @@ define(function (require, exports, module) {
 
         // Determine whether or not this URL/path is likely to be an image.
         var parsed = PathUtils.parseUrl(tokenString);
-        var hasProtocol = parsed.protocol !== "";
+        // If the URL has a protocol, check if it's one of the supported protocols
+        var hasProtocol = parsed.protocol !== "" && validProtocols.indexOf(parsed.protocol.trim()) !== -1;
         var ext = parsed.filenameExtension.replace(/^\./, '');
         var language = LanguageManager.getLanguageForExtension(ext);
         var id = language && language.getId();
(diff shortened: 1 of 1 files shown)


src/extensions/default/QuickView/main.js
```diff
@@ -474,7 +477,8 @@ define(function (require, exports, module) {
 
         // Determine whether or not this URL/path is likely to be an image.
         var parsed = PathUtils.parseUrl(tokenString);
-        var hasProtocol = parsed.protocol !== "";
+        // If the URL has a protocol, check if it's one of the supported protocols
+        var hasProtocol = parsed.protocol !== "" && validProtocols.indexOf(parsed.protocol.trim()) !== -1;
```
abose: to lower case

swmitra: Done. Good catch :+1:

src/extensions/default/QuickView/main.js
```diff
@@ -474,7 +477,8 @@ define(function (require, exports, module) {
 
         // Determine whether or not this URL/path is likely to be an image.
         var parsed = PathUtils.parseUrl(tokenString);
-        var hasProtocol = parsed.protocol !== "";
+        // If the URL has a protocol, check if it's one of the supported protocols
+        var hasProtocol = parsed.protocol !== "" && validProtocols.indexOf(parsed.protocol.trim().toLowerCase()) !== -1;
```
petetnt: Maybe something like `hasValidProtocol` would be a better name for the variable?

Is the empty check (`parsed.protocol !== ""`) still needed?

swmitra: Yeah, that would have been a better name , but as this was an existing variable , I haven't renamed it. 
Secondly , having (parsed.protocol !== "") check will help in cutting down the array search in lot of cases like a relative path image url.

petetnt: @swmitra yup, that makes sense.

ficristo:
This seems to workaround a bug in the library:
`require("thirdparty/path-utils/path-utils").parseUrl("https://foo.bar.com:8080/baz/").protocol`
"https:"
`require("thirdparty/path-utils/path-utils").parseUrl("foo.bar.com:8080/baz/").protocol`
"foo.bar.com:"
`require("thirdparty/path-utils/path-utils").parseUrl("foo.bar.com/baz/").protocol`
""
I would expect an empty string for the second test too, where there is a port.

nethip:
@swmitra Did you get a chance to address comment put in by @ficristo ?

ficristo:
@nethip I've filed a bug at https://github.com/jblas/path-utils/issues/2
I am not sure if there is anything more to do here.

nethip:
LGTM

nethip:
Thanks @ficristo for filing a bug. 

And thanks @swmitra for this PR

PR #12389, opened 2016-04-28
Homepage url validation for extension listing

swmitra:
This PR adds some extra validation on homepage entry in extension listing under ExtensionManager view to make sure local binaries are not getting listed as homepage url.

Ping @nethip @abose @peterflynn  for review.


src/extensibility/ExtensionManagerView.js
```diff
@@ -336,6 +337,19 @@ define(function (require, exports, module) {
             context[helper] = registry_utils[helper];
         });
 
+        // Do some extra validation on homepage url to make sure we don't end up executing local binary
+        if (context.metadata.homepage) {
+            var parsed = PathUtils.parseUrl(context.metadata.homepage);
+            // Check if the homepage refers to a local resource
+            if (parsed.protocol.trim().toLowerCase() === "file:") {
```
nethip: @swmitra If have something like `C:\Windows\System32\notepad.exe`, will `parsed.protocol` list it under `file:`?

swmitra: Great catch as always:+1: @nethip .
You are right. It gives us "C:" as protocol :smile: 
Seems like we can use browser itself to give us all the relevant details of a url instead of relying on the text processing done by path-utils.

swmitra:
@nethip Local resources without explicit mention of protocol should be checked now with the changes. Only http/https resources will get skipped.

nethip:
@swmitra There seems to be some undesired changes, now part of the PR. Could you check and revert them?

src/extensibility/ExtensionManagerView.js
```diff
@@ -337,6 +338,25 @@ define(function (require, exports, module) {
             context[helper] = registry_utils[helper];
         });
 
+        // Do some extra validation on homepage url to make sure we don't end up executing local binary
+        if (context.metadata.homepage) {
+            var parsed = PathUtils.parseUrl(context.metadata.homepage);
+
+            // We can't rely on path-utils because of known problems with protocol identification
+            // Falling back to Browsers protocol identification mechanism
+            var tmpLink = document.createElement('a');
```
nethip: Do you want to create an 'a' element everytime. You can resuse it? The GC of V8 is memory hungry, so will never know when this will be GC'ed.

nethip:
Travis is failing because of the API limitation. I have manually run `grunt build` and the build seems to be running fine. LGTM.

Issue #8208, opened 2014-06-20
Expand/Collapse all can cause some files in list to do the opposite


peterflynn:
1. Do a Replace In Files...
2. Click to collapse one item (optionally collapse a couple more items individually too)
3. Ctrl/Cmd+click that item (should mean "expand all")

UPDATE: this only happens for "Replace in Files...", not "Find in Files...", so updated recipe.

Result: that item expands but all the ones you didn't touch collapse instead. Any time you do an expand/collapse all, those items will now be inverted relative to the individually-touched ones.

PatrickDallarosa:
Looking for a simple way to contribute for a first timer. Providing the unit testing seems simple enough. Can I undertake this bug? Any guidance would be appreciated.

Denisov21:
PR Added unit testing for issue [#12390](https://github.com/adobe/brackets/pull/12390)

PR #12390, opened 2016-04-29
Added unit testing for issue #8208

ghost:
Created unit testing for collapsing and expanding item list after a find replace. Created a new describe block under describe("Replace") describe("UI"). The bug was stated as a replace bug exclusively. Added a new section to "UI", "Disclosure Arrows". Presumably, more extensive unit testing could be done, but I believe this covers the issue.


nethip:
Thanks @PatrickDallarosa LGTM

PR #12436, opened 2016-05-19
Fixed require of PathUtils in ExtensionManagerView

ficristo:
There are some ExtensionManager tests that fail because of this.


ficristo:
Closing in favor of https://github.com/adobe/brackets/pull/12437 which is a more complete fix.

Issue #12281, opened 2016-03-08
regular expression search does not work across lines


dakaraphi:
An expression such as 'sometext[\s\S]{0,200}othertext' can be used to find two text items that are within 200 characters of each other across lines; however, this does not work in Brackets.

petetnt:
Hi @dakaraphi,

there's issue #11588 which asks for similar feature but no-one is implementing it at the moment (AFAIK). If you would try to take an stab on it the community will help you out for sure :+1:

dakaraphi:
@petetnt I will see if I can take a look.  Do you have any pointers for what code to start looking at?

petetnt:
@dakaraphi https://github.com/adobe/brackets/blob/master/src/search/FindInFilesUI.js#L184 Would be  a good starting place.

dakaraphi:
I'm making progress on this issue.
The reason multi-line regular expressions currently do not work is that code mirror's search cursor only does a line by line regex match.
I'm working on a new search cursor implementation that can execute a regular expression against the entire document.
So far it is looking good and may improve performance as well.

bananacat2015:
Thank you so much for your effort! I started the issue #11588 about this before. Sorry that I cannot help more with the issue beyond cheering you on. I really think that being able to find/replace multiple lines will make brackets amazing. I really really appreciate your work towards a resolution, and I will check back.

PR #12442, opened 2016-05-21
Multiline regular expression selection support

d-akara:
#12281

This change is a new implementation of the search cursor which can support multi line regular expression selections.
**Notable changes:**
1. Multi line selections are now supported
2. The search cursor scans the entire document at once vs. the old implementation which was line by line.
3. The new search cursor can be used by other extensions that wish to take advantage of the performance of using the internal document line index.
4. Performance improvements
   1. A copy of the entire document is no longer created for every key press when typing in the search box.
   2. Next and previous navigation is much faster as it now just uses the cached locations.
   3. Performance is no longer strongly related to document size.  It is now more related to the number of matches.
   4. For comparison, previous implementation would take 6 sec for 3 million matches.  The new implementation can find 3 million matches in less than 300 ms.
   5. Increased the highlight limit to 10000 from 2000 as it seems we can now hightlight a greater number within the same time.
5. Considerations
   1. Although performance is faster in most use cases, there is a trade off for very large numbers of matches.  So with a very large document, using "." to match everything can create a pause on typing.  We might want to limit total matches or require more than "." to be entered before starting search.


ficristo:
Wow, number are really promising!
Ooc, have you looked at improving the search cursor implementation directly in CodeMirror?

dakaraphi:
It may be possible to improve CodeMirror's cursor in a compatible way; however, it will be more work to make it compatible with CodeMirror's use cases.

I made the new search cursor work the same as the Brackets search in all files, in that it converts plain string queries to regular expressions for consistency and allows use of same code path for strings and RegExp's.

CodeMirror will split multiline string queries at '\n' and has some special handling for those cases which also includes some special handling for code folding when it does splitting.
Also, the consideration I mentioned above about performance for extremely large matches could affect CodeMirror.  CodeMirror's search cursor does not expose any feature to limit the number of matches.  This might prove necessary for that performance use case and would affect other consumers of CodeMirror's search cursor requiring possible changes on their part.

Denisov21:
Great job @dakaraphi !

nethip:
Awesome!

src/search/FindReplace.js
```diff
@@ -34,31 +34,38 @@
...
+        BracketsSearchCursor = require("search/BracketsSearchCursor"),
+        _                    = require("thirdparty/lodash"),
+        CodeMirror           = require("thirdparty/CodeMirror/lib/codemirror");
 
     /**
      * Maximum file size to search within (in chars)
      * @const {number}
      */
-    var FIND_MAX_FILE_SIZE  = 500000;
+    var FIND_MAX_FILE_SIZE  = 16000000;
```
busykai: >15MB does seem as too high to me (given Brackets is an editor for HTML5 files). Is it just because the size limitation is no longer particularly limiting performance? Or it's taken from some real-life scenarios?

busykai:
@dakaraphi: an idea for your performance concern with all-inclusive regexps. For files larger than a certain size, we can take a small window of text (e.g. pos+2KB) and scan only it for the matches until "match density" (in my example it would be 2KB divided by number of matches) reaches a "reasonable" number as the user types in. Once it does, entire document is scanned on each regular expression edit by the user. If the file itself is smaller than, say 100KB, it's just scanned entirely.

dakaraphi:
@busykai I think the size isn't strongly correlated to performance as it was before.  That being said, some others should give it a try as my machine may not represent typical performance.  If the following idea on something like match density works out well, then size will probably be irrelevant for the most part.

I was thinking about something similar to your idea of match density.  I will look into it further and see how it performs.

dakaraphi:
@busykai Made numerous changes to address large documents:

**Notable changes:**

1.  Removed limit for document size.  We are now able to effectively search anything Brackets can open.
1.  Set default match limit to 1 million.  I looked at possible progressive algorithms as mentioned before, but 1 million is still responsive and seems more than reasonable as it is far past the capabilities of the other Node built desktop editors at the moment.
1.  Highlighting now enabled up to 1 million matches.  This is now possible due to dynamic highlighting.   For large numbers of matches, we now dynamically highlight the viewport area on scroll.
1.  Adding scroll ticks could cause some typing lag when there were many.  This is now asynchronous and debounced for more responsive typing in the search field.
1.  For large documents, we additionally debounce some search tasks for more responsive feel.
1.  Fast visual feedback in editor.  The initial visual feedback of highlighting is now virtually the same speed regardless of the size of document.  This is done by initially highlighting any matches that are in the viewport only before performing the full search.

Note, I performed most of these tests on a 14MB document.  It is noteworthy that other editors either can't perform or even crash when attempting searches on the same scale.

nethip:
@busykai Did you get a chance to review this change? This would be a great addition to Brackets!

busykai:
@dakaraphi, @nethip actually, I started to look at it this weekend. I'll be looking at it throughout this week. The feel with the change is just awesome though!

busykai:
@dakaraphi, there are a couple of corner cases where I don't feel like the UI is intuitive:

1. When searching for `$\n`:

screen shot 2016-07-21 at 16 30 48

Note that there are no search highlighting for the match itself, but all the scrollbar is ticked.

2. When everything matches (e.g. `.`):

screen shot 2016-07-21 at 16 31 18

Note that each match is highlighted, but no ticks on the scrollbar. I understand the difference is due to `FIND_HIGHLIGHT_MAX` and `FIND_SCROLLTICK_MAX` are treated as mutually exclusive, but they should not be, probably.

I also got few general comments on the function documentation:

1. In many places the function documentation uses verbs in the imperative as opposed to singular 3rd person (i.e. `Create` instead of `Creates`). Function documentation should start with a verb conjugated in 3rd person singular.

2. Template placeholders for argument types are left over in many places.

3. Argument description is omitted.

4. Sentences and paragraphs should start with a capital letter.

I'll leave these out from the inline comments for now, just to be less verbose. Please make a re-pass over the docs.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,746 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE
+ SOFTWARE.
```
busykai: Extra newline.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,746 @@
...
+/*global define, WeakMap, Uint32Array */
+
+define(function (require, exports, module) {
+    "use strict";
+    var StringUtils = require("utils/StringUtils");
+    var _           = require("thirdparty/lodash");
+
+    // store document text and index
+    // key: CodeMirror.Doc
+    // value: {text, index, generation}
```
busykai: this is a confusing description. something like `value: { text: {string}, index: {Object<number, Object>}, generation: {number} }`.

even though we don't use closure compiler to type check, we do use [the notation](https://developers.google.com/closure/compiler/docs/js-for-compiler).

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,746 @@
...
+     * determine if the current document has changed since we last stored the docInfo
+     * @param {CodeMirror.Doc} doc
+     * @return boolean
+     */
+    function _needToIndexDocument(doc) {
+        var docInfo = _documentMap.get(doc);
+
+        // lastmodtime is not changed when undo is invoked.
+        // so we will use the generation count to determine if the document has changed
+        if ((docInfo) && (docInfo.generation === doc.history.generation)) {
```
busykai: no need for parenthesis around `docInfo`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,746 @@
...
+ *
+ */
+
+/*jslint vars: true, plusplus: true, devel: true, nomen: true, indent: 4, maxerr: 50 */
+/*global define, WeakMap, Uint32Array */
+
+define(function (require, exports, module) {
+    "use strict";
+    var StringUtils = require("utils/StringUtils");
+    var _           = require("thirdparty/lodash");
```
busykai: The module `requires` should be joined into a single `var`.

dakaraphi:
@busykai 

1. I'll look into the case for `$\n`
1. I believe the highlight and tick behavior have not changed.  Just the amount of highlights available.  We previously would drop the tick marks when the number of highlights were too many.  I'm happy to try something different, but I don't know what that should be.  The main issues are:
   1. I'm not sure if more than 2000 tickmarks is meaningful.  The bar just becomes saturated.
   1.  The tickmarks cause a significant performance hit.  Being able to handle more would require addressing the performance first if that's what we want.
1.  I'll work on making the recommended doc changes.

dakaraphi:
@busykai 

1. The issue for `\n` is that codemirror's markText does not support marking regions without actual text.  I searched around a bit, but there appears no obvious solution.  Ideally, markText would support this as a property flag.
1. I think I now have a way of possibly supporting tickmarks efficiently for any number.  Hopefully will be able to try this out soon.

busykai:
@dakaraphi, thanks! I assumed CM did not support the newline selection, but since I'm not an expert on it, I was not sure. As for 2), I have a lots of code yet in this PR yet to be reviewed and I assume the improvement you mentioned will be somewhat localized. I'll continue with the review and will try your change out as soon as you push it.

dakaraphi:
@busykai 

1. In previous commit, I had inadvertently commented out the function call which does the early render in the viewport.  I fixed this, so for large documents the initial highlighting should be faster for what is currently in the viewport.
1. Tick marks no longer have a limit on the amount we can display.  I implemented something similar to video games that allow a LOD (level of detail) setting.
    1.  The current LOD is set to 1000.  So you can have up to 1000 physically rendered tickmarks.
    1.  If there are more than 1000 actual line matches, the tickmarks are representative of multiple lines.  This allows the tickmarks to still display a visual pattern of matches regardless of the number.
1.  Documentation is still work in progress.  I made some updates.  Although, I'm not exactly sure I have the syntax correct in some cases.  Will follow up later with more detail questions.

src/search/FindReplace.js
```diff
@@ -494,6 +576,82 @@ define(function (require, exports, module) {
...
+                findBar.showFindCount(Strings.FIND_NO_RESULTS);
+            }
+
+            state.foundAny = (state.searchCursor.getMatchCount() > 0);
+            indicateHasMatches(state, state.searchCursor.getMatchCount());
+
+        });
+    }
+
+    var performSearchUsingCursorDebounced = _.debounce(performSearchUsingCursor, 0);
```
busykai: @dakaraphi did you have any specific rationale in mind for debounce timeouts? I thought of it as the following: given average speed of typing is ~200 characters per minute, it yields about 300ms per character. I instrumented the code to measure and I see a lots of debounce ops actually being cancelled as I type without visually noticeable impact if this timeout is increased to 300. Whereas with 0 all the searched never have a chance of being cancelled.

At any rate, I am interested to learn what was your reasoning for choosing these specific delays. Also, in other debounced calls (drawing ticks and such), did you try to "pipeline" the operations by choosing growing delays?

busykai:
@dakaraphi thanks for addressing the issues, I will do my best to do another pass over the weekend. One general question I had: do we really need to keep two flows for "small" and "large" documents, in your opinion? Wouldn't it make the code simpler if we just had single (using debouncing for all the cases) approach?

dakaraphi:
@busykai I probably type quite a bit faster than 200 characters per minute.  I generally was thinking that the faster you type, the more noticeable and annoying would be any lagging feel.

So I chose the current values through a series of simply trying it and settling on values that seem to be a good compromise between feel of typing and visual responsiveness.

I'm not exactly sure what you are describing in 'growing delays'.

I considered having only one flow for "small" and "large" and I am still open to that as I agree I would normally prefer keeping the code simpler.  However, I also didn't want to regress in any noticeable way of the responsive feel.  It is subtle, but there is faster visual feedback.  Also the same is true for the dynamic highlighting, would prefer doing it only one way; however, for small number of matches if you scroll very fast the static highlighting will look better.

busykai:
> I'm not exactly sure what you are describing in 'growing delays'.

You have longer debouncing delays on some operations (0 for performing the search and 300 ms for ticks, for example). I was wondering what you were trying to accomplish with the difference as opposed to debouncing them all to a fixed small interval.

dakaraphi:
@busykai So the timings are different in that each operation seems to have a different impact.  For example, ticks are generally expensive.  So I went with having this occur less often by having a higher time.

The debouncing of search of 0 of large documents is to allow the search to start as soon as possible after the highlighting of the viewport.  Without debouncing the search in this flow, the early highlighting of the viewport will not occur as it waits for the search of the full document to also occur.  So this is done so that the event queue empties allowing the viewport highlighting to be applied first then perform the search of the whole document.

busykai:
@dakaraphi 

> Without debouncing the search in this flow, the early highlighting of the viewport will not occur as it waits for the search of the full document to also occur.

This is why asked. It does sound that the definition of "expensive" is relative. If search causes visual impact, then independently of when it's happen (synchronously or immediately after), it will take the same amount of time. So why not give it a chance to cancel: the next character put in by the user will render the results of the operation useless. Which gets us back to this point:

>  I probably type quite a bit faster than 200 characters per minute.  I generally was thinking that the faster you type, the more noticeable and annoying would be any lagging feel.

I was not talking about your typing speed, but average (taken from google search "average typing speed"). I assume a programmer would type slightly faster than that. However, that's not the point. The point is: we need to give a chance for a search to cancel. If it has 0 debounce, it will start anyways and as the user types, the time will be spent doing all these searches (most of which will be useless) over the entire document. I think a 200ms bounce is a reasonable enough starting point.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        _           = require("thirdparty/lodash");
+
+    /**
+     * Store document text and index
+     * key: {CodeMirror.Doc}
+     * value: { text: {string}, index: {Array<number>}, generation: {number} }
+     * text = The text of the document
+     * index = The document line index for lookups
+     * generation = the document history generation number
+     */
```
busykai: Add `@private`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+    var _documentMap = new WeakMap();
+
+    /**
+     * Determines if the current document has changed since we last stored the docInfo
+     * @param {CodeMirror.Doc} doc
+     * @return boolean
+     */
```
busykai: Add `@private`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+    function _needToIndexDocument(doc) {
+        var docInfo = _documentMap.get(doc);
+
+        // lastmodtime is not changed when undo is invoked.
```
busykai: `generation` is not really last modification time.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        return true;
+    }
+
+    /**
+     * Creates an array which stores the sum of all characters in the document
+     * up to the point of each line.
+     * This is needed to efficiently convert character index offsets to position objects of line and character offset.
+     * @param {String} text The string to index
+     * @param {String} lineSeparator The ending character that splits lines
+     */
```
busykai: Add `@private`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+    function _createLineCharacterCountIndex(text, lineSeparator) {
```
busykai: Referring to the same index by different names (`index`, `lineCharacterCountIndex`, `getDocumentIndex`) is a little bit confusing. What the index in question really is, it is the offset, if characters, of the line number `n` in the document. So two points: 1. We need to use a single name consistently; 2. I suggest it could be `lineOffsets`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        for (lineNumber = 0; lineNumber < lineCount; lineNumber++) {
+            totalCharacterCount += lines[lineNumber].length + lineSeparatorLength;
+            lineCharacterCountIndex[lineNumber] = totalCharacterCount;
+        }
+        return lineCharacterCountIndex;
+    }
+
+    /**
+     * Creates the document index and store in our map
+     * @param {CodeMirror.Doc} doc The codemirror document
```
busykai: Add `@private`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        var docText = doc.getValue();
+        var docLineIndex = _createLineCharacterCountIndex(docText, doc.lineSeparator());
+        _documentMap.set(doc, {text: docText, index: docLineIndex, generation: doc.history.generation});
+    }
+
+    /**
+     * Gets the document index
+     * @param {CodeMirror.Doc} doc The codemirror document
+     * @returns {Array<number>} see '_createLineCharacterCountIndex' for contents of Array
+     */
```
busykai: Add `@private`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        return _documentMap.get(doc).text;
+    }
+
+
+    /**
+     * Converts plain text query into regular expression
+     * If already regular expression, then just set the flags as appropriate
+     * @param {Object}  stringOrRegex A string or regular expression
+     * @param {boolean} ignoreCase True to ignore case for searchers
+     */
```
busykai: Add `@private`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+    function _convertToRegularExpression(stringOrRegex, ignoreCase) {
```
busykai: Maybe just `_makeRegExp`?

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+
+    /**
+     * Finds the line number for the given index
+     * @private
+     * @param {Array<number>} lineCharacterCountIndexArray See '_createLineCharacterCountIndex'
+     * @param {number}        startSearchingWithLine       Line number to start search
+     * @param {number}        indexWithinDoc               The index of the character offset from the start of the document
+     * @returns {number}      The line number for the given index.
+     */
+    function _lineFromIndex(lineCharacterCountIndexArray, startSearchingWithLine, indexWithinDoc) {
```
busykai: `lineStart` could be a more concise name than `startSearchingWithLine`, also `charStart` instead of `indexWithinDoc`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+     */
+    function _getDocumentIndex(doc) {
+        return _documentMap.get(doc).index;
+    }
+
+    /**
+     * Gets the document text
+     * @param {CodeMirror.Doc} doc The codemirror document
+     * @returns {String} document text content
+     */
```
busykai: Add `@private`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+    function _indexDocument(doc) {
+        var docText = doc.getValue();
+        var docLineIndex = _createLineCharacterCountIndex(docText, doc.lineSeparator());
+        _documentMap.set(doc, {text: docText, index: docLineIndex, generation: doc.history.generation});
+    }
+
+    /**
+     * Gets the document index
+     * @param {CodeMirror.Doc} doc The codemirror document
+     * @returns {Array<number>} see '_createLineCharacterCountIndex' for contents of Array
```
busykai: Typo: `@returns` = > `@return`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+     */
+    function _getDocumentIndex(doc) {
+        return _documentMap.get(doc).index;
+    }
+
+    /**
+     * Gets the document text
+     * @param {CodeMirror.Doc} doc The codemirror document
+     * @returns {String} document text content
```
busykai: Typo: `@returns` => `@return`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+
+    /**
+     * Given the character offset from the beginning of the document
+     * creates an object which has the position information
+     * @param {Array<number>} lineCharacterCountIndexArray See '_createLineCharacterCountIndex'
+     * @param {number} startSearchingWithLine Line number to start search
+     * @param {number} indexWithinDoc The index of the character offset from the start of the document
+     * @return {{line: number, ch: number}} Line and character offsets
+     */
+    function _createPosFromIndex(lineCharacterCountIndexArray, startSearchingWithLine, indexWithinDoc) {
```
busykai: Same comment on var naming applies as for `_lineFromIndex()` function.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        return {line: lineNumber, ch: indexWithinDoc - previousLineEndingCharacterIndex };
+    }
+
+    /**
+     * Returns the character offset from the beginning of the document based on
+     * object properties as pos.from.line and pos.from.ch
+     * where line is the line number in the document and ch is the character offset on the line
+     * @param {Array<number>} lineCharacterCountIndexArray See '_createLineCharacterCountIndex'
+     * @param {{line: number, ch: number}} pos Object describing the position within the document
+     */
```
busykai: Document `@return`. Add `@private`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        }
+        // Add the number of characters offset from the start and return
+        return indexAtStartOfLine + pos.ch;
+    }
+
+    /**
+     * Scans entire document and callback with each match found.
+     * Uses the documentIndex to more efficiently create the position objects on found matches.
+     *
+     * @param {Array<number>} lineCharacterCountIndexArray See '_createLineCharacterCountIndex'
```
busykai: `@param` is inconsistent with the actual argument name.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+     * @param {String}        documentText                 Text to scan
+     * @param {RegExp}        regex                        Regular expression used to search
+     * @param {{from: {line: number, ch: number}, to: {line: number, ch: number}}} range Area to scan for matches
+     * @param {function({line: number, ch: number}, {line: number, ch: number}, Array )} fnEachMatch Function is called with start position, end position and Regex match Array
+     */
+    function _scanDocumentUsingRegularExpression(documentIndex, documentText, regex, range, fnEachMatch) {
```
busykai: Name `_scanDocumentUsingRegularExpression()` is excessive, it seems. Just `_scanDocument()` would do just fine since it's the only scanning function in the module.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        var toPos   = _createPosFromIndex(docLineIndex, fromPos.line, indexEnd);
+
+        return {from: fromPos, to: toPos};
+    }
+
+    /**
+     * Comparison function for binary search of index positions within document.
+     * @param {number} matchIndex First match index for compare
+     * @param {number} posIndex   Second match index for compare
+     * @returns {number} Result of comparison
```
busykai: `@return`.

busykai: A better documentation for return value: "negative, zero or positive value if the first argument is less than, equal to or greater than the second argument, correspondingly."

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+     */
+    function _compareMatchResultToPos(matchIndex, posIndex) {
```
busykai: Names `matchIndex` and `posIndex` are irrelevant. So if the long name of the function. Since you separate and declare comparator, it should be just `_compareOffsets()`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        if (matchIndex === posIndex) {
+            return 0;
+        } else if (matchIndex < posIndex) {
+            return -1;
+        } else {
+            return 1;
+        }
+    }
```
busykai: A more efficient implementation would be to `return index1 - index2;` and compare if its 0, positive or negative in the caller (instead of equality to `1`, `0`, or `-1`). Or if the caller compares using `switch`, then use `return Math.sign(index1 - index2);`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        var compare;
+
+        var length = regexIndexer.getItemCount();
+        var upperBound = length - 1;
+        var lowerBound = 0;
+        var searchIndex;
+        while (lowerBound <= upperBound) {
+            searchIndex = Math.floor((upperBound + lowerBound) / 2);
+            compare = fnCompare(regexIndexer.getMatchIndexStart(searchIndex), pos);
+            if (compare === 0) {
```
busykai: Use `switch` (more readable).

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+     * Finds the result that is at or nearest the position passed to function.
+     * If a match result is not at the position, it will then locate the closest
+     * match result which is in the search direction.
+     * If there is no match found before the end or beginning of the document
+     * then this function returns false.
+     * @param {!Object} regexIndexer instance of the regex indexer. see _createRegexIndexer
+     * @param {number} pos Starting index position to search from
+     * @param {boolean} reverse direction to search.
+     * @param {function(number, number)} fnCompare function to compare positions for binary search
+     */
```
busykai: Document return value.

busykai: Add `@private`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+    }
+
+    /**
+     * Enhances array with functions which facilitate managing the array contents
+     * by groups of items.
+     * This is useful for both performance and memory consumption to store the indexes
+     * of the match result beginning and ending locations.
+     * @param {Array} array The array to enhance
+     * @param {number} groupSize The number of indices that belong to a group
+     * @returns {GroupArray} Enhanced Array
```
busykai: `@return`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+     * Performs a search using the supplied RegExp and adds all results
+     * of matched locations to the group array
+     *
+     * @private
+     * @param   {object}     query                           regular expression query
+     * @param   {String}     docText                         the text to search
+     * @param   {GroupArray} groupArray                      group array to hold values of index locations
+     * @param   {number}     [matchCountLimit=10000000]      search will stop when match limit met
+     * @param   {number}     [searchEndIndex=docText.length] search will stop when last match exceeds index location
+     * @returns {number}                                     Count of results
```
busykai: `@return`.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+    }
+
+    /**
+     * Determines if the values of specified items in a group array are equal
+     * @private
+     * @param   {GroupArray} groupArray1 First group array
+     * @param   {GroupArray} groupNum1   Item within group array
+     * @param   {GroupArray} groupArray2 Second group array
+     * @param   {GroupArray} groupNum2   Item within 2nd group array
+     * @returns {boolean}  true when values are equal
```
busykai: `@return`.

busykai:
General comment: we use [closure jsdoc notation](https://developers.google.com/closure/compiler/docs/js-for-compiler). In it, string datatype is `string` (lowercase). It seems to be needing correction in all occurrences.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+/*jslint vars: true, plusplus: true, devel: true, nomen: true, indent: 4, maxerr: 50 */
+/*global define, WeakMap, Uint32Array, Uint8Array */
+
```
busykai: This entire file merits a general description of what it does, what the used datatypes are, and what the general flow. Here is [an example](https://github.com/adobe/brackets/blob/master/src/filesystem/FileSystem.js#L27-L91) of how such a description may look like. It should prepare a developer to navigate the implementation.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+     * by groups of items.
+     * This is useful for both performance and memory consumption to store the indexes
+     * of the match result beginning and ending locations.
+     * @param {Array} array The array to enhance
+     * @param {number} groupSize The number of indices that belong to a group
+     * @returns {GroupArray} Enhanced Array
+     */
+    function _makeGroupArray(array, groupSize) {
+        var _currentGroupIndex = -groupSize;
+        _.assign(array, {
```
busykai: May be it could be a better implementation if `GroupArray` was a real datatype (function) which would contain its internal state. The same is true for the "indexer" and the cursor objects.

Instead of using `_make...` functions and `_.assign` or `_.extend`. You'll have a constructor and use `new`.

For example, for `GroupArray` you'd use:

```javascript
function GroupArray(array, groupSize) {
    this._currentGroupIndex = -groupSize;
    this._size = groupSize;
    this._array = array;
}

GroupArray.prototype = Object.create(null); // do not inherit from object
GroupArray.prorotype.constructor = GroupArray;

GroupArray.prototype.nextGroupIndex = function () {
    // your implementation referring to this instead of make function context...
};
```

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+    /**
+     * Determines if the values of specified items in a group array are equal
+     * @private
+     * @param   {GroupArray} groupArray1 First group array
+     * @param   {GroupArray} groupNum1   Item within group array
+     * @param   {GroupArray} groupArray2 Second group array
+     * @param   {GroupArray} groupNum2   Item within 2nd group array
+     * @returns {boolean}  true when values are equal
+     */
+    function _isGroupValuesEqual(groupArray1, groupNum1, groupArray2, groupNum2) {
```
busykai: This should be a member function of the type.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        if (_.isEqual(groupArray1.getGroupValue(groupNum1, 0), groupArray2.getGroupValue(groupNum2, 0)) &&
+                _.isEqual(groupArray1.getGroupValue(groupNum1, 1), groupArray2.getGroupValue(groupNum2, 1))) {
```
busykai: Either group size should not be a variable (i.e. always assume 2) or this implementation should be made size-independent (again, easy to do if equals belongs to the type).

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+     * Creates the regex indexer which finds all matches within supplied text using the search query.
+     * Uses a lookup index to efficiently map regular expression result indexes to position used by Brackets
+     * @param {String} docText       The text to search for matches
+     * @param {Array}  docLineIndex  Array used to map indexes to positions
+     * @param {RegExp} query         A regular expression used to find matches
+     * @param {number} maxResults    The limit of the number of matches to perform
+     * @param {{to: {line: number, ch: number}, from: {line: number, ch: number}} startPosition Start searching at this position
+     * @returns {RegexIndexer} A new instance of the regular expression indexer
+     */
+    function _createRegexIndexer(docText, docLineIndex, query, maxResults, startPosition) {
```
busykai: Would a name `SearchResults` would be more appropriate for this datatype? Judging by the interface, it would be a more intuitive name for it.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,823 @@
...
+        });
+    }
+
+    /**
+     * Creates an updatable search cursor which can be used to navigate forward and backward through the results.
+     *
+     * @param {!{document: CodeMirror.Doc, searchQuery: string|RegExp, position: {line: number, ch: number}, ignoreCase: boolean}} properties
+     * @return {Object} The search cursor object
+     */
+    function createSearchCursor(properties) {
```
busykai: Why passing `properties` instead of named arguments? It's understandable while in development (while exact list of arguments is not defined), but not in its final form.

busykai:
@dakaraphi, I think I'm done with the first pass. I strongly recommend refactoring the datatypes in `BracketsSearchCursor.js` as per my comments. We used this pattern elsewhere (as opposed to lodash utilities and anonymous types).

nethip:
@dakaraphi Just wanted to check if you got a chance to work on the review comments.

dakaraphi:
@nethip Started, but have had much less time available recently.  Hopeful to make progress this weekend.

nethip:
@dakaraphi No problem! Take your time.

dakaraphi:
@busykai Next changes ready for review.
BTW, after merging with latest on master, I can no longer run any unit tests.  The window just hangs.  Not sure what is the issue.

zaggino:
@dakaraphi for the test issues, try running `npm install` in `brackets` folder, if that doesn't help try getting latest `brackets-shell` master and rebuilding it

adityapaturi:
@dakaraphi : The match case find is not working as expected. It is always returning the results as if match case button is not pressed.

nethip:
@adityapaturi It would be great if you can point out the piece of code where the case matching is not working.

busykai:
@dakaraphi, I'll review this by the end of the week. Thank you for taking your time to address the review comments!

dakaraphi:
@adityapaturi This should now be fixed.  Thanks.

busykai (changes requested):
@dakaraphi thank you for making these changes, it's getting better! I think at this point the major concern is very few tests for the amount of the code added. I don't see multi-line searches properly tested, there's just one case which does not even use `\n` proper. Also, if you could add some tests for `GroupArray` and `SearchResults`, it'd be great. Though they are tested indirectly, some tests for these classes would be a great addition to the documentation.

Other general comments:
1. Please change @private to be the first tag in the function documentation, like this:

```
/**
 * @private
 * 
 * Gets a number
...
```

Right now @private is located in rather random positions.
2. I will try to come up with a better naming for functions/structures/variables, but it will take me some time. Instead, I would like to ask you to review the naming throughout the `BracketsSearchCursor.js` on your own and see if you could come up. In my opinion, the logic is hard to grasp because the use of terminology is so confusing. To make the code more sustainable, it would be very helpful to include an explanation of the semantics of GroupArray, how it is structured, what the groups are. In SearchResults, what's the different between a match and a result (if any) etc.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,848 @@
...
+     * Enhances array with functions which facilitate managing the array contents
+     * by groups of items.
+     * This is useful for both performance and memory consumption to store the indexes
+     * of the match result beginning and ending locations.
+     * @private
+     * @param {Array} array The array to enhance
+     * @param {number} groupSize The number of indices that belong to a group
+     * @return {GroupArray} Enhanced Array
+     */
+    function _makeGroupArray(array, groupSize) {
```
dakaraphi: This is factory function pattern.  It is usually considered a good thing, but I will change if you wish.

src/search/BracketsSearchCursor.js
```diff
@@ -0,0 +1,848 @@
...
+
+    /**
+     * Scans the entire document for regular expression matches and calls fnEachMatch for each found match.
+     * Unlike the search cursor, this creates no retained results to be used for back and forward navigation.
+     * This is provided for consumers who wish to leverage the speed provided by the document index, but do
+     * not need to use the features of a cursor.
+     *
+     * @param {!{document: CodeMirror.Doc, searchQuery: string|RegExp, ignoreCase: boolean, fnEachMatch: function({line: number, ch: number}, {line: number, ch: number, match: Array})}} properties
+     */
+    function scanDocumentForMatches(properties) {
```
dakaraphi: I used options pattern. http://www.codereadability.com/what-are-javascript-options-objects/
Will still change if you wish.

src/search/BracketsSearchCursor.js
busykai: It would be OK if it created closures or had any logic in it. There's no need to wrap object creation into a factory function.

src/search/BracketsSearchCursor.js
busykai: I do not insist on this one, but I always found it rather less readable and more prone to errors. In the case of this specific example, a signature like this would make sense to me:
```javascript
function scanDocumentForMatches(document, query, options)
```
It makes it much more understandable as to what is needed for the function to complete successfully.

dakaraphi:
@busykai I'm unfortunately not going to be able to complete the remaining changes.  I have other shifting priorities to which I need to focus.
I started this initially thinking this would be a bug fix that turned into a small project.  I've put more time into than planned.  I apologize for anyone waiting on this.
If anyone else wants to pickup the work, I will assist explaining how this works as much as I can.

sobisht:
Added more documentation, removed extra wrappers and added more unit tests to address the review comments.

Ping @swmitra @busykai

PR #12445, opened 2016-05-22
Make file stats available to indexFilter

d-akara:
Small change to make the file stat information available to the file filter.

There are extensions available to extend the filtering capability, but they are all limited in that it is difficult to have custom filtering for directories or file sizes etc without more information provided to the filter.

### DIFF
--- a/src/filesystem/Directory.js
+++ b/src/filesystem/Directory.js
@@ -170,9 +170,9 @@ define(function (require, exports, module) {
                 names.forEach(function (name, index) {
                     var entryPath = this.fullPath + name;
 
-                    if (this._fileSystem._indexFilter(entryPath, name)) {
-                        var entryStats = stats[index],
-                            entry;
+                    var entryStats = stats[index];
+                    if (this._fileSystem._indexFilter(entryPath, name, entryStats)) {
+                        var entry;
 
                         // Note: not all entries necessarily have associated stats.
                         if (typeof entryStats === "string") {


ficristo:
IMHO, this raises the question if `_indexFilter` should be a public api.
Improving it only for the sake of extensions is seems to me a bit risky...

dakaraphi:
I agree about there should be discussion on if `_indexFilter` should be public.
I'm not sure I understand the risk statement.  Maybe you could elaborate further to help me understand the concern?

ficristo:
The _risk_ is only from a maintanaince point of view.

zaggino:
@ficristo 
I don't think this PR makes `_indexFilter` public. It's an underscored method which hints `private` and it's up to developers not to abuse it, this PR doesn't change that. From maintenance point of view, I'm sure @dakaraphi understands, that arguments passed to this function can change with any Brackets release and break his extensions. Voting for merge here.

ficristo:
If the only motivation behind this change is for extensions sake, I'm against it.
IMHO we should spend 0 time on improving private API for that solely reason.
And if we start with one we create a precedent.
We should instead discuss if this kind of API should be made public and eventually add the necessary parameters.
That said, we can hear more opinions of others committers and decide what to do.

dakaraphi:
I just need a solution to provide this capability.  All extensions which do filtering currently use this function.  In order to get this functionality I must do a very ugly monkey patch in my extension.

Brackets is not usable in my project without this capability as it is a very complex directory structure with many hundred thousands of files.  This certainly can stay private if Brackets wants to include such functionality by default.

For reference, my extension: https://github.com/dakaraphi/brackets-filter-extension

ficristo:
If we are talking about filtering extensions, personally I always felt they were an hack done because of the limitations of Brackets.
The right fix is to provide these features natively in Brackets and not though an extension.
And I think, with the latest changes of @zaggino, we are almost there.

zaggino:
@dakaraphi in the meantime, you can do this:
```
FileSystem.prototype._indexFilter = function (path, name) {
        
        var stats = this.getFileForPath(path)._stat;
        if (stats) {
            ...
        }
```
This would require probably calling `projectRefresh` event after it's first loaded because first run of this won't have stats.

ficristo:
I think that before we have something in the tree we will have to wait still a lot of time.
I still believe this is not right, but meh. We will survive with it.
@zaggino could you check if this is still valid and land?

@dakaraphi keep in mind that is not a supported behaviour and if this will be in the way we will remove it.

zaggino:
👍

Issue #12456, opened 2016-05-25
Code folding triangles don't update on unfold when block above them is removed


jwestbrook:
Scenario : 

In a PHP file with mixed HTML and PHP I'm folding the `<tr>`s of a table. If I have a set of 3 rows and I unfold row 2 which has 2 PHP blocks within it, and remove one of the blocks, then go to row 3 to unfold - the row will unfold but the triangle will not update and I think the folded state will not save.

For example

Step 1
```html
<table>
<tr>...</tr>
<tr>...</tr>
<tr>...</tr>
</table>
```

Step 2
```html
<table>
<tr>...</tr>
<tr>
    <td>
       <?php
       if($variable)
       {
           foreach($variable as $v)
           {

           }
       }
       ?>
   </td>
</tr>
<tr>...</tr>
</table>
```

Step 3
```html
<table>
<tr>...</tr>
<tr>
    <td>
       <?php
       if($variable) //remove
       {                  //remove
           foreach($variable as $v)
           {

           }
       }             //remove
       ?>
   </td>
</tr>
<tr>...</tr>
</table>
```

Step 4
```html
<table>
<tr>...</tr>
<tr>
    <td>
       <?php
           foreach($variable as $v)
           {

           }
       ?>
   </td>
</tr>
<tr>...</tr>
</table>
```
At this point unfolding row 3 will show the code contained, but the triangle will not rotate and will not respond to any more clicks unless you reload Brackets.

OSX 10.11.5

Release 1.6 build 1.6.0-16680 (release 3df0ac6fa) 
build timestamp: Tue Jan 12 2016 06:18:55 GMT-0800

Don't know if its worked before

I've attempted this with and without extensions enabled.

No errors logged in console

thehogfather:
@jwestbrook thanks for this. I'm able to reproduce this with some small variations on your test.
I am fairly sure this is related to #12232

jwestbrook:
Yep that looks like the same behavior I get - glad I could illustrate a test case

ficristo:
Should be fixed in 1.8 by https://github.com/adobe/brackets/pull/12521

PR #12521, opened 2016-06-14
Addresses #12456 

thehogfather:
Addresses issue where code folding triangles in the gutter were sometimes not correctly updated after deleting lines above a folded region.


abose:
Tagging @swmitra

ficristo:
I've looked a bit at this.
I was able to reproduce the issue (trying to delete a line of code above a folded region) and indeed this PR fixes it.
@thehogfather can you add a test?

I've noticed a small issue: if collapse a folded region and then expand it, the code folding triangles are not updated until you move the cursor.
fold

I think this is a minimal fix to solve the issue, but the right fix seems to me that we should update all the code of the code folding addon of CodeMirror.

thehogfather:
@ficristo I will have a go at adding some more tests.

Could you clarify what you mean by 'update all the code of the code folding addon of CodeMirror'?

ficristo:
If I'm not mistaken, the `foldcode.js` and `foldgutter.js` are a port of the CodeMirror ones adapted to work in Brackets. I think we should backport all the changes that are in CodeMirror (better in another PR).

thehogfather:
Ah yes - better a different PR. There are lots of subtle differences, mainly to do with the ability to persist and restore fold states.

thehogfather:
Changes ready for another review.
Main changes are updates to tests which now include html files  as well as js.

src/extensions/default/CodeFolding/foldhelpers/foldcode.js
```diff
@@ -90,18 +90,24 @@ define(function (require, exports, module) {
...
-            CodeMirror.signal(cm, "unfold", cm, from, to, pos.line);
+            delete cm._lineFolds[from.line];
+            CodeMirror.signal(cm, "unfold", cm, from, to);
         });
 
         if (force === "fold") {
             delete range.cleared;
-            cm._lineFolds[pos.line] = range;
+            // in some cases such as in xml style files, the start of  line folds can span multiple lines
+            // render a gutter marker for both the beginning and end of the line
```
ficristo: Use capital letter at the start of a sentence. This needs some dots too.

src/extensions/default/CodeFolding/unittest-files/test.html
```diff
@@ -0,0 +1,38 @@
...
+                     Choice 1
+                </label>
+                 <input type="radio" name="radio-choice-1" id="radio-choice-1" tabindex="2" value="choice-1" />
+
+                 <label for="radio-choice-2">Choice 2</label>
+                 <input type="radio" name="radio-choice-2" id="radio-choice-2" tabindex="3" value="choice-2" />
+            </div>
+        </form>
+    </body>
+</html>
```
ficristo: Add a newline at the end of file please.
Are the empy lines necessary? If not remove.

src/extensions/default/CodeFolding/unittest-files/test.js
```diff
@@ -1,33 +1,36 @@
...
-            return;
-        }
+  * Synchronises the code folding states in the CM doc to cm._lineFolds cache.
+  * When an undo operation is done, if folded code fragments are restored, then
+  * we need to update cm._lineFolds with the fragments
+  * @param {Object}   cm        cm the CodeMirror instance for the active  editor
+  * @param {Object}   from      starting position in the doc to sync the fold states from
+  * @param {[[Type]]} lineAdded a number to show how many lines where added to the document
+*/
+/*global define, brackets, document, window, $, CodeMirror, isFold, prefs*/
```
ficristo: Put this as the first line followed by an empty line.
I think you can remove some of this too.
`isFold` and `prefs` cannot be required?

src/extensions/default/CodeFolding/unittest-files/test.js
```diff
@@ -1,33 +1,36 @@
...
+
+function syncDocToFoldsCache(cm, from, lineAdded) {
+    "use strict";
+    var minFoldSize = prefs.getSetting("minFoldSize") || 2;
+    var opts = cm.state.foldGutter.options || {};
+    var rf = opts.rangeFinder || CodeMirror.fold.auto;
+    var i, pos, folds, fold, range;
```
ficristo: Brackets style is to use a combined `var`.

src/extensions/default/CodeFolding/unittest-files/test.js
```diff
@@ -1,33 +1,36 @@
...
+    if (lineAdded <= 0) {
+        return;
+    }
```
ficristo: Can you put this directly aftrer `"use strict"`. I don't remember if JSLint complains.

src/extensions/default/CodeFolding/unittests.js
```diff
@@ -22,8 +22,29 @@ define(function (require, exports, module) {
             foldMarkerOpen = gutterName + "-open",
             foldMarkerClosed = gutterName + "-folded";
         var extensionPath = FileUtils.getNativeModuleDirectoryPath(module),
-            testDocPath = extensionPath + "/unittest-files/",
-            testFilePath = testDocPath + "test.js";
+            testDocumentDirectory = extensionPath + "/unittest-files/",
+            testFiles = ["test.js", "test.hmtml"].map(function (file) {
```
ficristo: I guess this should be `"test.html"`

src/extensions/default/CodeFolding/unittests.js
```diff
@@ -22,8 +22,29 @@ define(function (require, exports, module) {
...
+            // The line numbers referenced below are dependent on the files in /unittest-files directory.
+            // Remember to update the numbers if the files change.
+            testFilesSpec = {
+                js: {
+                    filePath: testDocumentDirectory + "test.js",
+                    foldableLines: [ 11, 17, 21, 25, 27, 30],
+                    sameLevelFoldableLines: [17, 21],
+                    firstSelection: {start: {line: 2, ch: 0}, end: {line: 10, ch: 0}},
+                    secondSelection: {start: {line: 5, ch: 0}, end: {line: 8, ch: 4}}
+
```
ficristo: Remove newline

ficristo:
@thehogfather thank you a lot!
Could you merge / rebase with master? We have updated CodeMirror, so I would like to test with the new version.
Meanwhile I left some comments.

src/extensions/default/CodeFolding/unittest-files/test.html
thehogfather: This is a file used in the test as opposed to a regular source file. I would argue that this file should not need to conform to the brackets style  - but I am open to good reasons not to. For instance the line break on line 27 is intentional - it is a special case of when the fold gutter appears on a line different from when the beginning of the range starts.

src/extensions/default/CodeFolding/unittest-files/test.js
thehogfather: Similar to the html, this is a test file loaded up during the unit tests. the `isFold` and `prefs` are just there to silence jshint errors.

src/extensions/default/CodeFolding/unittest-files/test.js
thehogfather: Similarly, I dont think this file should conform to the brackets style - they are test files and should be able to accommodate plausible scenarios that might appear in projects of anyone using brackets regardless of style.

src/extensions/default/CodeFolding/unittests.js
thehogfather: good catch. This must be a redundant line. Will remove.

ficristo:
If the newlines I mentioned are necessary as the globals to make testing easier, then is fine to leave them.

ficristo:
@thehogfather could you take a look at the test failures?

So now when a tag span on multiple lines you have two code folding triangles. I'm not an UI / UX expert so I'm not sure is actually a good or bad idea.
html-foldable

/cc @MarcelGerber

thehogfather:
@ficristo regarding the folding triangles, I think this is fine. A better design might be to hide the opening tag when the range is collapsed or to only show the fold triangles at the start of the folded/foldable range (so in your example we dont show line 25 as foldable - only line 29). This is however a little tricky - would need a rewrite of the xml-range finder. We are currently using whatever version is from code-mirror. The same workaround is present in code-mirror. I think it is ok as it doesn't reduce functionality nor does it put the editor in an unusable state.

redmunds:
Another option would be to collapse both tag attributes and tag contents something like this:

tag-collapse2

ficristo:
I still see some test failures on the latest commit.

thehogfather:
@ficristo would you kindly point out the failing test if possible? I typically run the extension test as opposed to 'All tests' so I might have missed something. I've just run the tests again and all are passing. Note that the last commits need a full reload of brackets so that the changes in code mirror are reflected in the tests.

thehogfather:
@redmunds interesting idea - I'd however question  what happens to the fold mark in the gutter when either of the collapsed ranges in the editor is clicked. They should probably be treated as two independent ranges (as displayed in the editor) and marked as such in the gutter with two fold triangles when open.

ficristo:
Since the behaviour match the CodeMirror one I'm going to take this as is and I'll open a new issue to discuss more the UI / UX.
I've run multiple times the tests and seems there are some issues with the focus of the test window: when I give the focus on it all the tests pass.

@thehogfather can you squash a bit the commits?

thehogfather:
@ficristo not sure how much more I can squash this. I have tried without much success. Maybe whoever merges on github can use the Squash and merge button? Alternatively I would probably create a new PR with all the changes if that is preferable.

ficristo:
@thehogfather thank you.
I've merged as is (I didn't want to mess with github)

PR #12606, opened 2016-07-20
Added processing as a recognized programing language

jschluger:
Added Processing ([https://processing.org/](url))as a recognized programing language so that it can automatically get syntax highlighting.

### DIFF
--- a/src/language/languages.json
+++ b/src/language/languages.json
@@ -160,6 +160,14 @@
         "blockComment": ["/*", "*/"],
         "lineComment": ["//"]
     },
+    
+    "processing": {
+        "name": "Processing",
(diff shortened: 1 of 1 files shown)


ficristo:
I'm not part of the Brackets team, but IMHO this change should be integrated as an extension.

abose:
Thanks @jschluger For contributing to brackets.
As a general guideline, brackets core is targetted towards web development, and hence we avoid languages not related to the web in core. Haskell - being an exception as we accidentally merged it in and forgot to rollback :sweat_smile: 
New language support is in the realm of extensions. That can bring in a much more complete support for the language.
Closing as extension idea.

PR #12654, opened 2016-08-06
add typescript and tsx definitions to language.json

zaggino:
note: there's currently no mime for typescript with jsx, that should be altered later ref: https://github.com/codemirror/CodeMirror/issues/3893

### DIFF
--- a/src/language/languages.json
+++ b/src/language/languages.json
@@ -78,6 +78,22 @@
         "lineComment": ["//"]
     },
 
+    "typescript": {
+        "name": "TypeScript",
+        "mode": ["javascript", "text/typescript"],
+        "fileExtensions": ["ts"],
+        "blockComment": ["/*", "*/"],
+        "lineComment": ["//"]
+    },
+
+    "tsx": {
+        "name": "TSX",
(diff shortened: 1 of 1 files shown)


ficristo:
I think you need to change the `htmlmixed` in `LanguageManager` to include the new modes and update the `SVGAttributes.json` and `HTMLAttribute.json`.

But I'm not entirely convinced it belongs to the core.
For example I tried the following TSX example and the html is not highlighted.
```tsx
declare namespace JSX {
    interface IntrinsicElements {
        foo: any
    }
}


declare namespace JSX JSX {
  interface ElementClass {
    render: any;
  }
}

class MyComponent {
  render() {}
}
function MyFactoryFunction() {
  return { render: () => {} }
}

<MyComponent />; // ok
<MyFactoryFunction />; // ok

class NotAValidComponent {}
function NotAValidFactoryFunction() {
  return {};
}
```

zaggino:
Hmm, thinking about it, maybe you're right. Putting it to core would mean people would expect code completion and stuff like that. Better leave it for the extension to define the language then.

PR #12671, opened 2016-08-12
Avoid the edit icon to move in the file exclusion set

ficristo:
This should fix or at least help with #10963

### DIFF
--- a/src/widgets/DropdownButton.js
+++ b/src/widgets/DropdownButton.js
@@ -264,14 +264,19 @@ define(function (require, exports, module) {
             posTop = Math.max(0, toggleOffset.top - $dropdown.height() - 4);
         }
 
+        // Take in consideration the scrollbar to prevent unexpected behaviours (see #10963).
+        var dropdownElement = this.$dropdown[0];
+        var scrollWidth = dropdownElement.offsetWidth - dropdownElement.clientWidth + 1;
+
         if (clip.right > 0) {
-            // Right is clipped, so adjust left to fit menu in editor
-            posLeft = Math.max(0, posLeft - clip.right);
+            // Right is clipped, so adjust left to fit menu in editor.
+            posLeft = Math.max(0, posLeft - clip.right - scrollWidth);
         }
 
         $dropdown.css({
             left: posLeft,
-            top: posTop
+            top: posTop,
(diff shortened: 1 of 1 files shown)


MarcelGerber:
I suppose the empty line isn't supposed to show up any more? For me, it still does...
image

Also, there are the related issues #9305 and #10962. Maybe you can find a fix that handles all 3 of them?

ficristo:
It seems to me you are resizing the window so it's #10962?
Unfortunately fixing #10962 is out for me, I fought enough against CSS with this...

MarcelGerber:
No, the window is at full size and has been at this size the whole time.

ficristo:
With latest changes I should have fixed your issue too. But still not #10962.

src/widgets/DropdownButton.js
```diff
@@ -264,14 +264,19 @@ define(function (require, exports, module) {
             posTop = Math.max(0, toggleOffset.top - $dropdown.height() - 4);
         }
 
+        // Take in consideration the scrollbar to prevent the edit icon to move down. See #10963
```
MarcelGerber: Maybe we shouldn't mention the edit icon here because DropdownButton is far more versatile.
I'd go for something along the lines of `Take the scrollbar into consideration so we don't get unintented line breaks (see #10963)`

MarcelGerber:
Thanks! LGTM now and handles both issues indeed :shipit: 
Just that tiny thing about the comment and we're ready to go!

src/widgets/DropdownButton.js
```diff
@@ -264,14 +264,19 @@ define(function (require, exports, module) {
...
+        var dropdownElement = this.$dropdown[0];
+        var scrollWidth = dropdownElement.offsetWidth - dropdownElement.clientWidth + 1;
```
MarcelGerber: Also, where does the `+ 1` come from? Is it just a safety measure so we have a tiny little more space available or does it serve a real purpose?

ficristo:
Actually that 1 pixel is the fix!
Only after found it i decided to look at scrollbar to fix #10963 too.

MarcelGerber:
Thank you!
Merging.

ficristo:
Thank you for all the reviews.

PR #12673, opened 2016-08-14
Ensures fold-gutter is always inserted after the line-number gutter (if it exists)

thehogfather:
Addresses issue where line numbers appears after fold gutter if the line number visibility is toggled in the View Menu.

Uses [MutationObserver](https://developer.mozilla.org/en/docs/Web/API/MutationObserver) to watch changes to the brackets code mirror gutters.

Was PR #11593
Addresses #11577, #10864

@ficristo I noticed that globals are no longer placed in files, I've used globals here for MutationObserver (since it isn't used anywhere else) - but open to moving to a more central location if needed.

### DIFF
--- a/src/extensions/default/CodeFolding/main.js
+++ b/src/extensions/default/CodeFolding/main.js
@@ -52,7 +52,12 @@ define(function (require, exports, module) {
         collapseAllKey          = "Alt-1",
         expandAllKey            = "Shift-Alt-1",
         collapseAllKeyMac       = "Cmd-1",
-        expandAllKeyMac         = "Cmd-Shift-1";
+        expandAllKeyMac         = "Cmd-Shift-1",
+        SAVE_FOLD_STATES        = "saveFoldStates",
+        HIDE_UNTIL_MOUSE_OVER   = "hideUntilMouseover",
+        MAKE_SELECTION_FOLDABLE = "makeSelectionFoldable",
+        ALWAYS_USE_INDENT_FOLD  = "alwaysUseIndentFold",
+        LINE_NUMBER_GUTTER      = "CodeMirror-linenumbers";
 
     ExtensionUtils.loadStyleSheet(module, "main.less");
 
@@ -69,11 +74,10 @@ define(function (require, exports, module) {
         indentFold              = require("foldhelpers/indentFold"),
         selectionFold           = require("foldhelpers/foldSelected");
 
-
     /** Set to true when init() has run; set back to false after deinit() has run */
     var _isInitialized = false,
         gutterObservers = {};
-
+        
     /**
       * Restores the linefolds in the editor using values fetched from the preference store
       * Checks the document to ensure that changes have not been made (e.g., in a different editor)
@@ -112,7 +116,7 @@ define(function (require, exports, module) {
             });
         }
 
-        var saveFolds = prefs.getSetting("saveFoldStates");
+        var saveFolds = prefs.getSetting(SAVE_FOLD_STATES);
         if (!editor || !saveFolds) {
             return;
         }
@@ -147,7 +151,7 @@ define(function (require, exports, module) {
       * @param {Editor} editor the editor whose line folds should be saved
       */
     function saveLineFolds(editor) {
-        var saveFolds = prefs.getSetting("saveFoldStates");
+        var saveFolds = prefs.getSetting(SAVE_FOLD_STATES);
         if (!editor || !saveFolds) {
             return;
         }
@@ -263,6 +267,7 @@ define(function (require, exports, module) {
       */
     function createGutter(editor) {
         var cm = editor._codeMirror;
+        var rootElement = editor.getRootElement();
         var path = editor.document.file.fullPath, _lineFolds = prefs.getFolds(path);
         _lineFolds = _lineFolds || {};
         cm._lineFolds = _lineFolds;
@@ -270,11 +275,11 @@ define(function (require, exports, module) {
 
         // Reuse any existing fold gutter
         if (gutters.indexOf(GUTTER_NAME) < 0) {
-            var lineNumberIndex = gutters.indexOf("CodeMirror-linenumbers");
+            var lineNumberIndex = gutters.indexOf(LINE_NUMBER_GUTTER);
             if (lineNumberIndex < 0) {
-                $(editor.getRootElement()).addClass("linenumber-disabled");
+                $(rootElement).addClass("linenumber-disabled");
             }
-            $(editor.getRootElement()).addClass("folding-enabled");
+            $(rootElement).addClass("folding-enabled");
             gutters.splice(lineNumberIndex + 1, 0, GUTTER_NAME);
             cm.setOption("gutters",  gutters);
             cm.refresh();  // force recomputing gutter width - .folding-enabled class affects linenumbers gutter which has existing cached width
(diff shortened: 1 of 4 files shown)


ficristo:
@thehogfather thank you for the PR.
In general, from now onwards, the inline directives are mostly forbidden (there are still some exceptions).
In this case you can put `MutationObserver` after `ArrayBuffer` both in `.eslintrc.json` and `.brackets.json`.

ficristo:
I think @MarcelGerber or @petetnt are better to review this one.

MarcelGerber:
Looks like #10864 is fixed now, but I still see the other issue with the active line highlight
image
(This is Brackets running without extensions on the preinstalled Brackets Dark theme)

thehogfather:
@MarcelGerber could you share a link to the active line highlight issue? This PR only addresses the issues mentioned in the description. Neither of those have anything to do with active line highlights.

MarcelGerber:
Ah, I see now. #11577 mentions the issue, but I thought it's all about it.
This may not even be the fault of this extension.

src/extensions/default/CodeFolding/main.js
```diff
@@ -270,9 +270,12 @@ define(function (require, exports, module) {
 
         // Reuse any existing fold gutter
         if (gutters.indexOf(GUTTER_NAME) < 0) {
-            var lnIndex = gutters.indexOf("CodeMirror-linenumbers");
+            var lineNumberIndex = gutters.indexOf("CodeMirror-linenumbers");
+            if (lineNumberIndex < 0) {
+                $(editor.getRootElement()).addClass("linenumber-disabled");
```
MarcelGerber: This might also go well here: https://github.com/adobe/brackets/blob/7814da193b98ff7889714a601d630973bddf02d0/src/editor/Editor.js#L2324-L2328

That way, stylesheets and other extension have an easy way to reuse this state.
cc @nethip

thehogfather:
No I don't think the extension introduced the issue - although to be honest I am not actually sure what the expectation is here - so a little difficult to explore a solution.

src/extensions/default/CodeFolding/main.js
```diff
@@ -316,6 +320,38 @@ define(function (require, exports, module) {
         if (editor._codeMirror.getOption("gutters").indexOf(GUTTER_NAME) === -1) {
             createGutter(editor);
             restoreLineFolds(editor);
+            // Watch mutations on code mirror gutters and ensure line numbers are added before fold gutter.
+            var config = {childList: true};
+            var gutters,
+                lineNumberIndex,
+                foldGutterIndex,
+                cm = editor._codeMirror,
+                guttersContainer = $(".CodeMirror-gutters", editor.getRootElement()),
```
petetnt: `editor.getRootElement()` is used here, on L342, L344 and L346 so it could / should be saved to an variable

src/extensions/default/CodeFolding/main.js
```diff
@@ -371,6 +407,11 @@ define(function (require, exports, module) {
         Editor.forEveryEditor(function (editor) {
             CodeMirror.commands.unfoldAll(editor._codeMirror);
             removeGutter(editor);
+            //disconnect any mutation observers on the gutter
+            if (gutterObservers[editor.document.file.fullPath]) {
```
petetnt: `gutterObservers[editor.document.file.fullPath]` could be cached to an variable too

src/extensions/default/CodeFolding/main.less
```diff
@@ -49,6 +49,24 @@
     padding-right: 5px;
 }
 
+// If line numbers are not shown and codefolding is enabled we remove the left padding.
+ // We add the same padding below to the fold gutter
+ .show-line-padding {
+     .folding-enabled.linenumber-disabled pre{
```
petetnt: Nit: Space between `pre` and `{`

src/extensions/default/CodeFolding/main.less
```diff
@@ -49,6 +49,24 @@
...
+     }
+ }
+ 
+ .CodeMirror.linenumber-disabled {
+     // When there are no line numbers, .show-line-padding class adds 15px left padding to line numbers.
+     // Here we compensate for that gap by adding the same padding to the folding gutter.
+     .CodeMirror-foldgutter,
+     .CodeMirror-foldgutter-open,
+     .CodeMirror-foldgutter-folded,
+     .CodeMirror-foldgutter-blank{
```
petetnt: Nit: Space between `-blank` and `{`

petetnt:
Small general nit: Would be nice if all the magic strings would be moved to constants in the file, would make modifications somewhat easier / safer (can be done in another PR too though)!

Good job @thehogfather once again :+1:

src/extensions/default/CodeFolding/main.js
```diff
@@ -263,34 +267,38 @@ define(function (require, exports, module) {
...
         var gutters = cm.getOption("gutters").slice(0);
 
         // Reuse any existing fold gutter
         if (gutters.indexOf(GUTTER_NAME) < 0) {
-            var lnIndex = gutters.indexOf("CodeMirror-linenumbers");
-            $(editor.getRootElement()).addClass("folding-enabled");
-            gutters.splice(lnIndex + 1, 0, GUTTER_NAME);
+            var lineNumberIndex = gutters.indexOf(LINE_NUMBER_GUTTER);
+            if (lineNumberIndex < 0) {
+                $(rootElement).addClass("linenumber-disabled");
```
thehogfather: @MarcelGerber originally wrote : 

>This might also go well here: https://github.com/adobe/brackets/blob/7814da193b98ff7889714a601d630973bddf02d0/src/editor/Editor.js#L2324-L2328

>That way, stylesheets and other extension have an easy way to reuse this state.
cc @nethip

zaggino:
Is this waiting for anything @petetnt ?

petetnt:
I'll try to test this out once more and run the tests ASAP and merge this in if all is good.

petetnt:
There are some failing Code Folding tests which are unrelated to this PR which are being tracked in #12724.

This PR LGTM, merging.

zaggino:
Getting back to this, related to https://github.com/adobe/brackets/issues/12726#issuecomment-242946372 I believe this should be reverted, using MutationObserver is not a good thing. Especially if some other extensions tries to do the same thing, I expect we start to get infinite loops?

cc @MarcelGerber @petetnt @thehogfather

thehogfather:
This is a symptom of an underlying limitation (within the core libraries CodeMirror and brackets) - i.e., the absence of a rich API for manipulating the gutter in the editor. 

At the moment, every extension (that requires the gutter) splices and directly manipulates the gutter without regard for how any other extension might be using the gutter wrt relative position / placement  etc. Indeed it is difficult to preempt how many different ways the gutter might be used by other extensions.

What might be immediately useful in a hypothetical API would be the notion of an anchor gutter or gutters (say the line numbers gutter) such that  placement of other gutters are requested relative to said anchor gutter(s). So one might say `insert the code-folding gutter after the line-numbers`, `insert the git gutter before the line-numbers`, `insert theseus-gutter after git, before line numbers` etc.

If such an API (or a much refined variation) existed, then there wouldn't be need for the MutationObserver (since all requests to manipulate would go through said API).

Re: infinite loops,
Totally possible if different extensions decided to use mutationobservers directly as well. The implementation in this PR itself should not cause infinite loops as it disconnects the observer while updating the gutter.

cc @zaggino @MarcelGerber @petetnt 
ref #12726

zaggino:
We should definitely set up an API in brackets, make this work with that API and let authors fix their extensions to use the API. Current state is not maintenable, and if we add this, another authors will do the same for their purposes and then we'll have a hell around a simple feature.

petetnt:
Haven't/didn't run to #12726 but let's revert this one and let's think of a nicer API for doing this.

PR #12675, opened 2016-08-14
Adds code folding support for handlebars template files

thehogfather:
Adds a new range finder to detect foldable regions and enable code folding for blocks and helpers in handlebars template files.


src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,158 @@
+/**
+ * Fold range finder for handlebars/mustache template type files.
+ * @author Patrick Oladimeji
+ * @license MIT
+ * @date 14/08/2016 22:04:21
+ */
```
ficristo: I _think_ you have to use the standard license comment of adobe on newer file now the code folding extension is in the core.

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,158 @@
...
+
+define(function (require, exports, module) {
+    "use strict";
+    var CodeMirror = brackets.getModule("thirdparty/CodeMirror/lib/codemirror"),
+        space = /\s/;
```
ficristo: Can you rename this to make clear is a regex? Something along spaceRegex or definetively something better?
Could you add a new line before comments?

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * Utility function for scanning the text in a document until a certain condition is met
+     * @param {object}   cm        The code mirror object representing the document
+     * @param {string}   startCh   The start character position for the scan operation
+     * @param {number}   startLine The start line position for the scan operation
+     * @param {function} condition A predicate function that takes in the text seen so far
+     */
```
ficristo: Add the return parameter?

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,158 @@
...
+            characterIndex = startCh,
+            currentLine = startLine;
+        while (currentLine <= cm.lastLine()) {
+            if (line.length === 0) {
+                characterIndex = 0;
+                line = cm.getLine(++currentLine);
+            } else {
+                seen = seen.concat(line[characterIndex] || '');
+                if (condition(seen)) {
+                    return {from: {ch: startCh, line: startLine}, to: {ch: characterIndex, line: currentLine}, string: seen};
```
ficristo: Could you hoist this in a variable and use some newlines?
At least something like:
```js
var obj = {
  from: {ch: startCh, line: startLine},
  to: {ch: characterIndex, line: currentLine},
  string: seen
};
return obj;
```
Better name welcome.
It will make diff easier to read.

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,158 @@
...
+                } else if (characterIndex >= line.length) {
+                    seen = seen.concat(cm.lineSeparator());
+                    if (condition(seen)) {
+                        return {from: {ch: startCh, line: startLine}, to: {ch: characterIndex, line: currentLine}, string: seen};
```
ficristo: Ditto here.

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,158 @@
...
+                    line = cm.getLine(++currentLine);
+                } else {
+                    ++characterIndex;
+                }
+            }
+        }
+    }
+    
+    function endTag(seen) {
+        return space.test(seen[seen.length - 1]) || seen[seen.length - 1] === '}';
```
ficristo: Do you think is worth hoisting this in a function `seen[seen.length - 1]`?

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,158 @@
...
+        if (openTagIndex < 0 || text[openTagIndex + 2] === '/') {
+            return;
+        }
+        
+        found = scanTextUntil(cm, openTagIndex + 2, currentLine, endTag);
+        if (!found) {
+            return;
+        }
+        
+        if (found) {
```
ficristo: You have already checked above.

src/extensions/default/CodeFolding/main.js
```diff
@@ -61,13 +61,14 @@ define(function (require, exports, module) {
     brackets.getModule(["thirdparty/CodeMirror/addon/fold/brace-fold"]);
     brackets.getModule(["thirdparty/CodeMirror/addon/fold/comment-fold"]);
     brackets.getModule(["thirdparty/CodeMirror/addon/fold/markdown-fold"]);
-
+
```
ficristo: Try to match the indentation.
(Personally I'm not keen on lines with only whitespaces but that is another story)

src/extensions/default/CodeFolding/main.js
```diff
@@ -393,9 +394,9 @@ define(function (require, exports, module) {
             return prefs.getSetting("alwaysUseIndentFold");
         }, indentFold);
 
-        CodeMirror.registerHelper("fold", "django", CodeMirror.helpers.fold.brace);
-        CodeMirror.registerHelper("fold", "tornado", CodeMirror.helpers.fold.brace);
```
ficristo: Are not needed anymore?

src/extensions/default/CodeFolding/unittests.js
```diff
@@ -41,6 +41,13 @@ define(function (require, exports, module) {
                     sameLevelFoldableLines: [8, 24],
                     firstSelection: {start: {line: 3, ch: 0}, end: {line: 10, ch: 0}},
                     secondSelection: {start: {line: 6, ch: 0}, end: {line: 17, ch: 4}}
+                },
+                hbs: {
+                    filePath: testDocumentDirectory + "test.hbs",
+                    foldableLines: [1, 7, 14, 16, 17, 21, 26, 28, 29, 32, 33],
+                    sameLevelFoldableLines: [1, 7, 14],
+                    firstSelection: {start: {line: 2, ch: 0}, end: {line: 10, ch: 0}},
+                    secondSelection: {start: {line: 5, ch: 0}, end: {line: 8, ch: 4}}
```
ficristo: :+1:

src/extensions/default/CodeFolding/unittests.js
```diff
@@ -163,7 +170,7 @@ define(function (require, exports, module) {
         /**
          * Helper function to return the fold markers on the current codeMirror instance
          *
-         * @returns {[[Type]]} [[Description]]
+         * @returns {Array<object>} An array of objects containing the line and the type of marker.
```
ficristo: :+1:

ficristo:
I don't know handlebar, so I left some general comments.
If nobody will look at the correctness of this, I'll give a try on some examples and eventually merge.

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
thehogfather: no problem - I'll remove it -artefact of automatic templates

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
thehogfather: Not sure I understand the question. I think wrapping the boolean function into the function makes the getRange function easier to read. I can rename and comment.

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
thehogfather: Good catch!

src/extensions/default/CodeFolding/main.js
thehogfather: I"ll get rid of the spaces

src/extensions/default/CodeFolding/main.js
thehogfather: No - it also turns out that the CodeMirror.helpers.fold.brace is undefined at the point this helpers are registered. Consequently no point in having them here. Also I don't think brackets actually support these modes so they are better addressed in an extension that provides support for the mode

src/extensions/default/CodeFolding/main.js
```diff
@@ -393,9 +394,9 @@ define(function (require, exports, module) {
...
-
+        CodeMirror.registerHelper("fold", "handlebars", handlebarsFold);
+        CodeMirror.registerHelper("fold", "htmlhandlebars", handlebarsFold);
+
```
thehogfather: will remove space.

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
ficristo: Sorry, I think it easy to forget the -1.
So having something:
```js
function lastchar(string) {
  return string[string.length - 1]
}
function endTag(seen) {
  return space.test(lastchar(seen)) || lastchar(seen) === '}';
}
```
or maybe leave the first condition as is and use
```js
StringUtils.endsWith(seen, "}");
```
on the second condition and elsewhere.
```js
function endTag(seen) {
  return space.test(seen[seen.length - 1]) || StringUtils.endsWith(seen, "}");
}
```

ficristo:
Another nit but in general Brackets use double quote for strings.

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,174 @@
+/**
```
ficristo: I think you have to include the full licence comment here (like in other source files).
Beware the date has to be 2016 - present.
```js
/*
 * Copyright (c) 2016 - present Adobe Systems Incorporated. All rights reserved.
 *
 * Permission is hereby granted, free of charge, to any person obtaining a
 * copy of this software and associated documentation files (the "Software"),
 *....
 *...
```

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,174 @@
...
+                    line = cm.getLine(++currentLine);
+                } else {
+                    ++characterIndex;
+                }
+            }
+        }
+    }
+
+    function endTag(seen) {
+        return endOfLineSpaceRegex.test(seen) || seen.endsWith("}");
```
ficristo: You have to use the `StringUtils.endsWith` counterpart.
According to [MDN](https://developer.mozilla.org/it/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) Chrome support it since version 41 and Linux is at 29, and IE doesn't support it and we will break the in browser approach.

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,198 @@
...
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+/**
+ * Fold range finder for handlebars/mustache template type files.
+ * @author Patrick Oladimeji
+ * @date 14/08/2016 22:04:21
+ */
+
```
thehogfather: @ficristo BTW I notice this empty line is being introduced throughout the code base - although it conflicts with an existing recommendation in the fairly stale [JS Docs guidelines](https://github.com/adobe/brackets/wiki/Brackets-JSDocs-Guidelines#header-comments). I am guessing that the guideline has simply not been updated?

ficristo: AFAIK the guidelines are still true. But until we can have stricter rules on our linter sometimes we miss things.

src/extensions/default/CodeFolding/main.js
```diff
@@ -393,8 +394,8 @@ define(function (require, exports, module) {
             return prefs.getSetting("alwaysUseIndentFold");
         }, indentFold);
 
-        CodeMirror.registerHelper("fold", "django", CodeMirror.helpers.fold.brace);
-        CodeMirror.registerHelper("fold", "tornado", CodeMirror.helpers.fold.brace);
+        CodeMirror.registerHelper("fold", "handlebars", handlebarsFold);
+        CodeMirror.registerHelper("fold", "htmlhandlebars", handlebarsFold);
```
ficristo: @thehogfather is the second parameter a CodeMirror mode? Where do you use it?
It seems to me that Mustache is not part of your PR, but handlebars is a Mustache like template.
And we use Mustache in the codebase.
So if it that param is a mode, I was wondering if we can use `text/x-brackets-html` or `htmlmixed`

thehogfather: yup that parameter is a CodeMirror mode. The name of the mode is used internally by CodeMirror to create a lookup table for the fold helper for a given mode.

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,198 @@
...
+define(function (require, exports, module) {
+    "use strict";
+    var CodeMirror = brackets.getModule("thirdparty/CodeMirror/lib/codemirror"),
+        endOfLineSpaceRegex = /\s$/,
```
ficristo: My little mind cannot understand this: are you looking for a single space, right? For which cases?

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
```diff
@@ -0,0 +1,198 @@
...
+        if (!found) {
+            return;
+        }
+
+        openPos = {
+            from: {line: currentLine, ch: openTagIndex},
+            to: found.to
+        };
+        openTag = found.string.substring(0, found.string.length - 1);
+        if (openTag[0] === "#" || openTag[0] === "~") {
```
ficristo: Here and below I think you have to add `^` and add a test for it and for `~`

ficristo:
I left a question and a request of change.
I didn't check everything, leaving that to the tests
Could merge master (again)? It will be easier for final check

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
thehogfather: This looks for a space that occurs at the end of a string - and I have used it to scan a line until a space is reached. Very useful in finding potential opening range tags - e.g., `{{` and `{{foo` in the following scenario:

```html
{{ 
    foo 
    bar=baz 
    var-1=goo
}}
```

or 

```html
{{foo 
    bar=baz 
    var-1=goo
}}
```

And based on what is currently on the stack of the opening tags we try to search for a closing tag. I will add better comments to these. Perhaps endOfLineSpaceRegex is actually not a good name and naming the variable might be unnecessary in this scenario as it is only used in one place.

src/extensions/default/CodeFolding/foldhelpers/handlebarsFold.js
thehogfather: Good point. I was working with the handlebars page as reference. `#*` should also be added.

thehogfather:
@ficristo PR has been updated.

ficristo:
@thehogfather thank you.
I have some problems running all CodeFolding tests. Can you run them and are all green?

ficristo:
On a PC where I was able to run tests I had a consistent failure on the ``can be enabled by setting `makeSelectionsFoldable' to true`` test.

thehogfather:
@ficristo - could you clarify if this use case also fails in use - and does it fail for all three file types, `js`, `html`, and `handlebars`? Could you change the preference in the project settings and  see if selection fold is enabled in the editor.
All tests are passing on my mac - and I have no access to a PC to test that out.

ficristo:
Actually I tested it on OSX too. I'll try to understand more the issue.

thehogfather:
And do you get the same failure on OSX?

ficristo:
yes

thehogfather:
Very strange. Let me know if I can help - unfortunately, I have now tried different macs and I still cannot reproduce.

ficristo:
I run again the tests on OSX and there still was a failure. I tryed to add a couple of breakpoints
and the test passed...
Then I tryed to enclose the `selectTextInEditor` call in a `runs` function and all the tests passed.
```js
runs(function () {
  selectTextInEditor(start, end);
});
```
I can only guess there is some async problem, does it make sense to you?

thehogfather:
To be honest I"m a little baffled as to why that would fix it as the `selectTextInEditor` function itself uses the `runs` in its definition so the next block should not run until the selection is ready.

ficristo:
I had some problem with code folding tests in the past too.
So, since they work for you, I decided to merge.
Thank you for you patience.

PR #12681, opened 2016-08-16
do nothing when trying to reset with the same content which is already present

zaggino:
I've encountered a case of https://github.com/adobe/brackets/issues/11826 and this fixes it (don't necessarily fix all possible cases but it is at least something)

cc @nethip @petetnt

### DIFF
--- a/src/editor/Editor.js
+++ b/src/editor/Editor.js
@@ -1047,6 +1047,12 @@ define(function (require, exports, module) {
      * @param {!string} text
      */
     Editor.prototype._resetText = function (text) {
+        var currentText = this._codeMirror.getValue();
+        if (text === currentText) {
+            // there's nothing to reset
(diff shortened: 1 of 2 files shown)


ficristo:
@zaggino it would be awesome if you can add a test case.

petetnt:
Most likely fixes things like zaggino/brackets-git#1274 to some extent. Looks good to me.

zaggino:
@ficristo test added, i've tested that it fails before and passes after the fix too

nethip:
Nice fix @zaggino . LGTM. 

Note: I am going ahead with the merge. Travis is failing because of the API limit issues.

nethip:
Thanks @zaggino

ficristo:
> test added

@zaggino great job, thank you.

PR #12718, opened 2016-08-25
ESLint: enable strict rule

ficristo:
I removed the node env globally and then added only to node files.
After I enabled the strict rules which is (almost) compatible with JSLint.

Since I've removed the IIFE in node file you see a lot of changes.
You can look at https://github.com/adobe/brackets/pull/12718/files?w=1 to see code changes only.


zaggino:
LGTM

PR #12719, opened 2016-08-26
ESLint: enable no-bitwise rule

ficristo:


zaggino:
Agreed, this is not something that you want normally to do. Should be turned off per line/file.

PR #12721, opened 2016-08-27
Adds autofill detail tokens to the autocomplete attribute

Dandandans:
Inspired by [Jason Grigsby](https://cloudfour.com/thinks/autofill-what-web-devs-should-know-but-dont/) and [Christian Heilmann](https://www.smashingmagazine.com/2016/08/contributing-open-source/), this update adds [autofill detail tokens](https://html.spec.whatwg.org/multipage/forms.html#autofill) to the `autocomplete` attribute.


Denisov21:
Good idea!

mcdado:
Lovely! I was about to make the same PR since Atom uses this list.

But there is one problem, I think: looking at [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete](<input> - HTML | MDN) I see that you did not include "on" and "off".

Dandandans:
@mcdado, yes, it's a nice +1 for both Atom and Brackets. 

I intended to include the `on` and `off` values, but I misread the [WHATWG standard](https://html.spec.whatwg.org/multipage/forms.html#autofill)  for the autofill anchor mantle. 

When the `autocomplete` attribute is wearing the autofill anchor mantle, `on` and `off` shouldn't be combined with autofill detail tokens — e.g. `autocomplete="yes street-address"`. 

The values `on` and `off` are still part of the spec, even though `off`'s usefulness is [debatable](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion), and should be included as values.

Thanks for helping me rethink this!

zaggino:
:+1:

PR #12725, opened 2016-08-27
Addresses typo in `selection fold` preference key

thehogfather:
Addresses issue #12724 where selected text was no longer foldable even when enabled in preferences - hence causing tests to fail.


thehogfather:
@petetnt I believe this PR would be unnecessary if we decide to revert 57278636e2bfacdea8473c3745f9cac7e1550a73

thehogfather:
Closing as no longer needed.

PR #12727, opened 2016-08-28
Revert "Ensures fold-gutter is always inserted after the line-number gutter (if it exists)"

petetnt:
Reverts adobe/brackets#12673

The PR had some issues regarding scroll states in some cases which I didn't catch/that didn't occur to myself.

Let's revert the PR and think of nicer API instead.

Reopen #11577 #10864 and #12725 when this gets merged. 

/cc @zaggino

### DIFF
--- a/.brackets.json
+++ b/.brackets.json
@@ -20,7 +20,6 @@
             "clearTimeout",
 
             "ArrayBuffer",
-            "MutationObserver",
             "XMLHttpRequest",
             "Uint32Array",
             "WebSocket"
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -61,7 +61,6 @@
         "clearTimeout": false,
 
         "ArrayBuffer": false,
-        "MutationObserver": false,
         "XMLHttpRequest": false,
         "Uint32Array": false,
         "WebSocket": false
--- a/src/extensions/default/CodeFolding/main.js
+++ b/src/extensions/default/CodeFolding/main.js
@@ -25,6 +25,7 @@
  * @author Patrick Oladimeji
  * @date 10/24/13 9:35:26 AM
  */
+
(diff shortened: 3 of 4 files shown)


thehogfather:
@zaggino 

Only thing I would say was missing is an ability to list current gutters and their priorities.  So:
```js
.getGutters():Array<{name: string, priority: number}>
```
This would enable extension authors to avoid/address any potential conflicts with existing gutters names and priorities. Both name and priorities should be unique - possibly throw errors on conflict.

Also is there any reason why all methods shouldn't be in `Editor.js`?

zaggino:
@thehogfather true, that method was missing.

The reason I'd put registerGutter, unregisterGutter, getRegisteredGutters on the EditorManager and not the Editor itself, it's that I'm able to just call `registerGutter` on startup of my extension, and there'll always be reserved place for it. I don't have to manage on which Editor's I have registered my gutter and on which not. Of course if my extension is language specific we probably don't want to have gutters on all languages and this makes it kind of complex.  We could add another parameter to be array of LanguageId's of default to be all languages.

But... if you want to make it simply, they can all live on the editor.

About the conflicts, I think we should hardcode priorities for inbuilt default gutters (line numbers, code folding) and console warning when someone else tries to use them. Besides that, let extensions set which priority they want. For example all my linting extensions would probably use the same priority, whatever the number would it be.

zaggino:
Merging this, @thehogfather please submit a PR with the API proposal or let me know if I should work on it. I've been hitting https://github.com/adobe/brackets/issues/12726 all the time and it's quite bad side effect. It can't make it to the release.

thehogfather:
@zaggino I'll submit a PR.

zaggino:
Thanks

PR #12730, opened 2016-08-28
Update main.less

alex-wdmg:
Scrollbar Styling in Dark Theme

### DIFF
--- a/src/extensions/default/DarkTheme/main.less
+++ b/src/extensions/default/DarkTheme/main.less
@@ -38,6 +38,49 @@
 @background: #1d1f21;
 @foreground: #ddd;
 
+/* Scrollbar Styling */
+
+::-webkit-scrollbar {
+    width: 12px;
+    height: 12px;
+}
+::-webkit-scrollbar-button {
+    width: 12px;
+    height: 12px;
+}
+::-webkit-scrollbar-thumb {
+    background: rgba(240,240,240, 0.3);
+    border: 3px solid @background;
+    border-radius: 12px;
+    margin: 3px 3px;
+}
+::-webkit-scrollbar-thumb:hover {
+    background: rgba(240,240,240, 0.5);
+    border: 3px solid @background;
+}
+::-webkit-scrollbar-thumb:active {
+    background: #ffffff;
+    border: 3px solid #2f2f2f;
+}
+::-webkit-scrollbar-track {
+    background: @background;
+    border: 0;
+}
+::-webkit-scrollbar-track:hover {
+    background: transparent;
+}
+::-webkit-scrollbar-track:active {
+    background: #2f2f2f;
+}
+::-webkit-scrollbar-button {
+    width: 0;
+    height: 0;
+    display: none;
+}
+::-webkit-scrollbar-corner {
+    background: transparent;
+}
+
 /* Code Styling */
 
 .CodeMirror, .CodeMirror-scroll {


zaggino:
@wdmg You need to sign the http://dev.brackets.io/brackets-contributor-license-agreement.html

wdmg:
Okey

MarcelGerber:
@wdmg The scrollbar should already be dark when using a dark theme, and it works fine for me on Windows.
If it doesn't work for you, what OS are you on?

wdmg:
Yes, you notice right ... the problem with displaying scrollbar main area Brackets code editor on the Mac (10.10.5). See screenshot (without CSS cutomize):


At the same time, the option to use styles for the scrollbar on what does not influence ((

zaggino:
How does this look on Mac? Linux?
Does it influence Windows at all?

wdmg:
I think this will affect the display of any of the platforms, with certain minor differences. But I did not notice that the scrollbar was white on Win. Perhaps there was another theme design... At the moment, I can not test on Win. And, Linux not use.

wdmg:
Add screenshot with CSS rules for scrollbar:

screenshots

zaggino:
Who can review this? @larz0 @MiguelCastillo @ficristo ?
Should this be platform specific so we don't change Windows scrollbars?

ficristo:
I cannot review meaningfully the styles but I still have some questions:
- does this work only with the default dark theme or with all darks themes?
- you have two classes `webkit-scrollbar-button`, is it intended?
- if it will works i much prefer something like this both for `-webkit-scrollbar-thumb` and `-webkit-scrollbar-track`:
```less
::-webkit-scrollbar-track {
  &:hover {
  }
  &:active {
  }
}
```

petetnt:
The scrollbar correctly shows as dark for me on MacOS 10.11.5 with Brackets Dark-theme without this PR. I have, however, seen situations where the scrollbar turns unstyled, but haven't been able to replicate that.

ficristo:
This is relevant https://github.com/adobe/brackets/issues/8982

petetnt:
@ficristo nice, I switching between magic trackpad and a regular mouse every now and then and most likely when I which back to regular (non-Apple) mouse the scrollbars will show up as white.

larz0:
@zaggino yes this should be platform specific 👍

zaggino:
for what's worth, I've launched brackets on mac yesterday and I definitely saw unstyled scrollbars with brackets dark theme, so there seems to be a bug somewhere ... per @larz0 comments, @wdmg this needs to be made platform specific so it doesn't affect windows (not sure about linux, it probably should affect linux) ... if anybody is able to pinpoint why some users have unstyled scrollbars on mac, that'd be even better

HRoland1:
I had the same issue, I applied the CSS and now it works and it's beautiful.
So thank you @wdmg! 👍

TomMalbran:
There is a brackets/src/styles/brackets_scrollbars.less file. The scrollbar changes should be done there.

wdmg:
> There is a brackets/src/styles/brackets_scrollbars.less file. The scrollbar changes should be done there.

The fact that there is a code of scrollbar styling ... but for some reason he does not work or works fail.

TomMalbran:
It does work if you use a touchpad or a wireless mouse, since it uses the default mac scrollbars, which are hidden by default. It only fails when you plug a mouse.

ficristo (changes requested):
As mentioned by @TomMalbran, the styles should be moved to `brackets/src/styles/brackets_scrollbars.less`.
Please also look at my previous comment.

madanbn:
Moving this to 1.9 as we are closer to the 1.8 release and will need more time to close out on this.

ficristo:
This needs more work. Closing for now.

Issue #12360, opened 2016-04-12
Wiki-pages related to the Linux version are out of date


petetnt:
The README.md points [to this article](https://github.com/adobe/brackets/wiki/Linux-Version) which is pretty out of date. Similarly that article points to https://github.com/adobe/brackets/wiki/Linux-Development-for-Contributors which is also badly out of date. 

AFAIK there has not been much work on the Linux side of things as of late, but the articles should be updated at least with common issues such as LibCrypt missing, CEF not being updated, missing features vs. the other versions etc.

PR #12761, opened 2016-09-07
Created a Guide for Common Linux Issues

NathanJPlummer:
Wrote a guide to fix common issues in Linux, especially Ubuntu.

These problems pop up in the "issues" section every few days despite having several solutions.

Added a link on your main page.  Furthermore moderators can post a link to this guide when someone reports these problems, rather then having to type out the same responses repeatedly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ Having problems starting Brackets the first time, or not sure how to use Bracket
 review [Troubleshooting](https://github.com/adobe/brackets/wiki/Troubleshooting), which helps 
 you to fix common problems and find extra help if needed.
 
+For a list of common Linux issues and workarounds you can [visit this guide](https://nathanjplummer.github.io/Brackets/).
+
 
 Helping Brackets
 ----------------


zaggino:
@NathanJPlummer great docs indeed. 
Even better place for this would be https://github.com/adobe/brackets/wiki

zaggino:
@NathanJPlummer you need to sign the http://dev.brackets.io/brackets-contributor-license-agreement.html

NathanJPlummer:
@zaggino

Thanks for the feedback.  Signed the form.

I agree the wiki is probably a better location.  From what my research is telling me, I can't do a pull request on a wiki?

I've included the original markdown file I used to build the linked page, but if there's a better solution to this please let me know.  If I see more issues like this pop up I'll attempt to write more documentation in the future.

[Brackets_Linux.zip](https://github.com/adobe/brackets/files/458412/Brackets_Linux.zip)

petetnt:
👍 
Relevant issue I made sometime ago: #12360 

I'd love to see the content in the wiki and I also want the page linked directly to the `README.md` at least as long as generic Linux problems are solved.

redmunds:
Unfortunately, spam bots were polluting the wiki, so currently only committers can update wiki pages.

zaggino:
Merging this then, so @NathanJPlummer can keep it up-to-date on his side until we find a better solution (hostings /docs folder instead of using wiki seems best to me)

NathanJPlummer:
@zaggino 

Works for me.  I'll check issues a couple times a day and add any Linux related issues I see to this.

Thanks!

PR #12767, opened 2016-09-09
Fixed leak in menu.js file. Fixes #12765

### LINKED ISSUE #12765: leaks in src/command/Menus.js
`Menu.prototype.addMenuItem(CMD_ID)` (1) calls constructor MenuItem(id, command) (2) which attaches event handlers to command (3). Calling `Menu.prototype.removeMenuItem(CMD_ID)` (4) doesn't remove these handlers leading to infinitely increasing numbers of event handlers on command.

1) https://github.com/adobe/brackets/blob/master/src/command/Menus.js#L546
2) https://github.com/adobe/brackets/blob/master/src/command/Menus.js#L586
3) https://github.com/adobe/brackets/blob/master/src/command/Menus.js#L295-L316
4) https://github.com/adobe/brackets/blob/master/src/command/Menus.js#L432-L466

One example of this would be https://github.com/zaggino/brackets-git/blob/master/src/Main.ts#L173-L203

daytonjallen:
Fixes https://github.com/adobe/brackets/issues/12765

### DIFF
--- a/src/command/Menus.js
+++ b/src/command/Menus.js
@@ -182,13 +182,13 @@ define(function (require, exports, module) {
     * Removes the attached event listeners from the corresponding object.
     * @param {Command} command
     */
-    function removeMenuItemEventListeners(command) {
-        command
-            .off("enabledStateChange", command._enabledChanged)
-            .off("checkedStateChange", command._checkedChanged)
-            .off("nameChange", command._nameChanged)
-            .off("keyBindingAdded", command._keyBindingAdded)
-            .off("keyBindingRemoved", command._keyBindingRemoved);
+    function removeMenuItemEventListeners(menuItem) {
+        menuItem._command
+            .off("enabledStateChange", menuItem._enabledChanged)
+            .off("checkedStateChange", menuItem._checkedChanged)
+            .off("nameChange", menuItem._nameChanged)
+            .off("keyBindingAdded", menuItem._keyBindingAdded)
+            .off("keyBindingRemoved", menuItem._keyBindingRemoved);
     }
 
     /**
@@ -457,13 +457,14 @@ define(function (require, exports, module) {
                 console.error("removeMenuItem(): command not found: " + command);
                 return;
             }
-            removeMenuItemEventListeners(commandObj);
             commandID = command;
         } else {
-            removeMenuItemEventListeners(command);
             commandID = command.getID();
         }
         menuItemID = this._getMenuItemId(commandID);
+
+        var menuItem = getMenuItem(menuItemID);
+        removeMenuItemEventListeners(menuItem);
 
         if (_isHTMLMenu(this.id)) {
             // Targeting parent to get the menu item <a> and the <li> that contains it
--- a/src/command/Menus.js
+++ b/src/command/Menus.js
@@ -179,6 +179,19 @@ define(function (require, exports, module) {
     }
 
     /**
+    * Removes the attached event listeners from the corresponding object.
+    * @param {Command} command
+    */
+    function removeMenuItemEventListeners(command) {
+        command
+            .off("enabledStateChange", command._enabledChanged)
+            .off("checkedStateChange", command._checkedChanged)
+            .off("nameChange", command._nameChanged)
+            .off("keyBindingAdded", command._keyBindingAdded)
+            .off("keyBindingRemoved", command._keyBindingRemoved);
+    }
+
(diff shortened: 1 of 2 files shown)


ficristo:
I know it is a bit confusing, but please, revert the changes to `src/config.json`.

daytonallen:
I don't remember touching that file but I've reverted the changes. Sorry about that. Now the test is failing, looking into that now.

daytonallen:
Checking the CI logs, I think the reason for the build failure isn't anything on my end. Right?

MarcelGerber:
Yes, the build failure was on our side. It passed now :thumbsup:

cc @zaggino for review

zaggino:
@daytonallen the code is great, but is missing a test and that should be done in this case as test for this is easy to write, please add it somewhere here: https://github.com/adobe/brackets/blob/master/test/spec/Menu-test.js#L660-L737

You should write it in a way that it fails when you comment out the `removeMenuItemEventListeners(commandObj);` call you added and passes when you uncomment it.

You'll need to compare contents of `._eventHandlers[eventName].length` before and after calling `removeMenuItem` in the test.

daytonallen:
How do I run the tests locally to check if it's done correctly? I tried the `debug`'s menu `run tests` option but it doesn't seem to be detecting a new test case.

zaggino:
@daytonallen did you setup your installed Brackets with your cloned Brackets folder like in here https://github.com/adobe/brackets/wiki/How-to-Hack-on-Brackets ? See the `setup_for_hacking` thingy.

zaggino:
If you did, please push the testcase here, maybe something else is wrong with it, you can still correct it with additional commits.

daytonallen:
Yeah I did

zaggino:
@daytonallen you need to look for it here, it runs for me:
image
image

daytonallen:
Turns out I was testing under the wrong grouping.

test/spec/Menu-test.js
```diff
@@ -734,6 +734,39 @@ define(function (require, exports, module) {
...
+                    expect(menuItem).toBeDefined();
+
+                    $listItems = menuDOMChildren(menuItemId);
+                    expect($listItems.length).toBe(1);
+
+                    var command = CommandManager.get(commandId);
+                    expect(typeof (command)).toBe("object");
+
+                    // Test event
+                    command.on("nameChange", menuItem._nameChanged);
```
zaggino: you shouldn't add event manually with `.on`, you should just do `expect` for events that are added by `addMenuItem`

zaggino:
@daytonallen 
You can remove stuff copied from other tests, it should be somthing simple like
```
menu.addMenuItem(commandId);
expect(command._eventHandlers....);
menu.addMenuItem(commandId);
expect(command._eventHandlers....);
```
everything else doesn't need to be there

src/command/Menus.js
```diff
@@ -179,6 +179,19 @@ define(function (require, exports, module) {
     }
 
     /**
+    * Removes the attached event listeners from the corresponding object.
+    * @param {Command} command
+    */
+    function removeMenuItemEventListeners(command) {
+        command
+            .off("enabledStateChange", command._enabledChanged)
```
zaggino: looking at the code now, `_enabledChanged` and other methods live on `MenuItem` not on `command` ... are you sure that you're not calling `.off("enabledStateChange", undefined)` ???

test/spec/Menu-test.js
```diff
@@ -734,6 +734,35 @@ define(function (require, exports, module) {
...
+                runs(function () {
+                    var menuItemId = "menu-test-removeMenuItem4";
+                    var commandId = "Menu-test.removeMenuItem.command4";
+                    CommandManager.register("Brackets Test Command Custom", commandId, function () {});
+                    var menu = Menus.addMenu("Custom", menuItemId);
+
+
+                    var menuItem = menu.addMenuItem(commandId);
+
+                    var command = CommandManager.get(commandId);
```
zaggino: to be sure you're detaching only stuff that's been added by `addMenuItem` I'd suggest adding a custom event here like this: `command.on("nameChange", function () {})` and that handler should stay attached even after you call `removeMenuItem`

zaggino:
Thanks @daytonallen merged to master! :+1:

daytonallen:
My pleasure. Thanks for all the guidance also.

PR #12778, opened 2016-09-12
Create config.json based on task run

ficristo:
This is a possibly fix for https://github.com/adobe/brackets/issues/12771 which rely on grunt names.
This works under the assumption that `grunt build` is always run after `grunt install` and `grunt set-release`. So if you want to try the installer locally you end up with the production file, unfortunately.
So maybe we really need an environment variable on the build machines.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,9 @@ Thumbs.db
 # ignore .disabled file for default extensions
 /src/extensions/default/*/.disabled
 
+# generate through grunt
+/src/config.json
+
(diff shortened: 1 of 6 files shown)


swmitra:
@ficristo Can you please resolve the merge conflict with master?

ficristo:
Rebased.

zaggino (approved):

swmitra:
@ficristo This is indeed beneficial for developers running at masters tip.  Great work as always 👍

PR #12780, opened 2016-09-14
set eslints max-len to 120

zaggino:
while 80 might be too narrow in some cases, anything over 120 means you have to scroll horizontally when reviewing pull requests on GH which is a real pain

cc @ficristo @petetnt


petetnt:
I totally okay with this. Might be fixable with `eslint --fix` too if there's tons of lines spanning over 120 characters.

ficristo:
I'm okay too.

PR #12785, opened 2016-09-16
add font-display property

samuelcarreira:
Docs: https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display

*****\* Values **********
auto
The font display strategy is defined by the user agent.
block
Gives the font face a short block period and an infinite swap period.
swap
Gives the font face no block period and an infinite swap period.
fallback
Gives the font face an extremely small block period and a short swap period.
optional
gives the font face an extremely small block period and no swap period.


zaggino (approved):

zaggino:
@samuelcarreira You need to sign http://dev.brackets.io/brackets-contributor-license-agreement.html before I can merge this in.

ficristo (changes requested):
I think you should sort the values alphabetically.
It seems to me that only `inherit` should be always the last value

ficristo (approved):

ficristo:
I should have read all the description...

PR #12820, opened 2016-10-08
Update CSSProperties.json

Ollie-w:
Added CSS grid properties. I'm using grid a lot so really need these!


sprintr:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/grid), this is an experimental technology. It will be better if you write an extension for it.

Ollie-w:
Its already been a draft for 3 years, the syntax isn't going to change and its definitely landing soon (and it will be very popular when it does so). It has reached W3C candidate recommendation stage and will land in all browsers unprefixed in early 2017.

Ollie-w:
https://css-tricks.com/snippets/css/complete-guide-grid/

petetnt:
Hi @Ollie-w,

I am okay with adding these into the core for the reasons you mentioned; the spec is rather stable and should land into userland rather soon.

You'll need to sign our Contributor License Agreement here first, as a it's a prerequisite for landing in PRs: http://dev.brackets.io/brackets-contributor-license-agreement.html

Ollie-w:
Thanks. I'm all signed up! :)

zaggino:
Quite confused here @petetnt , you said you're "okay with adding these into the core" and asked to sign the CLA but then you close the PR ?

petetnt:
@zaggino there was a follow up PR #12822 that included these changes and more so we close this one.

Issue #4954, opened 2013-08-27
Focus lost after closing Project Settings


peterflynn:
1. File > Project Settings
2. Click Cancel
3. Try to type anything

Result: Nothing happens. Keyboard focus is not in editor area.

We just need a simple `EditorManager.focusEditor()` call to fix this, so marking as a starter bug.  Or we could use this as an opportunity to finally fix post-dialog focus more generally :-)

webjocky:
n00b here, searching for my first bug to tackle. Unless I'm mistaken, the status of this issue should no longer be open since @Jerhamre seems to have added a fix that's been merged in #12824?

petetnt:
Thanks @webjocky 👍

PR #12824, opened 2016-10-10
Dialog returns focus to previous element after closing

Jerhamre:
Fixed issue #7511, #4954, #8535.

When creating a new dialog, save the currently focused element.
When closing the dialog, restore focus to previous element.


src/widgets/Dialogs.js
```diff
@@ -325,6 +328,10 @@ define(function (require, exports, module) {
             // Remove our global keydown handler.
             KeyBindingManager.removeGlobalKeydownHook(keydownHook);
 
+
+            // Restore previous focus
```
petetnt: Nit: Remove extra newline

src/widgets/Dialogs.js
```diff
@@ -325,6 +328,10 @@ define(function (require, exports, module) {
...
+            lastFocus.focus();
```
petetnt: `document.activeElement` could be null instead of something else some rare cases, so it might be good to wrap this inside a `null` check

``` js
if (lastFocus) {
  lastFocus.focus();
}
```

petetnt (changes requested):

petetnt:
Thanks for this awesome contribution @Jerhamre! Tested and it works great.

I left couple of tiny review comments, if you could fix those this is ready to be merged in 👍

Jerhamre:
Added a null check for when restoring focus and removed the extra newline

petetnt (approved):

petetnt:
Thanks for this contribution!

PR #12828, opened 2016-10-12
Fix for 6904

JoakimKLarsson:
Disabling the primary key ei "enter", when a user is typing in the search field. Enable it once a user tabs out of the search field or clicks outside it.

### DIFF
--- a/src/extensibility/ExtensionManagerDialog.js
+++ b/src/extensibility/ExtensionManagerDialog.js
@@ -44,7 +44,7 @@ define(function (require, exports, module) {
         ExtensionManagerViewModel   = require("extensibility/ExtensionManagerViewModel");
 
     var dialogTemplate    = require("text!htmlContent/extension-manager-dialog.html");
-
+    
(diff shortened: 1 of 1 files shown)


ficristo:
https://github.com/adobe/brackets/pull/7337 does a similar thing and has a test, but there was a discussion that was the wrong approach.
Have you saw that?

ficristo:
Just noticed you need to sing the CLA.
http://dev.brackets.io/brackets-contributor-license-agreement.html

petetnt:
@JoakimKLarsson thanks for contributing!

I think instead of removing/adding classes like in https://github.com/adobe/brackets/pull/7337 one could just remove the focus / add the focus to the button, right? I gave this a quick test and just canceling the enter press is not enough.

petetnt:
Superseded by #12839. However we do appreciate your contributions @JoakimKLarsson a lot and hope to see you send more PR's to address other issues 👍

PR #12829, opened 2016-10-12
Fix Issue#6822: Sidebar won't resize bigger than window size

henriknh:
#6822 Fixed Sidebar that could be resized larger than window size.


ficristo:
Have you saw https://github.com/adobe/brackets/pull/9122?
Can you comment why you prefer this approach over that before someonw look closer at this PR?

ficristo:

Issue #11801, opened 2015-10-10
Core: "Font Size" field on themes settings dialog could accept any non-ASCII characters and any number for font size.


Zhanxin:
1. Launch Brackets
2. Click View->Themes
3. Input some non-ASCII characters into the "Font Size" field or input small/large number.
   such as "ДЁЙФ龘龜び>_$#@&*ぷ龝" or 01px/1000000px

Result: "Font Size" field could accept any non-ASCII characters and any number for font size.

Expect: The "Font Size" field could add some control to forbid the incorrect format data.
![1](https://cloud.githubusercontent.com/assets/2245179/10411156/1a4b204e-6f8f-11e5-9f48-e0b505aad3c5.jpg)

jerrak0s:
If your font size input field for your theme is non-numeric, font is set to default that is 12px

MiguelCastillo:
We restrict the input when setting the value.  Perhaps we should add the pattern to the input field as well to only allow numbers and optionally the units.

https://github.com/adobe/brackets/blob/master/src/view/ViewCommandHandlers.js#L303

nyteksf:
Hello. If it's still possible, I'd like to take this issue on, please.

zaggino:
Go ahead @nyteksf , this issue still hasn't been solved.

99M8:
Hello! Has this issue been resolved?

Zhanxin:
Verified as fixed.
![forbid](https://user-images.githubusercontent.com/2245179/35957293-db445ee2-0cd5-11e8-8869-2368ce307f42.png)

PR #12830, opened 2016-10-12
Fix for #11801: "Font Size" field on themes settings dialog could accept any characters

geralexgr:
When the input field for font is not valid set the preferences in the default font size.

### DIFF
--- a/src/view/ViewCommandHandlers.js
+++ b/src/view/ViewCommandHandlers.js
@@ -301,6 +301,7 @@ define(function (require, exports, module) {
 
         // Make sure that the font size is expressed in terms we can handle (px or em). If not, simply bail.
         if (fsStyle.search(validFont) === -1) {
+        	prefs.set("fontSize", DEFAULT_FONT_SIZE + "px");
             return false;
         }
 


ficristo (changes requested):
This doesn't seem the right fix.
As suggested by @MiguelCastillo in the opening issue, have you tryed to add the pattern attribute to the html?
Something like `pattern="^[\d\.]+(px|em)$"`

petetnt:
Thanks for contributing to Brackets @jerrak0s!

I think the pattern attribute could work, not sure if there's a premade size validation regexp though but the example @ficristo gave should be close enough after adding `rem`s and so on. 

@jerrak0s you can see the documentation on `pattern` here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-pattern

jerrak0s:
My intention was to fix the wrong value in the `textfield` Font Size: after wrong input.
So when i save something wrong in the font Size, that is not saved.
capture

petetnt:
@jerrak0s Yup, and it's a good thing to do! With the `pattern` attribute however, you can ensure that users cannot write invalid values at all to the field 👍

ficristo:
This needs more work. Closing for now.
The issue #11801 is still open.

Issue #12766, opened 2016-09-09
Allow setting default file type for files without extension


CupOfTea696:
### Prerequisites
- [x] Can you reproduce the problem with `Debug \ Reload Without Extensions`?
- [x] Did you perform a cursory search to see if your bug or enhancement is already reported?
- [x] Did you read the [Troubleshooting guide](https://github.com/adobe/brackets/wiki/Troubleshooting)?
### Description

Setting the default language for files with no file extension has no effect.
### Steps to Reproduce
1. Open file with no file extension
2. Click file-type selector in bottom bar (says "Text" by default for unknown file types)
3. Select a different file-type (e.g. "Bash")
4. Reload Brackets or open another file with no extension

**Expected behavior:** The file-type is Bash and it is syntax highlighting the content

**Actual behavior:** The file-type is Text and there is no syntax highlighting
### Versions

OS X Yosemite
Release 1.7 build 1.7.0-16898 (release b0a363b71)

petetnt:
Dupe of #12655 but this one is more meaty so I'll just close the old one :+1:

petetnt:
I made some work on this a while ago but didn't have to finish it yet. See this commit: https://github.com/petetnt/brackets/commit/b069d3ea9ce85f7cb8af62a66530b3c2402e69f0

Things that are missing: 

- Following file renames and deletes
- Unit tests

alanbares:
Well been wanting to help with Brackets since it's my usual IDE but not sure if if I can handle this issue. I'll try and give it a shot anyway.

mpettaw2:
I would love to work on this issue as well if that's ok.

petetnt:
@mpettaw2 of course! @alanbares have you made much progress? Could you perhaps coordinate together? 👍

alanbares:
@petetnt Not really, I feel completely lost tackling the issue :(.

ficristo:
@petetnt do we need to land @sprintr PR first?

petetnt:
@ficristo maybe not *need* for this feature,, but it would be a great addition on it's own and would be a good starting point for this feature too.

mpettaw2:
so i found the getFilenameWithoutExtension function. i was going to change that instead of the getFileExtension because I didn't want change the file before it got written to the program. anyway i was thinking of something like 

```
function getFilenameWithoutExtension(filename) {
        var index = filename.lastIndexOf(".");
        return index===-1 ?filename+".bat":filename.slice(0, index);
        
[...]

PR #12837, opened 2016-10-17
Bugfix: Check if an empty extension rule is defined

haslam22:
#12766 -> Update getLanguageForPath() to take empty extensions into account

### DIFF
--- a/src/extensions/samples/InlineImageViewer/InlineImageViewer.html
+++ b/src/extensions/samples/InlineImageViewer/InlineImageViewer.html
@@ -1,4 +1,4 @@
-<div>
-    <div class="filename"><span/></div>
-    <img style="display: block; margin: 20px auto; opacity: 0;"/>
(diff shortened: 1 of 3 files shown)


haslam22:
The build failed because of API limits =/ Is there a way I can resend the request?

I updated getLanguageForPath() in LanguageManager to check if there's an empty ("") extension in the file extension languages map

So the getLanguageForPath() method currently goes something like:

-> Split the file name into parts
-> Check if the file name has a leading dot, remove it
-> Remove the first part of the fileName, which is assumed to be the title

<added part>
-> If there's no parts remaining, we can conclude that there was only a title and that it has no extension
-> try language = _fileExtensionToLanguageMap[""]
<end added part>

-> Continue shifting the parts until nothing is left or an extension is found

I'm not sure if I need a test for this, I'm fairly new so please let me know!

PR #12865, opened 2016-11-01
undo history not working, when line-endings change after save

zaggino:
For a bug identified in https://github.com/adobe/brackets/issues/11826#issuecomment-257429242

credits to @vahid-sanati for identifying this, put up a PR so we can have it merged asap

cc @petetnt @ficristo

### DIFF
--- a/src/editor/Editor.js
+++ b/src/editor/Editor.js
@@ -1048,7 +1048,11 @@ define(function (require, exports, module) {
      */
     Editor.prototype._resetText = function (text) {
         var currentText = this._codeMirror.getValue();
-        if (text === currentText) {
+
+        // compare with ignoring line-endings, issue #11826
+        var textLF = text ? text.replace(/(\r\n|\r|\n)/g, "\n") : null;
+        var currentTextLF = currentText ? currentText.replace(/(\r\n|\r|\n)/g, "\n") : null;
+        if (textLF === currentTextLF) {
             // there's nothing to reset
             return;
         }
--- a/test/spec/Document-test.js
+++ b/test/spec/Document-test.js
@@ -418,6 +418,24 @@ define(function (require, exports, module) {
                     expect(doc._masterEditor._codeMirror.historySize().undo).toBe(1);
                 });
             });
+
+            it("should not clean history when reset is called with the same text with different line-endings", function () {
+                runs(function () {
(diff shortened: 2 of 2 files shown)


ficristo:
The test added in https://github.com/adobe/brackets/pull/12681 probably should be updated or we should add a new one for this case.
It exists a `Document.normalizeText` which normalize line endings: maybe we should update to check also for `\r` and use it, but it doesn't check for `null` text, so maybe not.

Looking at the commits for 1.5 https://github.com/adobe/brackets/compare/release-1.4...release-1.5
there are a couple of upgrade of CodeMirror. In that version of CodeMirror it was changed a bit the way new line endings where used and maybe that broke some assumptions of Brackets.

petetnt:
Would it be worth it to extract and extend `Document.normalizeText` to [FileUtils.js](https://github.com/adobe/brackets/blob/master/src/file/FileUtils.js) with other line ending related methods? Then we could just use `FileUtils.normalizeLineEndings(text)` which would return normalized text or `null`.

The solution for the problem LGTM, just 🚲 🏠

zaggino:
I wanted to use `FileUtils.normalizeLineEndings` but when I included it in the file it lead to circular dependencies and whole Brackets broke 😒 Really hate requirejs

zaggino:
Feel free to push/modify this branch if you want, I have very little time to give to this urgently.

petetnt:
> Really hate requirejs

Ditto. I say lets merge this now and bikeshed the normalization later.

ficristo:
If you want to merge go ahead, but a test is really needed here. At least file a followup for it.

zaggino:
put up a test;

+ i believe that CM is returning always `\n` line-endings after the change @ficristo mentioned but i don't have time to investigate properly

ficristo:
AFAIK if the lineSeparator option is not set CodeMirror return always `\n` line endings.

ficristo (approved):

petetnt:
Kinda worried about possible performance issues with larger files (possibly leading to `External file changes`-dialogs?) but otherwise LGTM.

petetnt (approved):

swmitra (approved):

swmitra:
Merging.

swmitra:
Great work @vahid-sanati @zaggino @ficristo @petetnt

Issue #12864, opened 2016-11-01
folder tree not updating when creating or deleting a file


bstellar:
### Prerequisites

* [ ] Can you reproduce the problem with `Debug \ Reload Without Extensions`? yes
* [ ] Did you perform a cursory search to see if your bug or enhancement is already reported? yes
* [ ] Did you read the [Troubleshooting guide] (https://github.com/adobe/brackets/wiki/Troubleshooting)? yes

For more information on how to write a good bug report read [here](https://github.com/adobe/brackets/wiki/How-to-Report-an-Issue)
For more information on how to contribute read [here](https://github.com/adobe/brackets/blob/master/CONTRIBUTING.md)

The folder tree does not update to include changes made to the folder such as creating and deleting files.

### Steps to Reproduce

1. Add a file

**Expected behavior:** the file should be in the tree file list.

**Actual behavior:** It is not. 

### Versions

Please include the OS and what version of the OS you're running. mac osx El Capitan
Please include the version of Brackets. You can find it under `Help \ About Brackets` no I can't either. there is no about item under help menu. But, I am pretty sure it is 1.7 as it says it is up to date as latest version.

petetnt:
Hi @bstellar, can you post the errors you see under `Debug -> Show Developer Tools -> Console`?

ctants:
I have the same issue. 

Release 1.8 build 1.8.0-17108
OS X El Capitan 10.11.6

Anything else I can post here to help?

ctants:
Ok, I think I found the issue. When the project folder path has parenthesis or brackets the folder tree doesn't get updating. For example "User Folder/Projects (New)/Project1". This wasn't an issue in Brackets 1.7.

petetnt:
@ctants Interesting. I'll investigate.

ficristo:
@petetnt for the `(` case, there are some bugs on chokidar.

pelatx:
This issue remains for me in Brackets 1.14 on Windows 10, with and without extensions.

adeguntoro:
Yup, same problem with me. I add file from VSCode, but my Brackets won't update file / folder. Need to refresh first, if i reload my Brackets, it mean i can't undo Brackets.
sorry for my english.

PR #12866, opened 2016-11-01
Update ISSUE_TEMPLATE.md to cover platform differences

petetnt:
There are small differences between the platforms, as evident in https://github.com/adobe/brackets/issues/12864. This PR fixes that.

Could someone check that the Linux version has the menu under Help, as I don't have a GUI machine at hand.


ficristo:
Verified Linux too, LGTM.

Issue #10255, opened 2014-12-20
[Linux] Brackets depends on obsolete libgcrypt11 package which is no longer included by default


Romane-T:
Brackets not currently installable on Jessie. Brackets has dependency on libgcrypt11, which is not available in the Jessie repositories. Jessie repositories have libgcrypt20. Brackets can be installed if user locates a copy of libgcrypt11, which can be co-installed with libgryypt20, but considering Jessie in freeze to become stable, doing this this should be unnecessary. Removing libgcrypt11 from system removes Brackets at the same time.
Romane

prksingh:
@Romane-T Hey, could you try to build the appshell on your system. 
The repo is located at: https://github.com/adobe/brackets-shell. 
The build instructions are at: https://github.com/adobe/brackets-shell/wiki/Building-Brackets-Shell
This may help us locate if the error is with the installer.

You may want to check out this PR: https://github.com/adobe/brackets-shell/pull/489 
This may fix your issue but its not ready for testing yet.

Romane-T:
On 21/01/15 21:17, Prashant Kumar Singh wrote:
>
> @Romane-T <https://github.com/Romane-T> Hey, could you try to build 
> the appshell on your system.
> The repo is located at: https://github.com/adobe/brackets-shell.
> The build instructions are at: 
> https://github.com/adobe/brackets-shell/wiki/Building-Brackets-Shell
> This may help us locate if the error is with the installer.
>
> You may want to check out this PR: adobe/brackets-shell#489 
> <https://github.com/adobe/brackets-shell/pull/489>
> This may fix your issue but its not ready for testing yet.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/adobe/brackets/issues/10255#issuecomment-70820929>.
>
Good morning

Many thanks for your reply.

Sorry, but looked at the material at the links given, and too much over 
my head. The build instructions, though they seemed sort of straight 
forward, still left me confused. The last link to the "PR" - well, 
first, I have utterly no idea what a "PR" is, and what was on that page 
made absolutely zero sense to me.

I am _not_ a programmer. All that I learned about programming was a bit 
more than a quarter a century ago. Everything has dramatically since 
then, and I have forgotten it all anyway. I use Brackets solely to write 
my HTML pages, and even then fairly simple.

I have no idea what the difference between libgcrypt11 and libgcrypt20 
is, and have no interest in finding out. That is a developer issue.

[...]

Issue #12857, opened 2016-10-27
Proposal to switch to another shell (electron)


ficristo:
The reasons of this proposal are:
- **maintenance**: is quite high, especially if we think about the features that are in electron but not in brackets, it can be only higher. Probably we will need to replace GYP with GN since CEF is switching to it
- **theme**: let's say the truth, the dark theme is only truly supported on macOS. On Windows 10 is not right and on Linux there isn't
- **features**: there are more features (native menu on Linux or submenu on all platforms for example)
- **platforms**: as it is personally I consider Linux simply not supported by Brackets. Yes it run on it but ATM is still on an older version of CEF and there are problems that are only ageing, electron supports all three major platforms, I'm quite sure it supports ARM and there are people looking at BSD
- **easier for contributors**: from what I saw the shell is an high impedance for contributors who wanted to work on it, slow reviews or problems on compile (Linux on master still doesn't compile)
- **easier for maintainer (especially outside Adobe)**: AFAIK @zaggino creates the installer for brackets-electron through Travis, this will mean we will able to do at least prereleases by ourself

The possible downsides are:
- **control**: if we will have a problem with the shell we will need to cooperate/rely on the electron team.

Maybe there are more downsides but ATM I cannot think anything else.

If people cannot accept to loose the dark theme they can be redirected to Dreamweaver.
Last, AFAIK, Dreamweaver doesn't use this shell, so personally I feel like Brackets is the only one to suffer the burden of this shell.

I tryed to help a bit the Linux side but the effort is simply too much time consuming for me, switching to another shell seems the best choice, and since there is already [brackets-electron](https://github.com/zaggino/brackets-electron/) seems obvious what is my choice.

ficristo:
If I'm not mistaken @zaggino also replaced the sockets usage for node connection, which has always given some problems. This another reason to switch to the electron fork.

zaggino:
Yes, it uses electron's IPC and also for every node connection, new node process is spawned. This way they don't interfere with each other (sync functions in one slowing down all or one bug crashing all for example) and it feels a lot faster (didn't do measurements).

ficristo:
AFAIK the current shell scripts cannot handle patch releases, while the fork of @zaggino can.
There are issues that really deserve a patch release, think about the Ctrl-Z issue or https://github.com/adobe/brackets/pull/12145 which resulted in refreshed builds of the release.
Maybe this could be fixed in brackes-shell side, but since it is already available in the fork, why not use it? 😄

PR #12876, opened 2016-11-06
Update README.md to remove 'debian/ubuntu' #10255

jacebenson:
Changing Linux(Debian/Ubuntu) to just Linux as Debian based Linux flavors no longer install and won't continue based on comments by ficristo here https://github.com/adobe/brackets/issues/10255#issuecomment-243854796

Ultimately this has a work around but I don't see how to edit the wiki pages on adobe/brackets.  I'll be opening an issue with that as the ask which is to change the wiki pages to make it clear you can install on Debian.

I'm unsure what version of CEF is used for brackets here, but it does appear maintained if it's this one;
https://bitbucket.org/chromiumembedded/cef/

With that being said, they have a similar issue and have it as "WONTFIX"
https://bitbucket.org/chromiumembedded/cef/issues?q=libgcrypt

So pretty much unless someone finds and fixes there issue for them either the .deb will no longer work on debian os's that dont come packed with libgcrypt11 so all of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ How to install and run Brackets
 -------------------------------
 #### Download
 
-Installers for the latest stable build for Mac, Windows and Linux (Debian/Ubuntu) can be [downloaded here](http://brackets.io/).
+Installers for the latest stable build for Mac, Windows and Linux can be [downloaded here](http://brackets.io/).
 
 The Linux version has most of the features of the Mac and Windows versions, but
 is still missing a few things. See the [Linux wiki page](https://github.com/adobe/brackets/wiki/Linux-Version)


petetnt:
Hi @jacebenson,

could you sign our CLA? See http://dev.brackets.io/brackets-contributor-license-agreement.html

jacebenson:
Signed just now.

ficristo:
> Signed just now.

Thank you.

But personally I'm for WONTFIXing this. 
I think the "Debian/Ubuntu" is a way to say that we produce only .deb packages so other platforms are not supported. If you remove it, it will seems that we support more platforms which is not correct.

Linux builds are build from the `linux-1547` branch which use the 1547 version of CEF.
To fix the libgcrypto we have to update CEF version and probably merge this one https://github.com/adobe/brackets-shell/pull/501

PS: I have some PR on brackets-shell, which are already available in the repo, to build with a newer version of CEF, but we need support from Adobe to reach a conclusion. And sincerly with brackets-electron available I would like to simply switch to it: https://github.com/adobe/brackets/issues/12857.

jacebenson:
@ficristo changing to electron would be cool but I was under the impression that wasn't going to happen per your comment before #10255 was locked.  If it did that'd be awesome.  I've been playing with atom as I jump around from windows to linux and like the same editor.  I'll keep watching for this change but I am not holding my breath.

petetnt:
@jacebenson have you checked out zaggino/brackets-electron? It has (near) feature parity with the CEF version and it's maintained by one of our contributors. I think that switching at this point wouldn't be so farfetched.

ficristo:
I don't know if we will switch to electron officially (I'm totally in favor for it), but the fork of @zaggino is very well alive. And, especially if you are a Linux user, I strongly advice to switch to it.

Regarding this PR, I think it will make things a bit more confusing, so I close it for now.

PR #12885, opened 2016-11-09
improvements to file system events processing for search

zaggino:
Hi guys, I've reported elsewhere that my db dump script crashes both brackets (white-screen) and also brackets-electron. This is the work done to fix that (and I can confirm that with this fix there's no crash when running my script anymore) - it's not final and I'd like to have it reviewed before working on little things as docs and similar.

There're two main things, first aggregate duplicate events and debounce for a short while before processing. My script quickly deletes and creates back a lot of stuff, before this change some of the directories were removed and added back to search multiple times. With debouncing it's only once after script is done.

Second, no less important thing is fixing calls to `filesChanged` and `filesRemoved` functions. In case of my script these two were called about 1000 times with an array of size 1 which cause a lot of trouble for IPC connection in electron and socket connection in regular brackets. With the fix these are now properly called.

Marking this as high-priority because it fixes brackets completely crashing in some cases.

### DIFF
--- a/src/search/FindInFiles.js
+++ b/src/search/FindInFiles.js
@@ -830,8 +830,6 @@ define(function (require, exports, module) {
 
         var addPromise;
         if (entry.isDirectory) {
-            added = added || [];
-            removed = removed || [];
             if (added.length === 0 && removed.length === 0) {
                 // If the added or removed sets are null, must redo the search for the entire subtree - we
                 // don't know which child files/folders may have been added or removed.
@@ -872,12 +870,9 @@ define(function (require, exports, module) {
         // we need to reduce _cachedFileSystemEvents not to contain duplicates!
         _cachedFileSystemEvents = _cachedFileSystemEvents.reduce(function (result, obj) {
             var fullPath = obj.entry ? obj.entry.fullPath : null;
-            // merge added
-            if (result[fullPath] && obj.added) {
+            // merge added & removed
+            if (result[fullPath] && obj.isDirectory) {
                 obj.added = obj.added.concat(result[fullPath].added);
-            }
-            // merge removed
-            if (result[fullPath] && obj.removed) {
                 obj.removed = obj.removed.concat(result[fullPath].removed);
             }
             // use the latest event as base
@@ -895,9 +890,17 @@ define(function (require, exports, module) {
      * putting them to cache and executing a debounced function
      */
     _debouncedFileSystemChangeHandler = function (event, entry, added, removed) {
+        // normalize this here so we don't need to handle null later
+        var isDirectory = false;
+        if (entry && entry.isDirectory) {
+            isDirectory = true;
+            added = added || [];
+            removed = removed || [];
+        }
         _cachedFileSystemEvents.push({
             event: event,
             entry: entry,
+            isDirectory: isDirectory,
             added: added,
             removed: removed
         });
--- a/src/search/FindInFiles.js
+++ b/src/search/FindInFiles.js
@@ -744,7 +744,7 @@ define(function (require, exports, module) {
 
         /*
          * Remove existing search results that match the given entry's path
-         * @param {(File|Directory)} entry
+         * @param {Array.<(File|Directory)>} entries
          */
         function _removeSearchResultsForEntries(entries) {
             var fullPaths = [];
@@ -766,8 +766,8 @@ define(function (require, exports, module) {
         }
 
         /*
-         * Add new search results for this entry and all of its children
-         * @param {(File|Directory)} entry
+         * Add new search results for these entries and all of its children
+         * @param {Array.<(File|Directory)>} entries
          * @return {jQuery.Promise} Resolves when the results have been added
          */
         function _addSearchResultsForEntries(entries) {
@@ -859,8 +859,15 @@ define(function (require, exports, module) {
         });
     };
     
+    /**
+     * This stores file system events emitted by watchers that were not yet processed
+     */
     var _cachedFileSystemEvents = [];
     
+    /**
+     * Debounced function to process emitted file system events
+     * for cases when there's a lot of fs events emitted in a very short period of time
+     */
     _processCachedFileSystemEvents = _.debounce(function () {
         // we need to reduce _cachedFileSystemEvents not to contain duplicates!
         _cachedFileSystemEvents = _cachedFileSystemEvents.reduce(function (result, obj) {
@@ -883,6 +890,10 @@ define(function (require, exports, module) {
         _cachedFileSystemEvents = [];
     }, FILE_SYSTEM_EVENT_DEBOUNCE_TIME);
     
+    /**
+     * Wrapper function for _fileSystemChangeHandler which handles all incoming fs events
+     * putting them to cache and executing a debounced function
+     */
     _debouncedFileSystemChangeHandler = function (event, entry, added, removed) {
         _cachedFileSystemEvents.push({
             event: event,
--- a/src/search/FindInFiles.js
+++ b/src/search/FindInFiles.js
@@ -80,8 +80,12 @@ define(function (require, exports, module) {
         _fileNameChangeHandler,
         clearSearch;
     
-    /* Waits for FS changes to stack up until processing them (scripts like npm install can do a lot of movements on the disk) */
-    var FILE_SYSTEM_EVENT_DEBOUNCE_TIME = 1000;
+    /**
+     * Waits for FS changes to stack up until processing them
+     * (scripts like npm install can do a lot of movements on the disk)
+     * @const
+     */
+    var FILE_SYSTEM_EVENT_DEBOUNCE_TIME = 100;
 
     /** Remove the listeners that were tracking potential search result changes */
     function _removeListeners() {
@@ -665,7 +669,7 @@ define(function (require, exports, module) {
      * Inform node that the list of files has changed.
      * @param {array} fileList The list of files that changed.
      */
-    function filesChanged(fileList) {        
+    function filesChanged(fileList) {
         if (FindUtils.isNodeSearchDisabled() || !fileList || fileList.length === 0) {
             return;
         }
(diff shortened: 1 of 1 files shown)


src/search/FindInFiles.js
```diff
@@ -657,8 +665,8 @@ define(function (require, exports, module) {
      * Inform node that the list of files has changed.
      * @param {array} fileList The list of files that changed.
      */
-    function filesChanged(fileList) {
-        if (FindUtils.isNodeSearchDisabled() || fileList.length === 0) {
+    function filesChanged(fileList) {
```
petetnt: Some trailing whitespace here

src/search/FindInFiles.js
```diff
@@ -657,8 +665,8 @@ define(function (require, exports, module) {
      * Inform node that the list of files has changed.
      * @param {array} fileList The list of files that changed.
      */
-    function filesChanged(fileList) {
-        if (FindUtils.isNodeSearchDisabled() || fileList.length === 0) {
+    function filesChanged(fileList) {        
+        if (FindUtils.isNodeSearchDisabled() || !fileList || fileList.length === 0) {
```
petetnt: `!fileList` covers `fileList.length === 0` too.

src/search/FindInFiles.js
```diff
@@ -734,67 +742,81 @@ define(function (require, exports, module) {
...
+        function _removeSearchResultsForEntries(entries) {
+            var fullPaths = [];
+            entries.forEach(function (entry) {
+                Object.keys(searchModel.results).forEach(function (fullPath) {
+                    if (fullPath === entry.fullPath ||
+                            (entry.isDirectory && fullPath.indexOf(entry.fullPath) === 0)) {
+                        // node search : inform node that the file is removed
+                        fullPaths.push(fullPath);
+                        if (findOrReplaceInProgress) {
+                            searchModel.removeResults(fullPath);
```
petetnt: Should this be called once with a large array too?

src/search/FindInFiles.js
```diff
@@ -734,67 +742,81 @@ define(function (require, exports, module) {
...
-                if (fullPath === entry.fullPath ||
-                        (entry.isDirectory && fullPath.indexOf(entry.fullPath) === 0)) {
-                    // node search : inform node that the file is removed
-                    filesRemoved([fullPath]);
-                    if (findOrReplaceInProgress) {
-                        searchModel.removeResults(fullPath);
-                        resultsChanged = true;
+        function _removeSearchResultsForEntries(entries) {
+            var fullPaths = [];
+            entries.forEach(function (entry) {
```
petetnt: Might be worth checking out if couple of `Array.prototype.map`s would be faster than nested `forEach`s and a push

petetnt (commented):
Added some comments.

src/search/FindInFiles.js
zaggino: no it doesn't
```
var emptyArray = []; var a = !emptyArray; var b = emptyArray.length === 0; [a,b];
[false, true]
```

src/search/FindInFiles.js
zaggino: `fullPaths.push(fullPath);` is inside an `if` so if map would be used, we'd need to filter it out for nulls afterwards

src/search/FindInFiles.js
petetnt: Ugh, sorry about that @zaggino, I automatically assumed that `fileList` would never be an empty array just undefined. Again, sorry.

src/search/FindInFiles.js
petetnt: Valid point 👍

src/search/FindInFiles.js
zaggino: `removeResults` is a classic sync method, so it wouldn't do any benefit here

petetnt:
Should avoid doing code reviews when I am tired 😴 👍 

Nice work @zaggino

src/search/FindInFiles.js
```diff
@@ -73,12 +73,20 @@ define(function (require, exports, module) {
...
-    var _documentChangeHandler, _fileSystemChangeHandler, _fileNameChangeHandler, clearSearch;
+    var _documentChangeHandler,
+        _fileSystemChangeHandler,
+        _processCachedFileSystemEvents,
+        _debouncedFileSystemChangeHandler,
+        _fileNameChangeHandler,
+        clearSearch;
+    
+    /* Waits for FS changes to stack up until processing them (scripts like npm install can do a lot of movements on the disk) */
+    var FILE_SYSTEM_EVENT_DEBOUNCE_TIME = 1000;
```
ficristo: Can this be something like 100ms instead? Or is too low?

src/search/FindInFiles.js
```diff
@@ -734,67 +742,81 @@ define(function (require, exports, module) {
...
-            var visitor = function (child) {
-                // Replicate filtering that getAllFiles() does
-                if (ProjectManager.shouldShow(child)) {
-                    if (child.isFile && _isReadableText(child.name)) {
-                        // Re-check the filtering that the initial search applied
-                        if (_inSearchScope(child)) {
-                            addedFiles.push(child);
-                            addedFilePaths.push(child.fullPath);
+        function _addSearchResultsForEntries(entries) {
+            var fullPaths = [];
```
ficristo: Do you have a rough idea of the memory usage here? If we store a lot of path could casue some trouble?

ficristo:
> Nice work @zaggino

👍

src/search/FindInFiles.js
raita: zaggino@ Bro how are you today? Please Bro help us we want CoffeeScript in Brakets and also  
Babel , and many's other please

src/search/FindInFiles.js
zaggino: You're right here, re-tested this set to 100ms with my script and it's fine. So setting it to 100ms.

src/search/FindInFiles.js
zaggino: 👍

src/search/FindInFiles.js
zaggino: Leaving it as is, it's a very cheap check and better be safe than sorry :)

src/search/FindInFiles.js
zaggino: Well, this array is not being persisted besides the function execution lifetime and I wouldn't expect there to be a problem even if you have thousands of entries in it. On the other hand, without it, trying to send thousands of messages to a node process will definitely crash it (or socket server). In other words, I don't think memory usage is something to worry about.

src/search/FindInFiles.js
```diff
@@ -830,6 +854,36 @@ define(function (require, exports, module) {
...
+            return result;
+        }, {});
+        _.forEach(_cachedFileSystemEvents, function (arr) {
+            _fileSystemChangeHandler.apply(null, arr);
+        });
+        _cachedFileSystemEvents = [];
+    }, FILE_SYSTEM_EVENT_DEBOUNCE_TIME);
+    
+    _debouncedFileSystemChangeHandler = function (event, entry, added, removed) {
+        _cachedFileSystemEvents.push([ event, entry, added, removed ]);
```
zaggino: Sure, can do.

zaggino:
Pushed some changes, @petetnt @ficristo please review again :-)

src/search/FindInFiles.js
```diff
@@ -830,6 +856,56 @@ define(function (require, exports, module) {
...
+                obj.removed = obj.removed.concat(result[fullPath].removed);
+            }
+            // use the latest event as base
+            result[fullPath] = obj;
+            return result;
+        }, {});
+        _.forEach(_cachedFileSystemEvents, function (obj) {
+            _fileSystemChangeHandler(obj.event, obj.entry, obj.added, obj.removed);
+        });
+        _cachedFileSystemEvents = [];
```
zaggino: How exactly? There's a `_.forEach` above this that launches `_fileSystemChangeHandler` for every object in the array, before deleting it. The function is completely synchronous - if an fs event is emitted while this is running, it'll be added to the empty array once this function finishes.

src/search/FindInFiles.js
```diff
@@ -830,6 +856,56 @@ define(function (require, exports, module) {
...
+    var _cachedFileSystemEvents = [];
+    
+    /**
+     * Debounced function to process emitted file system events
+     * for cases when there's a lot of fs events emitted in a very short period of time
+     */
+    _processCachedFileSystemEvents = _.debounce(function () {
+        // we need to reduce _cachedFileSystemEvents not to contain duplicates!
+        _cachedFileSystemEvents = _cachedFileSystemEvents.reduce(function (result, obj) {
+            var fullPath = obj.entry ? obj.entry.fullPath : null;
```
ficristo: If `fullPath` is null maybe return immediately?

src/search/FindInFiles.js
ficristo: Probably I was overthinking, but I was wondering if there could be some races between `_debouncedFileSystemChangeHandler` and `_processCachedFileSystemEvents`.

src/search/FindInFiles.js
zaggino: Deeper in the code there's a handle for this case (doing a full refresh I think) so it didn't want to stop it here.

src/search/FindInFiles.js
zaggino: No, races certainly can't happen in this case. They are both completely sync functions - no promises or async-await there.

src/search/FindInFiles.js
ficristo: So the assignment below is safe: `result[fullPath] = obj;` ?

src/search/FindInFiles.js
zaggino: `debounce` just wraps a function and keeps postponing its execution until the wrapper stops to be called for the timeout time. But once the wrapped function is executed, all happens totally sync so the code is safe. I've used this pattern a few times before.

ficristo (approved):

zaggino:
@petetnt anything else? if not it should be good to merge

petetnt:
@zaggino LGTM, I'll run this in production use for today and merge it in if nothing pops up.

swmitra (approved):
LGTM

petetnt (approved):

petetnt:
Used it all day without any issues and Brackets did feel snappier. Great job once again @zaggino!

Issue #12898, opened 2016-11-11
Brackets 1.8 - Installing extensions from Extension Manager not working 


nerk:
### Prerequisites

* [x] Can you reproduce the problem with `Debug -> Reload Without Extensions`?
* [x] Did you perform a cursory search to see if your bug or enhancement is already reported?
* [x] Did you read the [Troubleshooting guide](https://github.com/adobe/brackets/wiki/Troubleshooting)?

For more information on how to write a good bug report read [here](https://github.com/adobe/brackets/wiki/How-to-Report-an-Issue)
For more information on how to contribute read [here](https://github.com/adobe/brackets/blob/master/CONTRIBUTING.md)

### Description

Installing any extensions from Extension Manager fails.

### Steps to Reproduce

1. Open Extension Manager
2. Install any extension.

**Expected behavior:** Extension gets installed

**Actual behavior:** Message box with "Installation failed - Unknown internal error" appears. 
Debug console shows "WebSocket connection to 'ws://localhost:51378/' failed: Error in connection establishment: net::ERR_CONNECTION_REFUSED /utils/NodeConnection.js:84"

If I enable "Node Debugger", the regular "Installing extension from https://..." message is posted, but it just hangs until I cancel it manually.

As a side node: "Reload Without Extensions" does not appear to do anything. I therefore removed the extensions directory manually just in case, but it does not help either.

### Versions

Windows 7 Enterprise 64-bit, Service Pack 1 
Release 1.8 build 1.8.0-17108

nerk:
Forgot to mention that I'm behind a proxy. Proxy is properly configured and list of extensions appear in Extension Manager.

fuecisla:
Same thing here. No way to install extensions with the extensions manager

Madebym:
Working on Win7 64bit, not working on Win7 32bit !!

nerk:
I'm on Windows 64-bit and it does not work. I would actually assume that it would be the other way round since the Brackets msi installer installs the 32-bit version on a 64-bit system.

Madebym:
Funny, it is the other way around on my end. 

Laptop is running Win10, all fine. Desktop is Win7, can't install. 

The only difference is that I had 1.8 pre-release installed on desktop.

MD-Nayeem:
Not working also for me win 32bit

nerk:
The problem seems to be related to the  proxy configuration:

[...]

PR #12933, opened 2016-11-22
Zaggino/update extension manager deps

zaggino:
Replacement for https://github.com/adobe/brackets/pull/10372 to keep from commiting `node_modules` to git as we should really abandon that practice where possible.

Related https://github.com/adobe/brackets/issues/12898

cc @ficristo 

Note: the webpack approach for semver is taken from brackets-electron, where it's implemented and works well if dependencies (in this case semver) don't provide amd build that we need to use in the browser code

Note2: sorry the diff is quite hard to read because of all the deletes, reading through commits might be easier


ficristo:
I would prefer if someone more familiar with webpack could take a look.
Maybe someone at Adobe?

zaggino:
@ficristo this is now up to date with master after https://github.com/adobe/brackets/pull/12972 merge. Not sure if anyone from Adobe will have time to look at this anytime soon.

ficristo:
I'm sorry, but I am really not confident in reviewing this.
Maybe @petetnt or @MiguelCastillo can?

ficristo:
`src/extensibility/node` is exported as a zip for the registry (see `brackets-registry/deps`), so it seems to me it needs to be a standalone project or at least we need some way to export the zip: from what I can tell this breaks it.
@zaggino do you know how the zip is created? Or maybe @ingorichter or @redmunds.

redmunds:
I believe that src/extensibility is used by both Brackets and the Brackets Extension Registry, but I don't know the process from getting code onto the registry. @ingorichter Do you know?

ingorichter:
we need to tar/gzip all the js files in `brackets/src/extensibility/node/` to create the archive for the registry. Then someone with permissions needs to update the registry. I don't have permissions anymore to the server. Before that update, we need to make sure that the changes work locally before updating the live system. I once ran into issues on that system, since the node version might be pretty outdated again and therefore required node_modules might not like it.

zaggino:
@swmitra can someone at @adobe take care of the update for https://github.com/adobe/brackets-registry or can I get access to that machine and work on it? we need to work on getting this merged so we can fix bugs related to brackets not working properly behind a proxy server

zaggino:
updated this PR to be merge-able

swmitra:
@zaggino We are looking into the registry server node upgrade request. Should be able to give you some concrete data by end of this week.

saurabh95:
@zaggino Since the server has upgraded node version, Could you resolve conflicts in this PR?

zaggino:
@saurabh95 Conflicts resolved.

swmitra:
@zaggino Not yet 😄 . Probably becuase I have merged @ficristo PR for write config grunt task. Can you please merge that one more.

zaggino:
@swmitra Done again :-)

swmitra:
Merging this PR. Build failure is because of cla check pull failure.

swmitra:
Thank a ton @zaggino for bringing this change. It makes dependency management a lot more trivial. Great work as always 👍

swmitra (commented):
Good to merge as registry server is also upgraded to latest node.

Issue #12937, opened 2016-11-23
CodeMirror update


zaggino:
@adobe/brackets-committers 

Latest master of CodeMirror doesn't contain built `codemirror.js` in the lib directory: https://github.com/codemirror/CodeMirror/tree/master/lib

Since we'll definitely want to upgrade for next version there are two options:

a) add a task to our grunt scripts that'll build `codemirror.js` by calling `npm run build` in the CodeMirror submodule
b) use CodeMirror as npm dependency

Let me know which solution you prefer.

petetnt:
Voting for b)

swmitra:
My vote for 'b'.
Off topic - We should do the same even for 'tern'.

zaggino:
@petetnt @swmitra 
Please also have a look at related https://github.com/adobe/brackets/pull/12933
confirmed to work in https://github.com/adobe/brackets/issues/12898#issuecomment-262322859

zaggino:
PR here https://github.com/adobe/brackets/pull/12940

ficristo:
Done in https://github.com/adobe/brackets/pull/12972
@swmitra Acorn and Tern are already installed from npm, see https://github.com/adobe/brackets/pull/11569
It would be nice to also look at https://github.com/adobe/brackets/issues/12744

PR #12940, opened 2016-11-23
CodeMirror update

zaggino:
PR for https://github.com/adobe/brackets/issues/12937

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -323,7 +323,7 @@ module.exports = function (grunt) {
     });
 
     // task: install
-    grunt.registerTask('install', ['write-config', 'less', 'npm-install-extensions']);
+    grunt.registerTask('install', ['write-config', 'less', 'copy-browser-dependencies', 'npm-install-extensions']);
 
     // task: test
     grunt.registerTask('test', ['eslint', 'jasmine', 'nls-check']);
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,6 @@ Thumbs.db
 # Files that can be automatically downloaded that we don't want to ship with our builds
 /src/extensibility/node/node_modules/request/tests/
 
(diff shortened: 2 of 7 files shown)


ficristo:
Why not use requirejs to create some logical mapping between the node_modules and thirdparty folders?
I think @petetnt has done that in https://github.com/adobe/brackets/issues/12006
If we go on this route we probably need to do something for distribution. (I don't remember exactly but I think we don't copy the node_modules over there)

zaggino:
Not sure how to do any logical mapping here, but happy to learn. We could use a separate package.json file inside the src folder for production dependencies. That's how brackets-electron does it and that's how electron authors recommend electron projects to be done.

ficristo:
The other idea I had in mind was exactly putting a package.json inside the src folder.
Both cases seems better to me.
For the logical mapping @petetnt know more.

petetnt:
One can use relative mapping with `../node_modules/...` for the `requirejs` config so you can use deps from the root folder: https://github.com/petetnt/brackets/blob/petetnt/poc-thirdparty-with-npm/src/main.js#L68. Not sure if that works in browser though.

Gruntfile.js
```diff
@@ -323,7 +323,7 @@ module.exports = function (grunt) {
     });
 
     // task: install
-    grunt.registerTask('install', ['write-config', 'less', 'npm-install-extensions']);
+    grunt.registerTask('install', ['write-config', 'less', 'copy-browser-dependencies', 'npm-install-extensions']);
```
le717: Fast note: the script is referenced here as `copy-browser-dependencies` but is actually named `move-browser-dependencies`.

zaggino:
Closing in favor of https://github.com/adobe/brackets/pull/12972

Issue #12441, opened 2016-05-20
Brackets turns into a web browser


dangowans:
While browsing the web in Firefox 46.0.1, I accidentally clicked and dragged a link from a webpage into Brackets.

When I let go of the link in Brackets, the editor disappeared without warning, and the window became a web browser.

Dragging this [Google Link](http://www.google.com) into the Brackets window has the same behaviour.

There's no consistent way to get back to the editor without closing Brackets completely and reopening it.  Any unsaved changes before the click-and-drag event are lost.  Menu options stop working too.

![Google Inside of Brackets](https://cloud.githubusercontent.com/assets/19495149/15437111/4ea7b880-1e93-11e6-940f-cc42f52ca76e.png)

I'm running Brackets 1.6 on Windows 7.  

(The same thing happens from Google Chrome 50.)

petetnt:
Hi @dangowans,

this issue is a duplicate of #11970 (and some other older ones). Due to the nature of the component that backs Brackets (CEF) this thing is bound to happen, but the fix should be relatively simple. If you feel like putting in a PR I can guide you through the steps (or I'll try to get around to it later myself).

dangowans:
Honestly, I only created an account here to submit this bug.  I would have to do a whole lot of reading up on git before I'd totally understand what a PR is, let alone be comfortable "putting in a PR".

petetnt:
@dangowans That's alright! Thanks for reporting the issue! 👍

wasifhyder:
Hey. I'd like to work on this. Can I?

petetnt:
Hi @wasifhyder,

of course! If this is your first time contributing to Brackets, check out the guide on [How to hack on Brackets first](https://github.com/adobe/brackets/wiki/How-to-Hack-on-Brackets) first, which outlines all the the steps you need from getting started to putting up an PR. 

If you have any questions (about the process or the issue), just post them here! Have fun 👍

jamran7:
Hi,
[...]

PR #12946, opened 2016-11-26
Browser issue

jamran7:
The solution code is in the commit: Browser issue fix

This is a bug fix for issue:(Brackets turns into a web browser #12441). 

Description: Filters out drag/drop events from text strings being dropped into Brackets that will actually change the DOM structure of the Brackets layout. It looks for a “text/url-list” type in the types list and stops the event propagation if there are any.

All unit tests and smoke tests were performed and passed.

Here is a test that verifies that the solution works: 
Settings should be as they are in a smoke test

1. Create an empty folder externally.
2. Open the folder with Brackets.
3. Enable "dragDropText" in Brackets preferences.
4. Drag a text or HTML file into Brackets.
5. The file should be open and focused in the workspace.
6. Open up a browser.
7. Type in any valid URL (a webpage with text) and go there in the browser.
8. Add a few lines to the test file that was opened in Brackets.
9. Highlight some text from the webpage and copy it to a line in the test file.
10. Verify that the text in Brackets can be highlighted and dragged to any other line.
11. Drag the URL from the Browser’s URL field into Brackets. Do this several times over all areas of the Brackets editor.
12. Verify that the URL did not drop into Brackets, or Brackets refuses the URL, and that Brackets remains an editor and not a browsable webpage. Failed test: If Brackets does become a webpage restart Brackets and perform steps 1 through 5.

13. Set the view to Vertical Split.
14. Open another test file and verify steps 1 through 5 in the other non-used split view.
15. There should be two test files open, one in each view. Perform steps 6 through 12 for both split editors.

16. Perform steps 13 through 15 for a Horizontal Split view

My repo: https://github.com/jamran7/brackets

This is on behalf of ChainsawFist

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -20,7 +20,7 @@
         "extension_url": "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default": true,
         "build_timestamp": "",
-        "healthDataServerURL": "https://healthdev.brackets.io/healthDataLog"
+        "healthDataServerURL": "https://health.brackets.io/healthDataLog"
     },
     "name": "Brackets",
     "version": "1.9.0-0",
--- a/src/config.json
+++ b/src/config.json
@@ -20,7 +20,7 @@
         "extension_url": "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default": true,
         "build_timestamp": "",
-        "healthDataServerURL": "https://health.brackets.io/healthDataLog"
+        "healthDataServerURL": "https://healthdev.brackets.io/healthDataLog"
     },
     "name": "Brackets",
     "version": "1.9.0-0",
--- a/src/utils/DragAndDrop.js
+++ b/src/utils/DragAndDrop.js
@@ -72,9 +72,13 @@ define(function (require, exports, module) {
     function stopURIListPropagation(files, event) {
         var types = event.dataTransfer.types;
             
-        if ((!files || !files.length) && types) {// stop default behavior if a url is dragged in so the browser does not takeove
+        if ((!files || !files.length) && types) { // We only want to check if a string of text was dragged into the editor
             types.forEach(function (value) {
-                if (value === "text/uri-list") { //plain text just has text/html
+                //Draging text externally (dragging text from another file): types has "text/plain" and "text/html"
+                //Draging text internally (dragging text to another line): types has just "text/plain"
+                //Draging a file: types has "Files"
+                //Draging a url: types has "text/plain" and "text/uri-list" <-what we are interested in
+                if (value === "text/uri-list") { 
                     event.stopPropagation();
                     event.preventDefault();
                     return;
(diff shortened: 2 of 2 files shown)


src/config.json
```diff
@@ -20,7 +20,7 @@
         "extension_url": "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default": true,
         "build_timestamp": "",
-        "healthDataServerURL": "https://healthdev.brackets.io/healthDataLog"
+        "healthDataServerURL": "https://health.brackets.io/healthDataLog"
```
petetnt: Unneeded change here, there's a fix coming for this file getting changed automatically hopefully soon but needs to be removed from this PR. 😕

src/utils/DragAndDrop.js
```diff
@@ -156,6 +156,18 @@ define(function (require, exports, module) {
             event = event.originalEvent || event;
 
             var files = event.dataTransfer.files;
+
+            var types = event.dataTransfer.types;
+            if(!files.length){// stop default behavior if a url is dragged in so the browser does not takeove
```
petetnt: Can `files` also be `null`? The file handler below seems to suggest that, so it might be worth guarding against it (`if(!files || !files.length)`.

src/utils/DragAndDrop.js
```diff
@@ -156,6 +156,18 @@ define(function (require, exports, module) {
...
+                types.forEach(function (value){
```
petetnt: Can `types` be null?

src/utils/DragAndDrop.js
```diff
@@ -174,6 +186,18 @@ define(function (require, exports, module) {
             event = event.originalEvent || event;
 
             var files = event.dataTransfer.files;
+
+            var types = event.dataTransfer.types;
```
petetnt: Might be worth it to move these blocks into a function call such as `stopURIListPropagation()` or similar instead of repeating the blocks of code.

petetnt (changes requested):
Reviewed, nice work ChainsawFist and @jamran7! 

If ChainsawFist is the original author as the PR comment points out, I think s/he should also sign the CLA.

jamran7:
ChainsawFist is the organization that I am representing in this PR; it was created as part of a college course project. We have a repo of Brackets and we have been using it to test our bug fixes. I have a cleaner repo of the latest commit, while the one with ChainsawFist is far behind and has multiple changes; we are using the later repo as "show-and-tell" so our professor can keep track of our progress. Right now I'm incorporating fixes that have not been closed by adobe/brackets as part of our assignment. I suppose I could create another repo in ChainsawFist that way there's no confusion and I don't have to keep mentioning the organization. I will make the changes as you suggested. Thank you.

petetnt:
Thanks for the clarification @jamran7!

jamran7:
Can you rerun that build. api.github.com gave Travis a 403.

jamran7:
I've fixed the indent in the function as well as cleaned up some of the comments.

petetnt:
This looks great, I'll do another round of testing and then merge this in if all is good 👍

petetnt:
Does this PR fix #12961 too? If it doesn't yet, I think it could be trivially added too, right?

jamran7:
Yes that issue seems to be fixed by this.

petetnt (approved):

petetnt:
Thanks for this great PR @jamran7 & ChainsawFist 👍

PR #12949, opened 2016-11-27
#7276 Live Preview highlight customization 

Worie:
**Live Preview highlight customization (colors of highlight can be customized via user preferences)**

I needed to easly customize the highlight color of Live Preview - and I've achieved it with the changes in the PR. I'm not sure if that's the desired way of doing so, as this is my first PR to Brackets.

1. RemoteFunctions takes whole config as a parameter instead just "experimental" flag.
2. There's a preference in which user can define desired styling of highlightning. 
3. The view is updated when preferences change without breaking live dev session.
4. If no user preferences are defined, the default configuration is used.

It also implements #7276 -> [Preview of Live Preview with margins and paddings showing](http://imgur.com/Gx1r4P3)

### DIFF
--- a/src/LiveDevelopment/Agents/RemoteAgent.js
+++ b/src/LiveDevelopment/Agents/RemoteAgent.js
@@ -131,7 +131,7 @@ define(function RemoteAgent(require, exports, module) {
         _stopKeepAliveInterval();
 
         // inject RemoteFunctions
-        var command = "window._LD=" + RemoteFunctions + "(" + LiveDevelopment.config.experimental + "," + PreferencesManager.get("livedev.wsPort") + ");";
+        var command = "window._LD=" + RemoteFunctions + "(" + JSON.stringify(LiveDevelopment.config) + "," + PreferencesManager.get("livedev.wsPort") + ");";
 
         Inspector.Runtime.evaluate(command, function onEvaluate(response) {
             if (response.error || response.wasThrown) {
--- a/src/LiveDevelopment/Agents/RemoteFunctions.js
+++ b/src/LiveDevelopment/Agents/RemoteFunctions.js
@@ -262,6 +262,180 @@ function RemoteFunctions(config, remoteWSPort) {
                 return;
             }
 
+            var showMarginPadding;
+            if (config.remoteHighlight.showPaddingMargin) {
+                showMarginPadding = 'block';
+            } else {
+                showMarginPadding = 'none';
+            }
+            
+            var elementStyling = window.getComputedStyle(element, null);
+            
+            var calculateSize = function () {
+                var sum = 0;
+                var i;
+                for (i = 0; i < arguments.length; i++) {
+                    sum += parseInt(arguments[i], 10);
+                }
+                return sum;
+            };
+            
+            var mainBoxStyles = config.remoteHighlight.stylesToSet;
+            
+            var paddingVisualisations = [
+                // padding-top
+                {
+                    "height": elementStyling.getPropertyValue('padding-top'),
+                    "width": elementBounds.width + "px",
+                    "position": "absolute",
+                    "top": -calculateSize(
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "left": -calculateSize(
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "box-sizing": "border-box",
+                    "display": showMarginPadding
+                },
+                // padding-right
+                {
+                    "width": elementStyling.getPropertyValue('padding-right'),
+                    "height": elementBounds.height + "px",
+                    "position": "absolute",
+                    "top": 0,
+                    "right": -calculateSize(
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "box-sizing": "border-box",
+                    "display": showMarginPadding
+                },
+                // padding-bottom
+                {
+                    "height": elementStyling.getPropertyValue('padding-bottom'),
+                    "width": elementBounds.width + "px",
+                    "position": "absolute",
+                    "bottom": -calculateSize(
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "left": 0,
+                    "box-sizing": "border-box",
+                    "display": showMarginPadding
+                },
+                // padding-left
+                {
+                    "height": elementBounds.height + "px",
+                    "width": elementStyling.getPropertyValue('padding-left'),
+                    "position": "absolute",
+                    "top": 0,
+                    "left": -calculateSize(
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "box-sizing": "border-box",
+                    "display": showMarginPadding
+                }
+            ];
+                
+            var marginVisualisations = [
+                // margin-top
+                {
+                    "height": elementStyling.getPropertyValue('margin-top'),
+                    "width": elementBounds.width + "px",
+                    "position": "absolute",
+                    "top": -calculateSize(
+                        elementStyling.getPropertyValue('margin-top'),
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "left": -calculateSize(
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "box-sizing": "border-box",
+                    "display": showMarginPadding
+                },
+                // margin-right
+                {
+                    "width": elementStyling.getPropertyValue('margin-right'),
+                    "height":
+                        calculateSize(
+                            elementBounds.height,
+                            elementStyling.getPropertyValue('margin-top'),
+                            elementStyling.getPropertyValue('margin-bottom')
+                        ) + "px",
+                    "position": "absolute",
+                    "top": -calculateSize(
+                        elementStyling.getPropertyValue('margin-top'),
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "right": -calculateSize(
+                        elementStyling.getPropertyValue('margin-right'),
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "box-sizing": "border-box",
+                    "display": showMarginPadding
+                },
+                // margin-bottom
+                {
+                    "height": elementStyling.getPropertyValue('margin-bottom'),
+                    "width": elementBounds.width + "px",
+                    "position": "absolute",
+                    "bottom": -calculateSize(
+                        mainBoxStyles['border-width'],
+                        elementStyling.getPropertyValue('margin-bottom')
+                    ) + "px",
+                    "left": -calculateSize(
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "box-sizing": "border-box",
+                    "display": showMarginPadding
+                },
+                {
+                    "height": calculateSize(
+                        elementBounds.height,
+                        elementStyling.getPropertyValue('margin-top'),
+                        elementStyling.getPropertyValue('margin-bottom')
+                    ) + "px",
+                    "width": elementStyling.getPropertyValue('margin-left'),
+                    "position": "absolute",
+                    "top": -calculateSize(
+                        elementStyling.getPropertyValue('margin-top'),
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "left": -calculateSize(
+                        elementStyling.getPropertyValue('margin-left'),
+                        mainBoxStyles['border-width']
+                    ) + "px",
+                    "box-sizing": "border-box",
+                    "display": showMarginPadding
+                }
+            ];
+            
+            var setupVisualisations = function (arr, config) {
+                var i;
+                for (i = 0; i < arr.length; i++) {
+                    var el = window.document.createElement("div");
+                    var styles = Object.assign(
+                        {},
+                        config,
+                        arr[i]
+                    );
+
+                    _setStyleValues(
+                        styles,
+                        el.style
+                    );
+
+                    highlight.appendChild(el);
+                }
+            };
+            
+            setupVisualisations(
+                marginVisualisations,
+                config.remoteHighlight.marginStyling
+            );
+            setupVisualisations(
+                paddingVisualisations,
+                config.remoteHighlight.paddingStyling
+            );
+            
             highlight.className = HIGHLIGHT_CLASSNAME;
 
             var offset = _screenOffset(element);
--- a/src/LiveDevelopment/Agents/RemoteFunctions.js
+++ b/src/LiveDevelopment/Agents/RemoteFunctions.js
@@ -262,12 +262,17 @@ function RemoteFunctions(config, remoteWSPort) {
                 return;
             }
 
-            var showMarginPadding;
-            if (config.remoteHighlight.showPaddingMargin) {
-                showMarginPadding = 'block';
-            } else {
-                showMarginPadding = 'none';
-            }
+            var setVisibility = function (el){
+                if (
+                    !config.remoteHighlight.showPaddingMargin || 
+                    parseInt(el.height, 10) <= 0 || 
+                    parseInt(el.width, 10) <= 0 
+                ) {
+                    el.display = 'none';
+                } else {
+                    el.display = 'block';
+                }
+            };
             
             var elementStyling = window.getComputedStyle(element, null);
             
@@ -293,9 +298,7 @@ function RemoteFunctions(config, remoteWSPort) {
                     ) + "px",
                     "left": -calculateSize(
                         mainBoxStyles['border-width']
-                    ) + "px",
-                    "box-sizing": "border-box",
-                    "display": showMarginPadding
+                    ) + "px"
                 },
                 // padding-right
                 {
@@ -305,9 +308,7 @@ function RemoteFunctions(config, remoteWSPort) {
                     "top": 0,
                     "right": -calculateSize(
                         mainBoxStyles['border-width']
-                    ) + "px",
-                    "box-sizing": "border-box",
-                    "display": showMarginPadding
+                    ) + "px"
                 },
                 // padding-bottom
                 {
@@ -317,9 +318,7 @@ function RemoteFunctions(config, remoteWSPort) {
                     "bottom": -calculateSize(
                         mainBoxStyles['border-width']
                     ) + "px",
-                    "left": 0,
-                    "box-sizing": "border-box",
-                    "display": showMarginPadding
+                    "left": 0
                 },
                 // padding-left
                 {
@@ -329,9 +328,7 @@ function RemoteFunctions(config, remoteWSPort) {
                     "top": 0,
                     "left": -calculateSize(
                         mainBoxStyles['border-width']
-                    ) + "px",
-                    "box-sizing": "border-box",
-                    "display": showMarginPadding
+                    ) + "px"
                 }
             ];
                 
@@ -347,9 +344,7 @@ function RemoteFunctions(config, remoteWSPort) {
                     ) + "px",
                     "left": -calculateSize(
                         mainBoxStyles['border-width']
-                    ) + "px",
-                    "box-sizing": "border-box",
-                    "display": showMarginPadding
+                    ) + "px"
                 },
                 // margin-right
                 {
@@ -368,9 +363,7 @@ function RemoteFunctions(config, remoteWSPort) {
                     "right": -calculateSize(
                         elementStyling.getPropertyValue('margin-right'),
                         mainBoxStyles['border-width']
-                    ) + "px",
-                    "box-sizing": "border-box",
-                    "display": showMarginPadding
+                    ) + "px"
                 },
                 // margin-bottom
                 {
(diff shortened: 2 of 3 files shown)


src/LiveDevelopment/Agents/RemoteAgent.js
```diff
@@ -128,10 +128,10 @@ define(function RemoteAgent(require, exports, module) {
         }
 
         _stopKeepAliveInterval();
-
-        // inject RemoteFunctions
-        var command = "window._LD=" + RemoteFunctions + "(" + LiveDevelopment.config.experimental + ");";
-
+
```
petetnt: Nit: trailing whitespace

src/LiveDevelopment/main.js
```diff
@@ -79,6 +79,21 @@ define(function main(require, exports, module) {
         description: Strings.DESCRIPTION_LIVE_DEV_MULTIBROWSER
     });
 
+    // "livedev.remoteHighlight" preference
+    var PREF_REMOTEHIGHLIGHT = "remoteHighlight";
+    var remoteHighlightPref = prefs.definePreference(PREF_REMOTEHIGHLIGHT, "object", {
+        "animateStartValue": {
```
petetnt: nit: Keys don't need to be quoted

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -268,16 +269,17 @@ function RemoteFunctions(experimental) {
                 "box-shadow": "0 0 1px #fff",
                 "box-sizing": "border-box"
             };
+            
+            var preferences = config.remoteHighlight.stylesToSet;
```
petetnt: You can use `Object.assign` here to simplify the merging:

``` js
var mergedStyles = Object.assign({}, stylesToSet,  config.remoteHighlight.stylesToSet)
```

and then later use `mergedStyles` instead of `stylesToSet`

src/LiveDevelopment/Agents/RemoteAgent.js
```diff
@@ -128,10 +128,10 @@ define(function RemoteAgent(require, exports, module) {
...
 
         _stopKeepAliveInterval();
-
-        // inject RemoteFunctions
-        var command = "window._LD=" + RemoteFunctions + "(" + LiveDevelopment.config.experimental + ");";
-
+        
+        // inject RemoteFunction
+        var command = "window._LD=" + RemoteFunctions + "(" + JSON.stringify(LiveDevelopment.config) + ");";
+
```
petetnt: Nit: trailing whitespace

petetnt (changes requested):
Nice work @Worie! Added some small comments.

Might require a test too.

Worie:
Thanks! I've applied the changes you have mentioned. Unfortunetly it seems like I've broken the test, as `Debug -> Run tests` throws `Uncaught TypeError: Cannot read property 'experimental' of undefined` in `SpecRunner.html:36` . I can't investigate on it now, though.

Also it looks like I've closed this PR 3 hours ago, is it automated process or did i just close it by accident?

petetnt:
These closing is not an automated process, I guess you closed it by accident but that's no biggie 👍 

I can help you out with debugging the test failures in a while so we can land this 👍

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -268,21 +269,17 @@ function RemoteFunctions(experimental) {
...
 
-            var animateEndValues = {
-                "background-color": "rgba(0, 162, 255, 0)",
-                "opacity": 1
-            };
+            var animateEndValues = config.remoteHighlight.animateEndValue;
 
             var transitionValues = {
-                "-webkit-transition-property": "opacity, background-color",
+                "-webkit-transition-property": "opacity, background-color, transform",
```
petetnt: I think we don't need to support the prefixed ones anymore

Worie:
It isn't pushed yet, but I've created `updateConfig` function in `RemoteFunctions.js`  so the changes in the `brackets.json` could be loaded to current LiveDev session. 

Also I'd like to try to implement https://github.com/adobe/brackets/issues/7276 in my free time, as I think it is an awesome case  and it could play well with changes made in this PR.

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -268,21 +269,15 @@ function RemoteFunctions(experimental) {
                 "box-shadow": "0 0 1px #fff",
                 "box-sizing": "border-box"
             };
+            
+            var mergedStyles = Object.assign({}, stylesToSet,  config.remoteHighlight.stylesToSet);
```
ficristo: I think Object.assign is not available on Linux (which uses Chrome 29). Could you try lodash assign or extend instead.

Worie: Those are remote functions, so I cannot use `_.assign` there I guess until i inject it to the browser. Am I right? 

I've left it as `Object.assign`, just for now - until I know what exactly am I supposed to do.

petetnt: Actually as they are remote functions that are run on the browser, we can just use `Object.assign` right?

Worie:
[Preview of Live Preview with margins and paddings showing](http://imgur.com/LDjNmEQ)

So, I've done it :) If user did not specify his own preferences, it looks as designed in https://github.com/adobe/brackets/issues/7276 . Also changing those values does not break current Live Development session.

I could probably add one condition to check if there's any padding, so there wont be any borders showing when padding value is equal to 0 or something.

src/LiveDevelopment/Agents/RemoteFunctions.js
ficristo: You mean a real browser and not Brackets itself, right? (Not much familiar with LivePreview)
Then I suppose we can use `Object.assign`. Sorry for the noise...

Worie: I cant test it on Linux, but it works well on OSX

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -244,6 +245,173 @@ function RemoteFunctions(experimental) {
...
+                    el.display = 'block';
+                }
+            };
+            
+            var elementStyling = window.getComputedStyle(element, null);
+            
+            var calculateSize = function () {
+                var sum = 0;
+                var i;
+                for (i = 0; i < arguments.length; i++) {
```
petetnt: Not a huge one of using `arguments`, can we instead pass an array and loop through that (with `Array.prototype.reduce` for example?)

```
var calculateSize = function (values) {
  return values.reduce(function (val) {
    return parseInt(val, 10);
  }, 0);
};
```

(Side note, I wish we could use ES2015 so this would be just `const calculateSize = (values) => values.reduce(val => parseInt(val, 10), 0)` 😸)

petetnt: `calculateSize` should probably return value in pixels too instead of using string concat on every use.

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -244,6 +245,173 @@ function RemoteFunctions(experimental) {
...
+                    sum += parseInt(arguments[i], 10);
+                }
+                return sum;
+            };
+            
+            var mainBoxStyles = config.remoteHighlight.stylesToSet;
+            
+            var paddingVisualisations = [
```
petetnt: Not sure if these would be better in their own file, for example `RemoteVisualizationUtils`  for example?

petetnt (changes requested):

src/LiveDevelopment/Agents/RemoteFunctions.js
Worie: Hmm, but `RemoteFunctions` are being injected into the browser (`window._LD`) and those values are calculated basing on the element that is highlighted there... I'm not sure how that sould work - how to access another file from there? Can't imagine this atm.

Or perhaps i simply didn't understand what you meant :)

petetnt: Yeah my bad, I was thinking about making it an utility file of methods that get passed to that injected method but it most likely won't work like that nor would be worth it. Feel free to disregard 👍

Worie:
Realized that elements that have css `transition` on them act a bit buggy (margins and paddings do not animate, and it causes that you have to wait for the transition to finish, then re-select this specific node to see changes), perhaps I'll do something about it next week, we'll see. 

But still, i have no idea how to get into fixing those tests - any help would be awesome, even pointing to some specific files or anything. :)

ficristo:
If you open Chrome at http://localhost:9234/, after clicked `Debug \ Run Tests`, there are two errors `Uncaught TypeError: Cannot read property 'experimental' of undefined`.
Their stack trace point to:
[spec/HTMLInstrumentation-test.js:40](https://github.com/adobe/brackets/blob/75406e8c94062a7cd06204632eda78494b8732ef/test/spec/HTMLInstrumentation-test.js#L40)
[spec/RemoteFunctions-test.js:33](https://github.com/adobe/brackets/blob/75406e8c94062a7cd06204632eda78494b8732ef/test/spec/RemoteFunctions-test.js#L33)

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -30,10 +30,24 @@
  * modules should define a single function that returns an object of all
  * exported functions.
  */
-function RemoteFunctions(experimental) {
+function RemoteFunctions(config) {
     "use strict";
 
+    var experimental = config.experimental;
     var lastKeepAliveTime = Date.now();
+	var req, timeout;
```
petetnt: This commit has `tabs`, while we prefer/enforce 4 spaces. You can easily change them with the `brackets-eslint` extension by right clicking the editor and selecting `Autofix with ESLint`

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -233,17 +247,190 @@ function RemoteFunctions(experimental) {
...
-
         _makeHighlightDiv: function (element, doAnimation) {
             var elementBounds = element.getBoundingClientRect(),
                 highlight = window.document.createElement("div"),
                 styles = window.getComputedStyle(element);
+		var transitionDuration = parseFloat(styles.getPropertyValue('transition-duration'));
+			
+		var animationDuration = parseFloat(styles.getPropertyValue('animation-duration'));
+			
+		if (transitionDuration)
```
petetnt: Nit: IIRC we prefer to have braces on single line ifs too:

```
if (transitionDuration) {
  animateHighlight(transitionDuration);
}
if (animationDuration) {
    animateHighlight(animationDuration);
}

File: src/LiveDevelopment/Agents/RemoteFunctions.js [hunk 11]
```diff
@@ -233,17 +247,190 @@ function RemoteFunctions(experimental) {
             }
             return false;
         },
-
         _makeHighlightDiv: function (element, doAnimation) {
             var elementBounds = element.getBoundingClientRect(),
                 highlight = window.document.createElement("div"),
                 styles = window.getComputedStyle(element);
+		var transitionDuration = parseFloat(styles.getPropertyValue('transition-duration'));
```
petetnt: Nit: these two variables could also be combined with the above by separating the values with , like the others are.

petetnt (changes requested):
Few more small comments. Nice work so far 👍

src/LiveDevelopment/Agents/RemoteFunctions.js
Worie: Yeah, I do too. I have no idea why did i write it like so. Thanks!

Worie:
Due to `getClientBoundingRect` including the transform when calculating the size of element and its position (https://i.imgur.com/tNNjXEw.png), i get pretty weird and unintuitive result - http://imgur.com/MTwZcfB .

I do not know how to bite this one without some heavy math calculations (nor I am any good at math).

Perhaps some of you have an idea how to get the current implementation to act like a chrome one easly?

petetnt:
@Worie I was thinking about that but didn't find any simple solutions yet, could you share an sample project / file that shows the behaviour in the GIF?

Worie:
Perhaps this will be enough: https://gist.github.com/Worie/caae957b805f08cb733f51406d32aabe 

Also, now every transform function is applied to the highlight div - probably only scale and rotate should be considered.

Worie:
I've started a little refactor of this code. I'll not focus on the transform issue atm, though.

Worie:
@petetnt In last commit, I've removed the `sum` function we introduced during this PR. I felt that it was making more harm than good and made it less clear what exacly is happening there. 

Of course, I can revert this change and apply the `sum` function to current implementation. But I guess it's not really necessary.

petetnt:
Hey Worie,

sorry for the lack of comms, been a bit busy as of late. Is this PR ready for another round of review?

Worie:
Yeah I guess. If we could only _somehow_ invoke the `inspect` on element as we can in Chrome Dev tools from `RemoteFunctions` this PR would be much shorter :/ But I don't think it is possible atm.

Anyway - the workaround for transform issue is that we simply do not apply transform to our custom highlightning divs. That means that they dont rotate nor scale, but translation works pretty good on them. 

I guess that we could perhaps split this into two issues - close this one when the code quality is good enough, and then take care of the transform issue. 

Because, I think that it might take a while to do (and I have no idea atm how, but I guess we'd need to get the data about the transformations from matrix3d and reverse it to something like `rotate(45deg) translateX()...`)

petetnt:
I think the scope is just fine right now, I could run another code review and test it out again myself and then merge it in (and create a issue for tracking the transform part). @ficristo you could also take a look if you want to 👍

ficristo:
I think @saurabh95 is more indicated

CodingAnonymous:
When will the Live Preview Highlight with Margin and Padding be available? Or can I activate it manually now?
Thanks for your help!

CodingAnonymous:
This feature is so great and important, can I use it right now through some code implementation? http://imgur.com/Gx1r4P3

Worie:
You can always clone my fork and set it as a source for Brackets, see [how to set up dev environment](https://github.com/adobe/brackets/wiki/How-to-Hack-on-Brackets#setting-up-your-dev-environment).

Can I somehow help you guys with this PR? I mean, we could discuss the code that's already here, or perhaps I can tell you my way of thinking when i was writing this. Be sure to ping me when anything like this will be necessary.

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -30,10 +30,29 @@
...
 
+    var experimental;
+    if (!config) {
+        experimental = false;    
+    } else {
+        experimental = config.experimental;    
+    }
     var lastKeepAliveTime = Date.now();
+	var req, timeout;
+	var animateHighlight = function (time) {
```
petetnt: L44-L55 have some indentation issues, convert tabs -> spaces and they should be all right

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -250,17 +269,149 @@ function RemoteFunctions(experimental, remoteWSPort) {
...
+                animateHighlight(animationDuration);
+            }
 
             // Don't highlight elements with 0 width & height
             if (elementBounds.width === 0 && elementBounds.height === 0) {
                 return;
             }
+            
+            var realElBorder = {
+              "right": elementStyling.getPropertyValue('border-right-width'),
```
petetnt: No need to quote the keys in L293-L296

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -250,17 +269,149 @@ function RemoteFunctions(experimental, remoteWSPort) {
...
+              "left": elementStyling.getPropertyValue('border-left-width'),
+              "top": elementStyling.getPropertyValue('border-top-width'),
+              "bottom": elementStyling.getPropertyValue('border-bottom-width'),
+            };
+            
+            var innerWidth = elementBounds.width - parseInt(realElBorder.left) - parseInt(realElBorder.right);
+            var innerHeight = elementBounds.height - parseInt(realElBorder.top) - parseInt(realElBorder.bottom);
+          
+            var visualisations = [];
```
petetnt: This should be an object literal instead of assigning `horizontal` and `vertical` as array properties?

``` js
var visualisations = {
 horizontal: "left, right",
 vertical: "top, bottom",
}
```

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -250,17 +269,149 @@ function RemoteFunctions(experimental, remoteWSPort) {
...
+            visualisations['horizontal'] = "left, right";
+            visualisations['vertical'] = "top, bottom";  
+          
+            var drawPaddingRect = function(side) {
+              var elStyling = {};
+              
+              if(visualisations['horizontal'].indexOf(side) >= 0) {
```
petetnt: Maybe prefer the dot-notation (`visualisations.horizontal` etc.) in L309-351.

petetnt: Nit: space between `if` and `(` -> `if (visualisations)...`

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -250,17 +269,149 @@ function RemoteFunctions(experimental, remoteWSPort) {
...
+              elStyling[side] = realElBorder[side];
+              elStyling['position'] = 'absolute';
+              
+              return elStyling;
+            };
+          
+          var drawMarginRect = function(side) {
+            var elStyling = {};
+            
+            var  margin = [];
```
petetnt: Nit: extra space between `var` and `margin`

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -250,17 +269,149 @@ function RemoteFunctions(experimental, remoteWSPort) {
...
+                    parseInt(el.width, 10) <= 0 
+                ) {
+                    el.display = 'none';
+                } else {
+                    el.display = 'block';
+                }
+            };
+            
+            var mainBoxStyles = config.remoteHighlight.stylesToSet;
+                mainBoxStyles['border'] = 'none';
```
petetnt: Nit: `mainBoxStyles` should be aligned to `var`

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -250,17 +269,149 @@ function RemoteFunctions(experimental, remoteWSPort) {
...
+                    arr[i]["box-sizing"] = "border-box";
+                    arr[i]["transform"] = "none";
+                    var el = window.document.createElement("div"),
+                        styles = Object.assign(
+                        {},
+                        config,
+                        arr[i]
+                    );
+
+                    _setStyleValues(
```
petetnt: Nit: This should be a one-liner as there's only couple of short properties

petetnt (changes requested):
Sorry about taking so long @Worie, this got a bit buried between other work.

The implementation looks a bit complex, but then again what it achieves is also a bit complex so that is okay.

I put some small review comments about the code style, and one that is most likely an error, but otherwise it looks pretty good to me.

You can run the files through `eslint` to catch some other errors there might be if any. 👍

src/LiveDevelopment/Agents/RemoteFunctions.js
Worie: Oh, okay. It's common practice for me to place parameters (even short ones) in separate lines so git displays the changes in more friendly way. But sure, I'll make it a one-liner. 👍

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -30,10 +30,29 @@
  * modules should define a single function that returns an object of all
  * exported functions.
  */
-function RemoteFunctions(experimental, remoteWSPort) {
+function RemoteFunctions(config, remoteWSPort) {
```
saurabh95: I think we can pass remoteWSPort as part of config only, something like `config.remoteWSPort`.
Then instead of checking for `remoteWSPort` in https://github.com/adobe/brackets/pull/12949/files#diff-a40df952af8fd1808c321dab548ae11aR1032, we can check for `!experimental` flag as we create websocket only when we are not using experimental live preview.
But I think you can leave this for later as you might have to make changes which are not relevant for this feature, once this is merged I can make the above changes.

saurabh95 (approved):

src/LiveDevelopment/Agents/RemoteFunctions.js
Worie: Hi, you're right. Unfortunetely I won't be accessible for a while, so I guess we could close this one and open new PRs as necessary, if that's OK.

Worie:
@petetnt what do you think?

petetnt (approved):

petetnt:
Also thanks for this great contribution and sorry that it took so long to get merged 👍

Worie:
Awesome! Sure, I'll create issues that were not covered in this PR. Thanks!

PR #12951, opened 2016-11-28
Fix FindInFiles tests

ficristo:
After https://github.com/adobe/brackets/pull/12885 there were some failures.
This PR fixes them but I have the bad feeling I'm workarounding a bug somewhere else.


ficristo:
@zaggino do you have time to look at this?

zaggino:
@ficristo check this https://github.com/adobe/brackets/pull/12973
still some are failing (case sensitive replace in file) but at least something, maybe I can spend more time on this next week

PR #12963, opened 2016-12-02
Move CodeMirror on node_modules

ficristo:
What I had in mind was something like this.
I think this will NOT work on distribution since the node_modules would be in a different location.
Also I find worrisome I had to change the css link.

Ref #12940

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "src/thirdparty/CodeMirror"]
-	path = src/thirdparty/CodeMirror
-	url = https://github.com/adobe/CodeMirror2.git
 [submodule "src/thirdparty/path-utils"]
 	path = src/thirdparty/path-utils
 	url = https://github.com/jblas/path-utils.git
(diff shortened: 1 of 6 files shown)


zaggino:
Looks good to me, let's test how the distribution mode works.

ficristo:
Otherwise we could put another `package.json` in the `src` folder.
But the two `package.json` doesn't seem necessary anymore with the newer versions of electron-builder and I think this is a bit of a contributor barrier.
I still don't have any idea how to change the path for distribution in case of single `package.json`.
At this point I don't have strong opinions: thoughts?

swmitra:
@ficristo The changes looks good to me. But I have a different question. When I tried to check the changes on my local machine, I had to do a npm install. Is that expected when someone clones the repo for dev setup?

zaggino:
@swmitra yes, `npm install` is expected, it has also been added here: https://github.com/adobe/brackets/wiki/How-to-Hack-on-Brackets#setting-up-your-dev-environment

ficristo:
I tryed to hack something but I wasn't able to make `grunt build` to pass. (In the specific `grunt requirejs`)
If nobody has some bright idea I don't think this will work, as is, on distribution.
I'll create a new `package.json` inside `src` and see how this will work.
Of course #12940 is still a choice.

ficristo:
Closing in favor of #12972.

PR #12972, opened 2016-12-07
Move npm dependencies inside src and add CodeMirror to them

ficristo:
Alternative to #12963
I was wondering if, instead of this approach, we could use some other tool to copy `node_modules` like using browserify, webpack or others.

Ref #12940

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ Thumbs.db
 /node_modules
 /npm-debug.log
 
+# ignore node_modules inside src
+/src/node_modules
+
 # ignore compiled files
 /dist
 /src/.index.html
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "src/thirdparty/CodeMirror"]
-	path = src/thirdparty/CodeMirror
-	url = https://github.com/adobe/CodeMirror2.git
 [submodule "src/thirdparty/path-utils"]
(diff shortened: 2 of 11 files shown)


zaggino:
Yeah, definitely prefer this one over all the others. Only thing that bothers me a little bit is the generation of `npm-shrinkwrap` file since you need to do that from src folder AND not to forget about it 👍

zaggino (approved):

zaggino:
That'd mean introducing another tool, we'd have to use `yarn` instead of `npm install`, not sure how `yarn` works with `npm install` scripts... to be honest I don't think `yarn` is worth the complexity it adds.

petetnt:
I agree with introducing another tool, but it's a tool that's fully compatible with `npm` and pretty much improvement in every way (especially with speed and security), especially if we need to fiddle around with `npm-shrinkwrap` which is basically somewhat broken at this point.

For the grunt script the work needed would be basically `s/npm install/yarn`, plus with enforcing `--pure-lockfile` we could ensure that right deps would always be installed.

Disclaimer: I have only used `yarn` after it came out and I am a huge fan 😸 Not saying that we would need to do it right now (or ever), we can punt it into an issue and think about it later too 👍

zaggino:
I think this is the best approach from the three we tried.
If we have problems with shrinkwrap, we can introduce `yarn` later.
For now merging this 👍

PR #12973, opened 2016-12-09
fixes 3 failing find-in-files tests

zaggino:
this PR fixes:
```
should add matches for a new file
should remove matches for a deleted file
should remove matches for a deleted folder
```
on my windows machine
ref: https://github.com/adobe/brackets/pull/12951

### DIFF
--- a/test/spec/FindInFiles-test.js
+++ b/test/spec/FindInFiles-test.js
@@ -785,7 +785,8 @@ define(function (require, exports, module) {
                 gotChange = false;
                 oldResults = null;
                 wasQuickChange = false;
-                FindInFiles.searchModel.clear();
+
+                FindInFiles.clearSearch(); // calls FindInFiles.searchModel.clear internally
                 FindInFiles.searchModel.on("change.FindInFilesTest", function (event, quickChange) {
                     gotChange = true;
                     wasQuickChange = quickChange;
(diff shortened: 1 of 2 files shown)


test/spec/FindInFiles-test.js
```diff
@@ -785,6 +785,7 @@ define(function (require, exports, module) {
                 gotChange = false;
                 oldResults = null;
                 wasQuickChange = false;
+                FindInFiles.searchModel.clear();
```
ficristo: Maybe call `FindInFiles.clearSearch`? It is marked as private method but it seems it is already used outside that file.

ficristo:
The test `should close the panel and detach listeners if a file is modified on disk` is still failing for me.
I suppose it needs a different fix.

`should do a case-sensitive search/replace from find bar` was failing even before, so it's fine (for now).

ficristo (approved):

test/spec/FindInFiles-test.js
zaggino: Since we're calling `FindInFiles.searchModel.on` on the next line, I think calling `FindInFiles.searchModel.clear` directly is more clearer.

zaggino:
Yeah, one test at a time I guess... I'll check what can fix the others next week when I'll have some free time.

test/spec/FindInFiles-test.js
ficristo: I don't have strong opinion here and I agree is it a bit odd.
My only doubt is `clearSearch` reset also `findOrReplaceInProgress`: I'm not sure how much it is important.

zaggino:
@ficristo ready to merge this one, i'll open another if I manage to fix some other test

PR #12974, opened 2016-12-09
Test against node 7

amilajack:


zaggino:
Brackets is distributed with Node 6, there's no plan of using Node 7 at the moment.

PR #12979, opened 2016-12-13
Fix for #12970: Non-default languages and user preferences bug

### LINKED ISSUE #12970: Brackets reset default language to plain/text
### Prerequisites

* [ ] Can you reproduce the problem with `Debug -> Reload Without Extensions`?
Only because this bug is linked to highlighting extensions.
* [x] Did you perform a cursory search to see if your bug or enhancement is already reported?
* [x] Did you read the [Troubleshooting guide](https://github.com/adobe/brackets/wiki/Troubleshooting)?

### Description

Brackets reset language for a file type to text if the language set as default for the file type is defined by an extension and this extension doesn't define this file type in the list provided to language manager

### Steps to Reproduce

1. Download an extension which highlight a language not supported by Brackets (this [one](https://github.com/Athorcis/brackets-twig) by example)
2. Open a file which file type is not supported by the extension you installed
3. Set the language to the one supported by the extension
4. Set this language as the default one
5. Reload Brackets (with extensions).

**Expected behavior:** The language for the file is the one supported by the extension

**Actual behavior:** The language is Text.

### Versions
Windows 10 / Brackets 1.8
Also reproduced under Windows 7 / Brackets 1.8 (with only one extension the one mentioned above)

haslam22:
This fixes #12970 by removing invalid mappings from the preference state, occurring when a language has not been loaded yet. When the language is eventually loaded, an `_updateFromPrefs` call restores the preferences.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -20,7 +20,7 @@
         "extension_url": "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default": true,
         "build_timestamp": "",
-        "healthDataServerURL": "https://healthdev.brackets.io/healthDataLog"
+        "healthDataServerURL": "https://health.brackets.io/healthDataLog"
     },
     "name": "Brackets",
     "version": "1.9.0-0",
--- a/src/config.json
+++ b/src/config.json
@@ -20,7 +20,7 @@
         "extension_url": "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default": true,
         "build_timestamp": "",
-        "healthDataServerURL": "https://health.brackets.io/healthDataLog"
+        "healthDataServerURL": "https://healthdev.brackets.io/healthDataLog"
     },
     "name": "Brackets",
     "version": "1.9.0-0",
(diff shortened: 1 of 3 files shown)


zaggino:
@Athorcis can you review and verify this fixes #12970 ?
I've tried this but it doesn't seem to resolve the steps described in your issue.

Athorcis:
@zaggino i'll try and i'll let you know

haslam22:
@zaggino Just tested it again on Windows & Ubuntu with the latest master and it works fine with the steps described in the issue. I may have missed something so I'll take another look later

haslam22:
Reset the branch, I think I messed up my fork somehow. Should work fine now.

haslam22:
I was able to produce a fix for the steps in the issue but I didn't try using .twig files (issue describes step 2 as any file type that's not supported). I can now see that overriding .twig is a specific scenario that doesn't work, but all other extensions are fine.

So I've added a fix for this as well. You can now override `.twig` files and make it use your own extension.

Apologies for any spam.

Athorcis:
Thx for your work, i don't have the bug anymore

zaggino:
@haslam22 let me know when you think it's ready, I see you're still adding some commits.

haslam22:
@zaggino All finished now, thanks for taking the time to review.

zaggino (approved):

zaggino:
reviewed & merged, thanks @haslam22 !

petetnt:
Nice job @haslam22 👍

PR #12781, opened 2016-09-14
Added will-change property

febLey:

PR #12982, opened 2016-12-14
Add will-change to css properties

ficristo:
Replace https://github.com/adobe/brackets/pull/12781


ficristo:
CodeMirror ATM doesn't really support this property, we need make a change there too.

ficristo:
I've opened https://github.com/codemirror/CodeMirror/pull/4434

PR #12993, opened 2016-12-20
Merge code with latest code

pk-codebox-evo:
Update to latest code


zaggino:
this seems like an error

PR #13019, opened 2017-01-02
Upgrade less from 2.5.1 to 2.7.2

ficristo:
I would like to upgrade grunt related dependencies and this was in the way: updating only the grunt dep without the thirdparty didn't seem good.
I renamed `less-2.5.1.min.js` to simply `less.min.js`, you can see the version in the comments inside the file.


zaggino:
could we maybe switch to npm dependency instead?

ficristo:
I'm all for moving to npm.
But has pointed out before by @busykai, we need to do something about the increase size of the bundle.
For example less 2.7.1 from npm is about 3MB, but the minified file is only 140kB.
I have some random thought but still nothing good.
So WDYT?

zaggino:
Is it about npm not providing minified file or about npm providing too many files that we don't need?

ficristo:
Too many files, I've opened #13020 as a possible solution

ficristo:
Now it is fetched from npm.

zaggino (approved):

zaggino:
👍  cool

ficristo:
Thank you.

PR #13020, opened 2017-01-03
Add a post install 'cleanup' for thirdparty dependencies

ficristo:
This is basically a revert of #12972 and it is similar of #12940 but instead of using npm from toplevel it uses the one in src.
After this, node dependencies should be added to the toplevel package.json, browser dependencies should be added to the package.json in src.
The copy:thirdparty grunt task should be used to filter the file we need in the browser.
I'm not particular happy with this, but using npm directly add a lot of files we don't need that will end in the final bundle.


zaggino (approved):

zaggino:
Sorry, I was busy and couldn't give this closer look, looks good. Merging

ficristo:
No problem, thank you :smile:

PR #6461, opened 2014-01-10
#5059 chmod all sub files and folders in temp directory

sriram-dev:
Programatically set permission to all sub files and folders of temp directory.

Passed Unit test and JsLint

PR #13023, opened 2017-01-05
chmod all sub files and folders in temp directory

### LINKED ISSUE #5059: SpecRunnerUtils.removeTempDirectory() should make all files & folders read/write
TODO in SpecRunnerUtils.removeTempDirectory:

Because the Mac cannot delete files or folders that are read only or write only -- we should recursively iterate over all of the files and folders and chmod everything rather than just the 2 hard coded directories.

ficristo:
Replaces #6461. Thanks to @sriram-dev to all the work, I only merged master to resolve a conflict.
Fixes #5059.

/cc @zaggino @ingorichter


zaggino:
looks good, merged

PR #13026, opened 2017-01-07
Add a maxsize check for filetree resize…

### LINKED ISSUE #6822: Sidebar can't be resized once filling whole window

ficristo:
… and now it is possible to pass a percentage value.
It replaces #9122 and #12829
This fixes #6822, I'm sorry to take a starter bug but I thought it was open long enough that's time to fix it.

@redmunds could you review?

### DIFF
--- a/src/utils/Resizer.js
+++ b/src/utils/Resizer.js
@@ -62,9 +62,11 @@ define(function (require, exports, module) {
         ViewUtils               = require("utils/ViewUtils"),
         PreferencesManager      = require("preferences/PreferencesManager");
 
-    var $mainView;
+    var $mainView,
+        $sideBar;
 
-    var isResizing = false;
+    var isResizing = false,
+        isWindowResizing = false;
 
     /**
      * Shows a resizable element.
@@ -131,6 +133,25 @@ define(function (require, exports, module) {
      */
     function isVisible(element) {
         return $(element).is(":visible");
+    }
+
+    function _isPercentage(value) {
+        return !$.isNumeric(value) && value.indexOf('%') > -1;
+    }
+
+    function _percentageToPixels(value, total) {
+        return parseFloat(value.replace('%', '')) * (total / 100);
+    }
+
+    function _sideBarMaxSize() {
+        var siblingsWidth = 0;
+        $sideBar.siblings().not(".content").each(function (i, elem) {
+            var $elem = $(elem);
+            if ($elem.css("display") !== "none") {
+                siblingsWidth += $elem.outerWidth();
+            }
+        });
+        return $(".main-view").width() - siblingsWidth - 1;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ficristo:
Argh... I noticed only now #10456. @redmunds WDYT?

redmunds:
@ficristo I'm surprised this issue wasn't fixed long ago. Thanks for putting up a PR. 

Your code works as advertised. LGTM.

@lucaslouca Brings up a good point in PR #10456:

> Also important is the scenario when the Sidebar is at its maximum size (filling the whole window) and the user decreases the Window's width.

You can hit this case even when sidebar is less than 100%. I was going to try out this code, but the remote branch seems to have been deleted, so I'm not sure if it actually fixes that case.

Do you want to try to fix that in this PR? If not, this is worth merging anyway.

ficristo:
With the new commit, it should take care of the window resize.
I tryed to use `maxsize= 100%` but it has the old problem but I think it is fine for now.

redmunds:
@ficristo Nice. Works great. I have a minor suggestion, but feel free to merge PR as is.

When window size is reduced, you enforce that leftSidebar.width is <= window.width - rightSidebar.width. Why don't you also apply the 80% to that calculation? This will maintain the same behavior as when dragging leftSidebar with.

ficristo:
Now it should take care of the percentage.
But I noticed an issue with the selection of current file...
current-file

This happen if you:
- resize the filetree at max
- resize the window to go over the file tree
- make sure the filetree was resized automatically
- resize the window to enlarge it

redmunds:
That only happens if there are enough files in WorkingSet to require scrollbars. That line is the shadow at bottom of WorkingSetView. The editor view is not getting refreshed.

ficristo:
Triggering the `panelResizeUpdate` and `panelResizeEnd` events seems to have fixed it.
Triggering only one had other side effects.
It feels a bit dirty but I don't have other ideas at the moment.

redmunds:
@ficristo I think triggering those events is the correct fix, although `panelResizeStart` should also be called to satisfy the API. Also, 1 typo, then LGTM.

ficristo:
I've added the call to `panelResizeStart` event and rebased.
Which is the typo?

src/utils/Resizer.js
```diff
@@ -133,6 +135,25 @@ define(function (require, exports, module) {
         return $(element).is(":visible");
     }
 
+    function _isPercentage(value) {
+        return !$.isNumeric(value) && value.indexOf('%') > -1;
+    }
+
+    function _percetageToPixels(value, total) {
```
haslam22: Your typo is `percetageToPixels` -> `percentageToPixels`. Also in all the function calls.

ficristo: Thank you for noticing! Fixed.

redmunds:
LGTM

ficristo:
@redmunds @haslam22 Thank you for the review.

PR #13044, opened 2017-01-17
Added Reverse Inspect in Live Preview using WebSockets, and now forwa…

saurabh95:
Now you can click on any element in Live Preview and then corresponding HTML element will be highlighted and the cursor in code view will shift to the corresponding HTML element.

Features that need to be incorporated in this PR
- [x] Get Socket Port Number from preferences.
- [x] Add commands to create and destroy server and create server should be called only when user clicks on Live Preview button.

I am currently working on these. Will update the PR soon.

### DIFF
--- a/src/LiveDevelopment/Agents/RemoteAgent.js
+++ b/src/LiveDevelopment/Agents/RemoteAgent.js
@@ -33,10 +33,11 @@
 define(function RemoteAgent(require, exports, module) {
     "use strict";
 
-    var LiveDevelopment = require("LiveDevelopment/LiveDevelopment"),
-        EventDispatcher = require("utils/EventDispatcher"),
-        Inspector       = require("LiveDevelopment/Inspector/Inspector"),
-        RemoteFunctions = require("text!LiveDevelopment/Agents/RemoteFunctions.js");
+    var LiveDevelopment     = require("LiveDevelopment/LiveDevelopment"),
+        EventDispatcher     = require("utils/EventDispatcher"),
+        Inspector           = require("LiveDevelopment/Inspector/Inspector"),
+        RemoteFunctions     = require("text!LiveDevelopment/Agents/RemoteFunctions.js"),
+        PreferencesManager  = require("preferences/PreferencesManager");
 
     var _load; // deferred load
     var _objectId; // the object id of the remote object
@@ -130,7 +131,7 @@ define(function RemoteAgent(require, exports, module) {
         _stopKeepAliveInterval();
 
         // inject RemoteFunctions
-        var command = "window._LD=" + RemoteFunctions + "(" + LiveDevelopment.config.experimental + ");";
+        var command = "window._LD=" + RemoteFunctions + "(" + LiveDevelopment.config.experimental + "," + PreferencesManager.get("livedev.wsPort") + ");";
 
         Inspector.Runtime.evaluate(command, function onEvaluate(response) {
             if (response.error || response.wasThrown) {
--- a/src/LiveDevelopment/Agents/RemoteFunctions.js
+++ b/src/LiveDevelopment/Agents/RemoteFunctions.js
@@ -22,7 +22,7 @@
  */
 
 /*jslint forin: true */
-/*global Node, document */
+/*global Node */
 /*theseus instrument: false */
 
 /**
@@ -102,7 +102,7 @@ function RemoteFunctions(experimental, remoteWSPort) {
     // Checks if the element is in Viewport in the client browser
     function isInViewport(element) {
         var rect = element.getBoundingClientRect();
-        var html = document.documentElement;
+        var html = window.document.documentElement;
         return (
             rect.top >= 0 &&
             rect.left >= 0 &&
@@ -864,20 +864,20 @@ function RemoteFunctions(experimental, remoteWSPort) {
             }));
         }
     }
-
-    window.document.addEventListener("click", onDocumentClick);
     
     
     function createWebSocket() {
         _ws = new WebSocket("ws://localhost:" + remoteWSPort);
         _ws.onopen = function () {
+            window.document.addEventListener("click", onDocumentClick);
         };
 				
         _ws.onmessage = function (evt) {
         };
 				
         _ws.onclose = function () {
-            // websocket is closed.
+            // websocket is closed
+            window.document.removeEventListener("click", onDocumentClick);
         };
     }
     
--- a/src/LiveDevelopment/Agents/RemoteFunctions.js
+++ b/src/LiveDevelopment/Agents/RemoteFunctions.js
@@ -319,6 +319,10 @@ function RemoteFunctions(experimental) {
             if (this.trigger) {
                 _trigger(element, "highlight", 1);
             }
+            
+            if (!window.event) {
+                element.scrollIntoViewIfNeeded();
+            }
             this.elements.push(element);
 
             this._makeHighlightDiv(element, doAnimation);
@@ -824,7 +828,39 @@ function RemoteFunctions(experimental) {
     if (experimental) {
         window.document.addEventListener("keydown", onKeyDown);
     }
+    
+    var _ws = null;
 
+    function onDocumentClick(event) {
+        var element = event.target,
+            currentDataId,
+            newDataId;
+        if (element && element.hasAttribute('data-brackets-id')) {
+            _ws.send(JSON.stringify({
+                type: "message",
+                message: element.getAttribute('data-brackets-id')
+            }));
+        }
+    }
+
+    window.document.addEventListener("click", onDocumentClick);
+    
+    
+    function createWebSocket() {
+        _ws = new WebSocket("ws://localhost:8125");
+        _ws.onopen = function () {
+        };
+				
+        _ws.onmessage = function (evt) {
+            var received_msg = evt.data;
+        };
+				
+        _ws.onclose = function () {
+            // websocket is closed.
+        };
+    }
+    
+    createWebSocket();
     return {
         "DOMEditHandler"        : DOMEditHandler,
         "keepAlive"             : keepAlive,
--- a/src/LiveDevelopment/Agents/RemoteFunctions.js
+++ b/src/LiveDevelopment/Agents/RemoteFunctions.js
@@ -22,7 +22,7 @@
  */
 
 /*jslint forin: true */
-/*global Node */
+/*global Node, document */
 /*theseus instrument: false */
 
 /**
@@ -97,6 +97,23 @@ function RemoteFunctions(experimental, remoteWSPort) {
         } else {
             element.removeAttribute(key);
         }
+    }
+    
+    // Checks if the element is in Viewport in the client browser
+    function isInViewport(element) {
+        var rect = element.getBoundingClientRect();
+        var html = document.documentElement;
+        return (
+            rect.top >= 0 &&
+            rect.left >= 0 &&
+            rect.bottom <= (window.innerHeight || html.clientHeight) &&
+            rect.right <= (window.innerWidth || html.clientWidth)
+        );
+    }
+    
+    // returns the distance from the top of the closest relatively positioned parent element
+    function getDocumentOffsetTop(element) {
+        return element.offsetTop + (element.offsetParent ? getDocumentOffsetTop(element.offsetParent) : 0);
     }
 
     // construct the info menu
@@ -320,8 +337,12 @@ function RemoteFunctions(experimental, remoteWSPort) {
                 _trigger(element, "highlight", 1);
             }
             
-            if (!window.event) {
-                element.scrollIntoViewIfNeeded();
+            if (!window.event && !isInViewport(element)) {
+                var top = getDocumentOffsetTop(element);
+                if (top) {
+                    top -= (window.innerHeight / 2);
+                    window.scrollTo(0, top);
+                }
             }
             this.elements.push(element);
 
@@ -835,7 +856,8 @@ function RemoteFunctions(experimental, remoteWSPort) {
         var element = event.target,
             currentDataId,
             newDataId;
-        if (element && element.hasAttribute('data-brackets-id')) {
+        
+        if (_ws && element && element.hasAttribute('data-brackets-id')) {
             _ws.send(JSON.stringify({
                 type: "message",
                 message: element.getAttribute('data-brackets-id')
@@ -852,7 +874,6 @@ function RemoteFunctions(experimental, remoteWSPort) {
         };
 				
         _ws.onmessage = function (evt) {
-            var received_msg = evt.data;
         };
 				
         _ws.onclose = function () {
@@ -860,7 +881,10 @@ function RemoteFunctions(experimental, remoteWSPort) {
         };
     }
     
-    createWebSocket();
+    if (remoteWSPort) {
+        createWebSocket();
+    }
+    
     return {
         "DOMEditHandler"        : DOMEditHandler,
         "keepAlive"             : keepAlive,
--- a/src/LiveDevelopment/Agents/RemoteFunctions.js
+++ b/src/LiveDevelopment/Agents/RemoteFunctions.js
@@ -30,7 +30,7 @@
  * modules should define a single function that returns an object of all
  * exported functions.
  */
-function RemoteFunctions(experimental) {
+function RemoteFunctions(experimental, remoteWSPort) {
     "use strict";
 
     var lastKeepAliveTime = Date.now();
@@ -847,7 +847,7 @@ function RemoteFunctions(experimental) {
     
     
     function createWebSocket() {
-        _ws = new WebSocket("ws://localhost:8125");
+        _ws = new WebSocket("ws://localhost:" + remoteWSPort);
         _ws.onopen = function () {
         };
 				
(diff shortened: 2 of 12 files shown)


src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -319,6 +319,10 @@ function RemoteFunctions(experimental) {
             if (this.trigger) {
                 _trigger(element, "highlight", 1);
             }
+            
+            if (!window.event) {
+                element.scrollIntoViewIfNeeded();
```
ficristo: According to the current version of [MDN](https://developer.mozilla.org/it/docs/Web/API/Element/scrollIntoViewIfNeeded), this function is supported only in Chrome, Opera and Safari.
I'm not familiar with LivePreview but this file is injected only in case of LivePreview with Chrome, or it is also used with MultiBrowserLivePreview?
In the seconda case, is it possible to change it to use `scrollIntoView` instead?

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -824,7 +828,39 @@ function RemoteFunctions(experimental) {
...
+    
+    
+    function createWebSocket() {
+        _ws = new WebSocket("ws://localhost:" + remoteWSPort);
+        _ws.onopen = function () {
+        };
+				
+        _ws.onmessage = function (evt) {
+            var received_msg = evt.data;
+        };
```
ficristo: If you don't use it, remove it?

src/LiveDevelopment/LiveDevelopment.js
```diff
@@ -195,6 +197,10 @@ define(function LiveDevelopment(require, exports, module) {
      * Handles of registered servers
      */
     var _regServers = [];
+    
+    PreferencesManager.definePreference("livedev.wsPort", "number", 8125, {
+        description: Strings.DESCRIPTION_LIVEDEV_WEBSOCKET_PORT
+    });
```
ficristo: Since you are doing for LiveDevelopment, maybe you should do the same for MultiBrowserLivePreview too. See also #11957

src/LiveDevelopment/transports/WebSocketTransport.js
```diff
@@ -0,0 +1,64 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+// This transport provides a WebSocket connection between Brackets and a live browser preview.
+// This is just a thin wrapper around the Node extension (WebSocketTransportDomain) that actually
+// provides the WebSocket server and handles the communication. We also rely on an injected script in
+// the browser for the other end of the transport.
```
ficristo: This probably should be a JSDoc.

src/LiveDevelopment/transports/WebSocketTransport.js
```diff
@@ -0,0 +1,64 @@
...
+    var domainPath = FileUtils.getNativeBracketsDirectoryPath() + "/" + FileUtils.getNativeModuleDirectoryPath(module) + "/node/WebSocketTransportDomain";
+
+    var WebSocketTransportDomain = new NodeDomain("webSocketTransport", domainPath);
+
+    // Events
+
+    WebSocketTransportDomain.on("message", function (obj, message) {
+        console.log("WebSocketTransport - event - message" + " - " + message);
+        var editor = EditorManager.getActiveEditor(), position;
+        position = HTMLInstrumentation.getPositionFromTagId(editor, parseInt(message, 10));
```
ficristo: Just initialize when declaring it?
```js
var editor = ...,
    position = HTMLInstrumentation...;
```

src/LiveDevelopment/transports/node/WebSocketTransportDomain.js
```diff
@@ -0,0 +1,134 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
```
ficristo: Add a JSDoc to describe the purpose of the file?

src/LiveDevelopment/transports/node/WebSocketTransportDomain.js
```diff
@@ -0,0 +1,134 @@
...
+function init(domainManager) {
+    _domainManager = domainManager;
+    if (!domainManager.hasDomain("webSocketTransport")) {
+        domainManager.registerDomain("webSocketTransport", {major: 0, minor: 1});
+    }
+    domainManager.registerEvent(
+        "webSocketTransport",
+        "message",
+        [
+            {name: "msg", type: "string", description: "JSON message from client page"}
```
ficristo: Put every param on its own line.

src/LiveDevelopment/transports/node/WebSocketTransportDomain.js
```diff
@@ -0,0 +1,134 @@
...
+        ]
+    );
+    domainManager.registerCommand(
+        "webSocketTransport",       // domain name
+        "start",                    // command name
+        _cmdStart,                  // command handler function
+        false,                      // this command is synchronous in Node
+        "Creates the WS server",
+        [
+            {name: "port", type: "number", description: "Port on which server needs to listen"}
```
ficristo: Put every param on its own line.

src/LiveDevelopment/transports/node/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "brackets-livedev-server",
+  "dependencies": {
+    "ws": "~0.4.31"
```
ficristo: I think you can put it on the root package.json and update the shrinkwrap file.
Right now, I don't think this works.

src/language/HTMLInstrumentation.js
```diff
@@ -87,6 +87,22 @@ define(function (require, exports, module) {
         return pos1 && pos2 && pos1.line === pos2.line && pos1.ch === pos2.ch;
     }
 
+    function getPositionFromTagId(editor, tagId) {
+        var marks = editor._codeMirror.getAllMarks(),
+            i,
+            markFound,
+            position;
```
ficristo: position is not used.

src/language/HTMLInstrumentation.js
```diff
@@ -87,6 +87,22 @@ define(function (require, exports, module) {
...
+        
+        for (i = 0; i < marks.length; i++) {
```
ficristo: I think you can replace this `for` with lodash

src/language/HTMLInstrumentation.js
```diff
@@ -87,6 +87,22 @@ define(function (require, exports, module) {
...
+            if (marks[i].tagID === tagId) {
+                markFound = marks[i];
+            }
+        }
+        if (markFound) {
+            return markFound.find().from;
+        }
```
ficristo: If `markFound` can be false maybe you should add a return value, otherwise add a `console.error` or something.

ficristo (changes requested):
This seems to have broken MultiBrowserLivePreview. At least using Firefox does throw some errors.
Also I thought that standard LivePreview was sort of deprecated and we should not add features to it.

ficristo:
I'm not really familiar with LivePreview to judge the approach. However I left some comments.

src/LiveDevelopment/transports/node/WebSocketTransportDomain.js
```diff
@@ -0,0 +1,134 @@
...
+        ],
+        []
+    );
```
petetnt: NIt: newline here

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -824,7 +828,39 @@ function RemoteFunctions(experimental) {
     if (experimental) {
         window.document.addEventListener("keydown", onKeyDown);
     }
+    
+    var _ws = null;
+
+    function onDocumentClick(event) {
+        var element = event.target,
+            currentDataId,
+            newDataId;
```
petetnt: Nit: new line here

src/LiveDevelopment/LiveDevelopment.js
```diff
@@ -1362,6 +1369,7 @@ define(function LiveDevelopment(require, exports, module) {
             // wait for server (StaticServer, Base URL or file:)
             prepareServerPromise
                 .done(function () {
+                    WebSocketTransport.createWebSocketServer(PreferencesManager.get("livedev.wsPort"));
```
petetnt: See @ficristo's comment above about the `MultiBrowserLivePreview` 👍

src/LiveDevelopment/transports/node/WebSocketTransportDomain.js
```diff
@@ -0,0 +1,134 @@
...
+    if (!domainManager.hasDomain("webSocketTransport")) {
+        domainManager.registerDomain("webSocketTransport", {major: 0, minor: 1});
+    }
+    domainManager.registerEvent(
+        "webSocketTransport",
+        "message",
+        [
+            {name: "msg", type: "string", description: "JSON message from client page"}
+        ]
+    );
```
petetnt: Nit: Newline here

src/LiveDevelopment/transports/node/WebSocketTransportDomain.js
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * Initializes the domain and registers commands.
+ * @param {DomainManager} domainManager The DomainManager for the server
+ */
+function init(domainManager) {
+    _domainManager = domainManager;
+    if (!domainManager.hasDomain("webSocketTransport")) {
+        domainManager.registerDomain("webSocketTransport", {major: 0, minor: 1});
+    }
```
petetnt: Nit: Newline here

petetnt (changes requested):

petetnt:
Left some small comments too, looking good so far 👍

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -824,7 +828,39 @@ function RemoteFunctions(experimental) {
...
+        if (element && element.hasAttribute('data-brackets-id')) {
+            _ws.send(JSON.stringify({
+                type: "message",
+                message: element.getAttribute('data-brackets-id')
```
redmunds: The message is just the element id, correct? When sending messages in other direction (from Brackets to browser), the message has the format:

```
{ method: method, id: id, params: params }
```

It would be nice to do the same here so that's it's easier to add any new messages in the future.

src/LiveDevelopment/transports/WebSocketTransport.js
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (c) 2014 - present Adobe Systems Incorporated. All rights reserved.
```
redmunds: Copyright date for new files should be year that file was added (2017).

src/LiveDevelopment/LiveDevelopment.js
saurabh95: Yeah once this is completed, I will try to do the same for multibrowser also.

src/LiveDevelopment/Agents/RemoteFunctions.js
saurabh95: This script is injected in case of MultiBrowserLivePreview, for now the reverse-inspect doesn't work with the multibrowser, I will take up this task once this PR is merged.(I tried a hacky solution to make it work for multibrowser and it was working. I will raise another PR once I am done with proper solution)
Now, I am using window.scrollTo instead of scrollIntoViewIfNeeded.
I didn't use scrollIntoView because it aligns the element either to top or bottom, but I wanted to align it to center, so I have now used window.scrollTo which is supported by all browsers.

src/LiveDevelopment/transports/node/package.json
saurabh95: Yeah, right now we have to run `npm install` inside the node directory, to run this.
I am trying to figure out a way to move this to root package.json as suggested by you.

src/LiveDevelopment/Agents/RemoteFunctions.js
saurabh95: Actually in multi-browser also we are sending the message from browser to brackets in the same format(NodeSocketTransportRemote.js), that's why I also used the same format, so that in future when we implement the same thing for multi-browser, we can have the same message to interpret in the brackets side.

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -22,15 +22,15 @@
  */
 
 /*jslint forin: true */
-/*global Node */
+/*global Node, document */
```
ficristo: Use `window.document` instead of adding it as a global.

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -824,7 +849,42 @@ function RemoteFunctions(experimental) {
...
+        
+        if (_ws && element && element.hasAttribute('data-brackets-id')) {
+            _ws.send(JSON.stringify({
+                type: "message",
+                message: element.getAttribute('data-brackets-id')
+            }));
+        }
+    }
 
+    window.document.addEventListener("click", onDocumentClick);
```
ficristo: I think it is better to move this call in the `onopen` callback below and
add a removeEventListener in the `onclose` callback.
Otherwise the handler will be called also when the LivePreview is terminated.

ficristo:
There are still a couple of unanswered comments.
Also in `HTMLInstrumentation.js` can you use lodash instead of the `for`.
`_.find` seems what you need.
Finally, you should check the tests and maybe add one?

src/LiveDevelopment/Agents/RemoteAgent.js
```diff
@@ -130,7 +131,7 @@ define(function RemoteAgent(require, exports, module) {
         _stopKeepAliveInterval();
 
         // inject RemoteFunctions
-        var command = "window._LD=" + RemoteFunctions + "(" + LiveDevelopment.config.experimental + ");";
+        var command = "window._LD=" + RemoteFunctions + "(" + LiveDevelopment.config.experimental + "," + PreferencesManager.get("livedev.wsPort") + ");";
```
saurabh95: I have run all tests related to Live Preview, and now MultiBrowserPreview is also working fine.
Regarding passing of config object, once #12949 is merged, I will do the respective changes.

saurabh95:
@ficristo just uploaded new changes, regarding the testing part, I was thinking of writing tests once I am done with MultiBrowserLivePreview as well(I am almost done with it), then I will write tests for both of them.
I have run the current LivePreview tests, and all are passing.

@petetnt @zaggino @MarcelGerber @abose Please review

ficristo:
This does not work on distribution.
To verify
- run `grunt build`
- open Brackets with the `Shift` keyword pressed
- select the `index.html` under `dist/` folder

You have to adjust the `copy` task in `Gruntfile.js`

saurabh95:
Thanks @ficristo for pointing out. I have fixed this and also updated the PR :)

ficristo:
I looked enough at this so approving but I see some failures on Live Preview tests, so please take a look.

ficristo (approved):

saurabh95:
@ficristo Actually there is some problem with the timeout of the Live Preview tests, I increased the timeout and all the Live Preview tests are passing.

ficristo:
@saurabh95 the increase of timeout should be part of this PR?

saurabh95:
Actually without my changes also some LivePreview tests(the same ones) were failing and on increasing timeout they were passing, so I thought we can take change it in some other, as it is independent of this change?
Anyway I will be taking up the task to write tests for my changes, we can then incorporate the changes related to timeout there itself.

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -319,6 +336,14 @@ function RemoteFunctions(experimental) {
             if (this.trigger) {
                 _trigger(element, "highlight", 1);
             }
+            
+            if (!window.event && !isInViewport(element)) {
```
petetnt: Question: not sure why the `window.event` check is needed here? Something to do with legacy IE's?

src/config.json
```diff
@@ -38,7 +38,8 @@
     "dependencies": {
         "anymatch": "1.3.0",
         "chokidar": "1.6.0",
-        "lodash": "4.15.0"
+        "lodash": "4.15.0",
```
petetnt: Lodash has 4.17.4 out now, but feel free to left as is

petetnt (approved):
LGTM with few comments

src/LiveDevelopment/Agents/RemoteFunctions.js
saurabh95: I am checking if there is any window event (mostly click event) is there on the LivePreview instance, then we don't need to bring the element in viewport, as it is already present in the viewport.

swmitra (approved):
Good to be merged.

swmitra:
Great work @saurabh95. This is indeed a great addition, keep more coming 👍

Issue #11658, opened 2015-09-02
Feature Request - Rename All Variable in Document


vimxts:
Renaming variables is currently a pain in Brackets. You can't find and replace with variable names being the same in parts of other text.

I recently used the OneDrive code edit function. (Upload a js and try open it on web). It has an option for the right click menu, to **rename all occurrences of variable name**.

Feature Request
-

I truthfully ask the Brackets team, can a core feature be added to Brackets? To add a **rename all occurrences of variable name** ability somewhere, but the right click menu is a good place. It could look like this:

```
Quick Edit
Quick Docs
Select All
----
Rename variable
---
...
```
The renaming process can take place in a popup modal.

sprintr:
There is an [extension](https://github.com/asgerf/bracket-rename) for this. I would love to see this as a core feature too.

MiguelCastillo:
I use [ternific](https://github.com/MiguelCastillo/Brackets-Ternific) for renaming variables...

ingorichter:
+1 for [ternific](https://github.com/MiguelCastillo/Brackets-Ternific).

pwnorbitals:
Have we seen any progress on this issue recently ?

PR #13047, opened 2017-01-18
[WIP] JS refactor->rename for named variables, members and functions in a single file

swmitra:
This PR implements "rename" functionality in JS mode to enable intelligent rename feature by using Terns scope analysis and inference functions. Renaming across multiple files has been intentionally blocked for the time being as we are not analyzing all the JS files in current project root right now. Once the other PR #11948 is merged, we can leverage on the JS analysis in node and include all files to build symbol table. 
The rename function works by selecting a variable def/ref or placing cursor in and using Ctrl+R. 

Know issues
--------------
In a particular case Tern inference engine is picking up wrong hit. For example - 
 `function fn1(error) {` 
   `     error.a = "abc";`
   `     var a = "test";`
   `     console.log(a);`
 `}`
If we try to rename `a` in line 3 or 4 , Tern picks up fake ref from line 2 as well. 

Yet to be done
----------------

- [x] To enable context menu option for rename
- [ ] Update JS doc and minor code cleanup  
- [ ] Couple of bug fixes ( one mentioned above and another to center primary selection )


src/extensions/default/JavaScriptCodeHints/main.js
```diff
@@ -859,6 +859,62 @@ define(function (require, exports, module) {
...
         /*
+         * Handle Rename menu/keyboard command.
+         */
+        function handleRename() {
+            var offset,
+                handleFindRefs;
+
+
+            // Only provide rename when cursor is in JavaScript content
+            if (!session || session.editor.getModeForSelection() !== "javascript") {
```
petetnt: Can `session.editor` be `null` or `undefined`? Most likely not, but might be worth checking out.

src/extensions/default/JavaScriptCodeHints/ScopeManager.js
```diff
@@ -392,6 +392,40 @@ define(function (require, exports, module) {
...
+            fileInfo: fileInfo,
+            offset: offset
+        });
+
+        return addPendingRequest(fileInfo.name, offset, MessageIds.TERN_REFS);
+    }
+    
+    function requestFindRefs(session, document, offset) {
+        var path    = document.file.fullPath,
+            fileInfo = {type: MessageIds.TERN_FILE_INFO_TYPE_FULL,
```
petetnt: Nit: Have we used some kind of style for multiline objects? The indentation looks a bit weird, I would prefer
```js
fileInfo = {
  type: MessageIds.TERN_FILE_INFO_TYPE_FULL,
  name: path,
  offsetLines: 0,
  text: filterText(session.getJavascriptText())
};

File: src/extensions/default/JavaScriptCodeHints/main.js [hunk 3]
```diff
@@ -859,6 +859,62 @@ define(function (require, exports, module) {
...
+
+            /**
+             * Make a find ref request.
+             * @param {Session} session - the session
+             * @param {number} offset - the offset of where to jump from
+             */
+            function requestFindRefs(session, offset) {
+                var response = ScopeManager.requestFindRefs(session, session.editor.document, offset);
+
+                if (response.hasOwnProperty("promise")) {
```
petetnt: Is there are case where response wouldn't have a `promise`? `requestFindRefs` looks like it's pretty null safe, but might be good to guard with `response && response.hasOwn...`

src/extensions/default/JavaScriptCodeHints/main.js
```diff
@@ -859,6 +859,62 @@ define(function (require, exports, module) {
...
+                    response.promise.done(handleFindRefs).fail(function () {
```
petetnt: Can we use `then` and `catch` to align the old Deferred syntax with the more modern Promises/A+ syntax?

src/extensions/default/JavaScriptCodeHints/main.js
```diff
@@ -859,6 +859,62 @@ define(function (require, exports, module) {
...
+                }
+            }
+
+            /**
+             * handle processing of the completed jump-to-def request.
+             * will open the appropriate file, and set the selection based
+             * on the response.
+             */
+            handleFindRefs = function (refsResp) {
+                function isInSameFile(obj) {
```
petetnt: Can be shortened to `return obj && obj.file === refsResp.file`

src/extensions/default/JavaScriptCodeHints/main.js
```diff
@@ -859,6 +859,62 @@ define(function (require, exports, module) {
...
+                    if (obj && obj.file === refsResp.file) {
+                        return true;
+                    } else {
+                        return false;
+                    }
+                }
+                
+                if (refsResp && refsResp.refs && refsResp.refs.refs) {
```
petetnt: Random comment: `refsResp.refs.refs` looks kinda silly 👯

src/extensions/default/JavaScriptCodeHints/tern-worker.js
```diff
@@ -231,6 +231,27 @@ var config = {};
...
+    
+            function getRefs(fileInfo, offset) {
+                var request = buildRequest(fileInfo, "refs", offset);
+                try {
+                    ternServer.request(request, function (error, data) {
+                        if (error) {
+                            _log("Error returned from Tern 'definition' request: " + error);
+                            return;
+                        }
+                        var response = {type: MessageIds.TERN_REFS,
```
petetnt: NIT: Same thing here with the new lines and objects

petetnt (changes requested):
Some initial comments here. Great work so far!

swmitra:
Thanks a lot @petetnt for the quick review. I will try to address the comments asap. I would urge all of you to use the feature as well and let me know if it's helpful.

src/base-config/keyboard.json
```diff
@@ -329,5 +329,8 @@
     ],
     "help.support": [
         "F1"
+    ],
+    "refactor.rename": [
+        "Ctrl-R"
```
ficristo: I believe on macOS this is translated to Cmd+R which is used to reload Brackets with Extensions.

src/extensions/default/JavaScriptCodeHints/main.js
```diff
@@ -867,6 +923,8 @@ define(function (require, exports, module) {
 
             return response;
         }
+        
+        CommandManager.register("Rename", "refactor.rename", handleRename);
```
ficristo: You should add the string in `command/Commands.js` instead of using the string explicit.

ficristo (changes requested):

ficristo:
From the code, this seems to work only for javascript files. I assume languages that derive from javascript in CodeMirror are NOT supported (for example json, typescript, ...)
Is it possible for extensions to easily override this behaviour so to support other languages?

swmitra:
@ficristo Yes, it's targeted for javascript mode.  So if CM identifies any block of code as javascript,  this would work there.  That includes script blocks in html,  any modes derived from JS by CodeMirror.  BTW,    is there any extension available today which implements this feature?

swmitra:
@ficristo Yes, it's targeted for javascript mode.  So if CM identifies any block of code as javascript,  this would work there.  That includes script blocks in html,  any modes derived from JS by CodeMirror.  BTW,    is there any extension available today which implements this feature?

abose:
@swmitra  Can you post a gif of how this works?

swmitra:
@abose I am not able to post a gif right now as I dont have access to any computer for a few days.  But what you can do is - put the cursor on top of the token you want to rename.  Then press ctrl+r . This will put selection on all the occurrences after doing a scope analysis.  Now just type the new name.  You will see multi cursor in action and should be able to rename all the occurrences.

abose:
👍 Cool feature

ficristo:
> is there any extension available today which implements this feature?

Looking at #11658 maybe [ternific](https://github.com/MiguelCastillo/Brackets-Ternific)?

PR #13055, opened 2017-01-20
Update index.html

jiajun0308:
There is a spelling mistake at line '157' where "ideas" was misspelled as "ieda". I changed the word to traditional chinese to match with the contents.


petetnt:
Thanks!

PR #13080, opened 2017-02-06
Added Extension Sorting capability according to downloads and last published date

saurabh95:
Now on opening the Extension Manager, a drop-down is shown on the left side of the search bar, which indicates the sorting of extensions according to the selected criteria(currently sort methods available are  Last Updated and Download Count), on change of the selection we store the the changed selection in the preferences and if the Extension Manager is invoked again, then we use the same sorting criteria which was selected by user earlier. The sorting is available for Available and Theme Extensions.
I have attached a snapshot of the feature in use where the sort order and the download count has been highlighted for illustration. 

downloadcount

### DIFF
--- a/src/extensibility/ExtensionManagerViewModel.js
+++ b/src/extensibility/ExtensionManagerViewModel.js
@@ -28,10 +28,11 @@ define(function (require, exports, module) {
 
     var _ = require("thirdparty/lodash");
 
-    var ExtensionManager = require("extensibility/ExtensionManager"),
-        registry_utils   = require("extensibility/registry_utils"),
-        EventDispatcher  = require("utils/EventDispatcher"),
-        Strings          = require("strings");
+    var ExtensionManager    = require("extensibility/ExtensionManager"),
+        registry_utils      = require("extensibility/registry_utils"),
+        EventDispatcher     = require("utils/EventDispatcher"),
+        Strings             = require("strings"),
+        PreferencesManager  = require("preferences/PreferencesManager");
 
     /**
      * @private
@@ -293,12 +294,12 @@ define(function (require, exports, module) {
     };
 
     ExtensionManagerViewModel.prototype._setSortedExtensionList = function (extensions, isTheme) {
-        this.filterSet = this.sortedFullSet = registry_utils.sortRegistry(extensions, "registryInfo")
+        this.filterSet = this.sortedFullSet = registry_utils.sortRegistry(extensions, "registryInfo", PreferencesManager.get("extensions.sort"))
             .filter(function (entry) {
                 if (!isTheme) {
-                    return entry.registryInfo !== undefined && !isTheme && entry.registryInfo.metadata.theme === undefined;
+                    return entry.registryInfo && !entry.registryInfo.metadata.theme;
                 } else {
-                    return entry.registryInfo !== undefined && entry.registryInfo.metadata.theme;
+                    return entry.registryInfo && entry.registryInfo.metadata.theme;
                 }
             })
             .map(function (entry) {
--- a/src/extensibility/ExtensionManagerViewModel.js
+++ b/src/extensibility/ExtensionManagerViewModel.js
@@ -30,7 +30,7 @@ define(function (require, exports, module) {
 
     var ExtensionManager = require("extensibility/ExtensionManager"),
         registry_utils   = require("extensibility/registry_utils"),
-        EventDispatcher = require("utils/EventDispatcher"),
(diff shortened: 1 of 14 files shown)


src/extensibility/ExtensionManagerViewModel.js
```diff
@@ -292,6 +292,20 @@ define(function (require, exports, module) {
             });
     };
 
+    ExtensionManagerViewModel.prototype._setSortedExtensionList = function (extensions, isTheme) {
+        this.filterSet = this.sortedFullSet = registry_utils.sortRegistry(extensions, "registryInfo")
+            .filter(function (entry) {
```
petetnt: There seems to be some redundancy here, seems like same could be achieved with:

``` js
return entry.registryInfo && entry.registryInfo.metadata;
```

Do correct me if I am wrong 👍

petetnt (commented):
Mostly LGTM, needs tests 👍

ficristo:
Nice! I think @zaggino and @dnbard could have some opinions on this.
See @dnbard extension: https://github.com/dnbard/brackets-extension-rating

I have two doubts:
- why we should threat the installed tab differently?
- `registry_utils` comes from brackets-registry: I guess it is necessary to change it but it seems to me it should be more "self contained" (= don't `require` Brackets code)

saurabh95:
@ficristo the installed tab is treated differently as we do not have the download count of the installed extensions in the installInfo, and also currently the sorting of installed extensions is done according to the update available.
Yeah we can isolate the registry_utils from Brackets code, we can pass the preference as the argument.

src/extensibility/ExtensionManagerViewModel.js
saurabh95: yeah that can be changed :)

saurabh95:
@ficristo @petetnt Addressed review comments and also added some tests.

petetnt:
Great stuff @saurabh95!

Some issues I spotted:

1. For me the extension manager opens on Installed page and the sort by-toggle is visible (and does nothing). Needs similar check to https://github.com/adobe/brackets/pull/13080/files#diff-3491ce7bb262eb073c8aabc3d3e064c0R376 when opening the window too.

2. The sort toggle breaks the layout with really narrow windows (but the search bar does it too). Feel free to disregard / bump to another issue.
image

3. The icons for download counts don't seem to work on Windows (10)
image

test/spec/ExtensionManager-test.js
```diff
@@ -829,6 +838,12 @@ define(function (require, exports, module) {
                 it("should start with the full set sorted in reverse publish date order", function () {
                     expect(model.filterSet).toEqual(["theme-1", "theme-2"]);
                 });
+                
+                it("should start with the full set sorted in reverse download count order", function () {
+                    PreferencesManager.set("extensions.sort", "downloadCount");
+                    model._setSortedExtensionList(ExtensionManager.extensions, true);
+                    expect(model.filterSet).toEqual(["theme-2", "theme-1"]);
+                });
```
petetnt: A simple check could be added that Brackets actually renders the `totalDownloads` to the UI 👍

petetnt (changes requested):
Reviewed again, couple of small issues, otherwise LGTM 👍

src/extensibility/registry_utils.js
```diff
@@ -136,8 +136,16 @@ define(function (require, exports, module) {
             sortedEntries.push(registry[key]);
         });
         sortedEntries.sort(function (entry1, entry2) {
-            return getPublishTime((subkey && entry2[subkey]) || entry2) -
-                getPublishTime((subkey && entry1[subkey]) || entry1);
+            if (sortBy !== "publishedDate") {
```
ficristo: Not a big deal, but in general I prefer if you spell out all cases:
```js
if (sortBy === "downloadCount") {
} else if (sortBy === "publishedDate") {
} else {
  console.error("Unknown sortBy " + sortBy ". Defaults to....")
}
```
Or something like this.

ficristo (approved):
I didn't test this so I trust @petetnt.
But in general seems good.

saurabh95:
@petetnt I have added a test for checking downloadCount in view and also added a check for active tab on launch of Extension Manager to hide/show the sort option, regarding the download icon, I will replace it with an image.

swmitra (approved):
Good stuff @saurabh95 👍

saurabh95:
A screenshot with the new icon. 

untitled

swmitra:
Merging Now with the new Icon.

PR #13091, opened 2017-02-10
Sync code hint font size with code view font size

swmitra:
This PR enhances Brackets user experience by syncing the code hint font-size with code view font size. This was a problem in high dpi monitors ( WQHD and above ), where code hints were not readable with the current font-size. With this PR , when user changes code view font size, code hint font size also gets changed accordingly.

Without this PR, code hints are rendered as - 

hdpihint

With the changes in PR - 

hdpihintfix

### DIFF
--- a/src/view/ViewCommandHandlers.js
+++ b/src/view/ViewCommandHandlers.js
@@ -21,6 +21,8 @@
  *
  */
 
+/*global less */
+
 /**
  * The ViewCommandHandlers object dispatches the following event(s):
  *    - fontSizeChange -- Triggered when the font size is changed via the
--- a/src/view/ViewCommandHandlers.js
+++ b/src/view/ViewCommandHandlers.js
@@ -72,6 +72,12 @@ define(function (require, exports, module) {
      * @type {string}
      */
     var DYNAMIC_CODEHINT_FONT_STYLE_ID = "codehint-dynamic-fonts";
+    
+    /**
+     * @const
+     * @type {string}
+     */
+    var DYNAMIC_CODEHINT_WITH_TYPE_STYLE_ID = "codehint-with-type-dynamic-width";
 
     /**
      * @const
@@ -110,6 +116,14 @@ define(function (require, exports, module) {
      * @type {string}
      */
     var DEFAULT_FONT_FAMILY = "'SourceCodePro-Medium', ＭＳ ゴシック, 'MS Gothic', monospace";
+    
+    /**
+     * @const
+     * @private
+     * The default hint span width for hints with annotated type details
+     * @type {number}
+     */
+    var DEFAULT_TYPED_HINT_SPAN_WIDTH = 300;
 
     /**
      * @private
@@ -119,19 +133,22 @@ define(function (require, exports, module) {
     function _removeDynamicProperty(propertyID) {
         $("#" + propertyID).remove();
     }
-
+    
     /**
      * @private
      * Add the style property to the DOM
      * @param {string} propertyID Is the property ID to be added
-     * @param {string} name Is the name of the style property
-     * @param {string} value Is the value of the style
-     * @param {boolean} important Is a flag to make the style property !important
+     * @param {object} ruleCfg Is the CSS Rule configuration object
+     * @param {string} ruleCfg.propName Is the name of the style property
+     * @param {string} ruleCfg.propValue Is the value of the style property
+     * @param {boolean} ruleCfg.priorityFlag Is a flag to make the style property !important
+     * @param {string} ruleCfg.ruleName Optional Selctor name to be used for the rule
+     * @param {string} ruleCfg.ruleText Optional selector definition text
      */
-    function _addDynamicProperty(propertyID, name, value, important, cssRule, cssText) {
-        cssRule = cssRule || ".CodeMirror";
+    function _addDynamicProperty(propertyID, ruleCfg) {
+        var cssRule = ruleCfg.ruleName || ".CodeMirror";
         var $style   = $("<style type='text/css'></style>").attr("id", propertyID);
-        var styleStr = cssText || StringUtils.format("{0}: {1} {2}", name, value, important ? " !important" : "");
+        var styleStr = ruleCfg.ruleText || StringUtils.format("{0}: {1} {2}", ruleCfg.propName, ruleCfg.propValue, ruleCfg.priorityFlag ? "!important" : "");
         $style.html(cssRule + "{ " + styleStr + " }");
 
         // Let's make sure we remove the already existing item from the DOM.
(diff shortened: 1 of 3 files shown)


ficristo:
In case of JS hints the icon doesn't look good.

src/view/ViewCommandHandlers.js
```diff
@@ -122,10 +128,10 @@ define(function (require, exports, module) {
      * @param {string} value Is the value of the style
      * @param {boolean} important Is a flag to make the style property !important
```
ficristo: `cssRule` and `cssText` aren't documented.

src/view/ViewCommandHandlers.js
```diff
@@ -122,10 +128,10 @@ define(function (require, exports, module) {
...
      */
-    function _addDynamicProperty(propertyID, name, value, important, cssRule) {
+    function _addDynamicProperty(propertyID, name, value, important, cssRule, cssText) {
         cssRule = cssRule || ".CodeMirror";
         var $style   = $("<style type='text/css'></style>").attr("id", propertyID);
-        var styleStr = StringUtils.format("{0}: {1}{2}", name, value, important ? " !important" : "");
+        var styleStr = cssText || StringUtils.format("{0}: {1} {2}", name, value, important ? " !important" : "");
```
ficristo: I believe you can remove the space between this two parameters `{1} {2}`.

src/view/ViewCommandHandlers.js
```diff
@@ -139,6 +145,18 @@ define(function (require, exports, module) {
      */
     function _removeDynamicFontSize() {
         _removeDynamicProperty(DYNAMIC_FONT_STYLE_ID);
+        _removeDynamicProperty(DYNAMIC_CODEHINT_FONT_STYLE_ID);
+    }
+
+    /**
+     * @private
+     * Adds a new embeded style top sync code-hint font size with codeview font size
```
ficristo: `fontSize` is not documented

src/view/ViewCommandHandlers.js
```diff
@@ -139,6 +145,18 @@ define(function (require, exports, module) {
...
+     */
+    function _addDynamicFontSizeForCodeHints(fontSize) {
+        var styleStr = "";
+        styleStr = styleStr + StringUtils.format("{0}: {1} {2};", "font-size", fontSize, " !important");
+        styleStr = styleStr + StringUtils.format("{0}: {1} {2};", "line-height", (parseInt(fontSize, 10) + 2) + fontSize.replace(parseInt(fontSize, 10), ""), " !important");
+        _addDynamicProperty(DYNAMIC_CODEHINT_FONT_STYLE_ID, "font-size", fontSize, true, ".codehint-menu .dropdown-menu li a", styleStr);
```
ficristo: You need to pass the values `font-size`, `fontSize` and `true` but they aren't actually used.
Perhaps it's personal style but it doesn't seems good.

ficristo (changes requested):
I would like to see at least the documentation fixed.

src/view/ViewCommandHandlers.js
petetnt: I think `function _addDynamicProperty(propertyID, options)` would make more sense (where options would be an object of `{ name, value, important, cssRule, cssText }` etc.)

swmitra:
Thanks a lot @ficristo for the quick review 👍  I will update the PR soon.

MiguelCastillo:
Question around the approach.  Adding the a method specific for code hints seems reasonable, but I am not sure how that will scale if we had to add more methods for different parts of the UI.  Is there a way we can leverage CSS for the fonts size/type?

swmitra:
@MiguelCastillo Good point to consider 👍. 
I had tried the most generic approach by increasing the base font size for body. In that way we can increase and sync the font size in other panes like project panel and Dialogs as well. But it's not convincing use case for every one. In fact most of the people I spoke with wanted to increase just the code view font size and nothing else. I will hold this PR for now and lets ask other developers about the actual feature scope itself. Please let me know your opinion about the feature scope.

zaggino:
I'd say that while @MiguelCastillo 's point is valid, we should not get into the habit of stacking feature on top of feature in PRs which makes them not only very large but also increases the time which they take to merge, sometimes are these even never merged as authors get discouraged. So if @swmitra 's PR works and solves the problem it was meant to solve (without introducing new ones), we should merge it and open further suggestions as separate issues.

swmitra:
@ficristo We no longer use that icon in JS code hints. I am a bit confused as I don't see those icons in JS code hints, Can you please confirm once more?

ficristo:
@swmitra you are right. After I had disabled my extensions I don't see it anymore.

swmitra:
Thanks a lot @ficristo for quick confirmation 👍 . Can you please have a look at the changes again as I have tried to address the comments? @petetnt I have wrapped the parameters in a configuration object as per your suggestion, can you please have a look?

swmitra:
@MiguelCastillo Can you please check the implementation now? I just tried to make it extensible by using a less template.

MiguelCastillo:
yup i am checking now.  BTW - thank you for clarifying the use case.  It definitely helped me understand the scope of what we are accomplishing.  I am happy with what we got here :)

src/view/fontrules/font-based-rules.less
```diff
@@ -0,0 +1,24 @@
...
+    font-size: @fontsize !important;
+    line-height: @lineheight  !important; 
+}
+
+span.brackets-js-hints-with-type-details { 
+    width: @defaultmenuwidth * (@fontsize / @defaultfontsize) !important; 
+}
+
+.codehint-menu .dropdown-menu { 
+    max-height: @maxvisibleitems * @lineheight !important;
```
MiguelCastillo: how does line height affect this when the fontsize is define in `px` vs `em`?

MiguelCastillo:
I like what you have done here!  very interesting!  I just had a question regarding the max-height value and how it is affected when font size is specified with different units.

swmitra:
@MiguelCastillo I had made core template calculations unit agnostic. Also I had to introduce an additional step to convert only em to px. Rest all units like pt, rem and px works well with this implementation. 
Thanks a lot for bringing the point up as it finally resulted in a better implementation 👍.

Please check the code and try out the implementation and let me know if see any issues.

MiguelCastillo:
@swmitra  yeah - this is good to merge for me.  I am not sure if there are any TODOs pending, are there?

swmitra:
@ficristo Can you please have a look now?

ficristo:
@MiguelCastillo review is more than enough for me.

ficristo (approved):

chylex:
Is there any way to revert this? It feels like I can see way fewer hints than I used to, and any code hints that include documentation look misaligned, so I'd rather go back to the default font size.

PR #13103, opened 2017-02-14
Added Reverse Inspect feature in MultiBrowserLive Preview using the existing Websocket transport architecture

saurabh95:
- [ ] Will add tests for reverse inspect in both Live Preview and Multibrowser Live Preview

### DIFF
--- a/src/LiveDevelopment/Agents/RemoteFunctions.js
+++ b/src/LiveDevelopment/Agents/RemoteFunctions.js
@@ -22,7 +22,7 @@
  */
 
 /*jslint forin: true */
-/*global Node */
+/*global Node, MessageEvent */
 /*theseus instrument: false */
 
 /**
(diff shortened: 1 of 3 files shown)


petetnt (approved):

src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -337,7 +337,7 @@ function RemoteFunctions(experimental, remoteWSPort) {
                 _trigger(element, "highlight", 1);
             }
             
-            if (!window.event && !isInViewport(element)) {
+            if ((!window.event || window.event instanceof MessageEvent) && !isInViewport(element)) {
```
saurabh95: check for `window.event instanceof MessageEvent` is added because in latest version of chrome, when it receives a message through websocket, then it considers it as window event and thus window.event is true, so we are bypassing such scenario by checking if the event is of type Message, then we can safely ignore this.

swmitra (approved):

swmitra:
@saurabh95 Please create a task to update and add tests for this feature post 1.9 release. Great work 👍

swmitra:
Merging...

PR #13105, opened 2017-02-16
Issue 1715

Blackweda:
Hi,

I am trying to solve the issue: https://github.com/mozilla/thimble.mozilla.org/issues/1715
It asked for nunjucks 2.4.3 to be added to both Brackets and Thimble.
It was already present in Thimble, but I didn't see it in Brackets.

I added it into the dependencies list, yet I'm unsure if that was a wise thing to do. Please consider using the
code and let me know if it works or not, and where to make changes elsewhere.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
         "anymatch": "1.3.0",
         "chokidar": "1.6.0",
         "lodash": "4.15.0",
-        "nunjucks": "^2.4.3",
+        "nunjucks": "2.4.3",
         "ws": "~0.4.31"
     },
     "devDependencies": {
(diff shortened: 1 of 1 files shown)


zaggino:
Hi, you're probably referencing a wrong issue in the description. Also, adding something to the dependency list doesn't do anything until you use it somewhere in the code.

ficristo:
Hi @Blackweda,
I think you have to open your PRs at https://github.com/mozilla/thimble.mozilla.org and at the mozilla fork of Brackets https://github.com/mozilla/brackets
We don't use nunjucks.

Blackweda:
@ficristo thanks, makes sense! will do.

PR #13108, opened 2017-02-17
download eslint package from npm and place it into default extensions

zaggino:
Related to https://github.com/adobe/brackets/issues/11984

As a part of this PR I'd like to make a change, that when `brackets-eslint` is found in `extensions/user` directory, it won't be loaded from `extensions/default` directory - I'd like your thoughts on that.


ficristo:
I think Atom added a new field in their package.json, packageDependencies, and use it to install their extensions. I think I prefer their approach.
For the linters, I still don't understand what Adobe wants to do...

zaggino:
@ficristo Can you shine a bit more light how these `packageDependencies` should work here? Happy to make changes to the PR

ficristo:
I didn't look at Atom implementation, but I think the usage is similar to the node dependencies.
In the package.json of Atom they have something like:
```json
"packageDependencies": {
  "atom-dark-syntax": "0.28.0",
  "atom-dark-ui": "0.53.0",
  "...": "..."
}
```
We could have something like:
```json
"dependencies": {
  "chokidar": "1.6.0",
  "...": "..."
},
"devDependencies": {
  "grunt": "0.4.5",
  "...": "..."
},
"packageDependencies": {
  "brackets-eslint": "1.0.0",
  "...": "..."
}
```
and use npm to install them.
If I understood correctly, you have done this but hardcoded the dependencies in the scripts.
I only think we have to move them on the package.json so to make easier to keep track of what we use.
Of course is totally possible there are some drawback that ATM I don't see.

zaggino:
@ficristo check now

ficristo:
I only skimmed the changes. From what I understood ESLint will installed in the default extensions so users will not able to remove this extension. Is it right?
I prefer if this one will be reviewed by someone at Adobe, sorry.

zaggino:
I don't think that's an issue (same way they can't remove JSLint now nobody really uses), but also a good point, I could make a PR to display default extensions in extension manager and delete them.

redmunds:
@zaggino Do you mean that default extensions would be physically deleted? That might be problematic from install dir due to permissions, and how would user re-install them?

Another option is to list default extensions in extension manager (possibly in a new tab) and allow them to be disabled/re-enabled.

Note that Brackets has not been tested without the default extensions so different combinations should be tested.

zaggino:
@redmunds Yeah I mean't disable/enable as a configuration preference, not sure why I wrote delete.

If some extensions are critical for Brackets to run, we could hard-code that they can't be disabled, but I think most of them won't cause this sort of problems.

redmunds:
Very Cool. I don't think any of the default extensions were intended to be critical -- they just haven't been tested to verify it.

zaggino:
ref: https://github.com/adobe/brackets/pull/13136

zaggino:
default extensions can be now disabled in the UI, so this can take another look guys

ficristo:
I was wondering if instead of downloading from npm wouldn't it be better to use brackets registry?

swmitra:
@zaggino Great work here 👍 . I took the liberty to resolve the conflicts, as I wanted to test this branch. I looked at the code and it looks good to me. If everything works fine in real usage for a couple of days, I will merge this PR. Do you recall any pending item in this PR which I might have missed?

zaggino:
@swmitra The only thing to consider is that after this, it's higly likely that people will have `brackets-eslint` as a default extension and also as a user extension installed. Only on of those (the user installed version probably) should load when starting up.

swmitra:
I have modified the registration and fetching logic of registered providers a bit based on the following idiom. 

- For a given language, we would be picking up only a single language specific linter + any generic linter registered with `'*'` 
- Choosing a linter from the registered set of language specific linters depends on `priority` and load order.

@zaggino @ficristo @petetnt Can you please have a look at this PR urgently? This needs to be in for next release to provide basic es6 support ...

swmitra:
I can no longer review this PR, but definitely work on the PR 😄

zaggino:
I have written most of the code so not much to review.
@swmitra Where's the actual priority being set? The one you use for sorting?

swmitra:
The priority field is for future use. As of now it will default to 0 in the priority check and collision resolution will happen on load order.

swmitra:
@zaggino I should have been more precise. My mistake. I meant the last commit.  As the existing code is already reviewed, I wanted to focus just on the open part to handle conflict between same providers. In the process,  I thought of limiting the linting by only a single language specific linter. Looking at both EsLint and JSLint together made me skeptical about the overhead.

zaggino:
Not sure about the single linter, on my projects I use both TypeScript (with TSLint) and ESLint to lint my files. I wouldn't want to have only one.

swmitra:
@zaggino Thanks for sharing the multiple linter use case. Removed the priority based sorting now. Instead, we now look for providers of a given language having same provider name. If found, remove it and retain the most recently loaded one. Tested with same extension co-existence in Brackets core and user exetensions, and what I have noticed is, user extensions are the ones that used for linting if at all there are any name conflicts. 
In case someone wants to disable JSLint, it can be done now by just disabling the JSLint extension unders default extensions tab in Extension manager.

swmitra:
@petetnt @ficristo Can you please review this PR now? 🙏

swmitra:
@nethip Can you please review this PR? 🙏

npm-shrinkwrap.json
```diff
@@ -4790,9 +4819,9 @@
       }
     },
     "xmldoc": {
-      "version": "0.1.4",
-      "from": "xmldoc@>=0.1.2 <0.2.0",
-      "resolved": "https://registry.npmjs.org/xmldoc/-/xmldoc-0.1.4.tgz",
+      "version": "0.1.2",
```
nethip: Version downgrade?

npm-shrinkwrap.json
```diff
@@ -4236,9 +4259,9 @@
       "resolved": "https://registry.npmjs.org/safe-buffer/-/safe-buffer-5.1.1.tgz"
     },
     "sax": {
-      "version": "0.6.1",
-      "from": "sax@>=0.6.1 <0.7.0",
-      "resolved": "https://registry.npmjs.org/sax/-/sax-0.6.1.tgz",
+      "version": "0.4.2",
```
narayani28: is this an intended downgrade of the version?

nethip:
@swmitra Code wise I am fine. I have taken the liberty in adding check boxes for pending items, so that it is easier to track the progress.

- [ ]  Build this branch on all the three platforms (Win, Mac, Linux), locally and see that you are able to succesfully use the `www` folder, built from this branch.
- [ ]  Run smokes on all the platforms, as the library versions have changed for some
- [ ]  Trigger installer builds on our build system to see for all the platforms if the build is going fine
- [ ]  Have a look out for app footprint size
- [ ]  Update documentation wherever required

narayani28 (approved):
LGTM

saurabh95 (approved):

petetnt (commented):
FWIW LGTM, bit too late though. Nice work!

PR #13121, opened 2017-02-24
Issue# 1596: Added the ability to add a tutorial after it has been removed

### LINKED ISSUE #1596: Tutorial creation option in file-tree is not available after when tutorial.html is deleted
After when tutorial.html is deleted, the tutorial creation option in the file tree is still unavailable to the user. Refreshing the page, however, fixes the issue. Similar to #1279.

I believe the issue to be with how the toggling between `tutorialAdded` and `tutorialRemoved` is handled. I have referenced the relevant code below:

https://github.com/mozilla/thimble.mozilla.org/blob/master/public/editor/scripts/editor/js/fc/bramble-menus.js#L210-L237

MargarytaChepiga:
This is the issue that I fixed: https://github.com/mozilla/thimble.mozilla.org/issues/1596
I followed the advice on fixing the bug, and simply added the code from the unlink case to the rm case. 
Here is the result:
image
image


ficristo:
Hi @MargarytaChepiga,
I believe you need to open your PR at https://github.com/mozilla/brackets

MargarytaChepiga:
Hey @ficristo, Thank you for noticing and letting me now, I am going to fix it ASAP.

PR #13123, opened 2017-02-24
#11801 - Restricted Font Size Input To Only Valid Entries

nyteksf:
Related to #11801.

Text was able to be inputted into Font Size within Themes Settings (e.g. "View > Themes...") even when it included non-ASCII characters or was in a generally unusable format.
 
Corrected this issue by ensuring all font sizes are to begin with a number, and are completed by either "px" or "em" only. 'Done' button remains disabled until said format is adhered to correctly--at which time the functionality of said button becomes restored allowing input to be saved.

@zaggino


src/config.json
```diff
@@ -20,7 +20,7 @@
         "extension_url": "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default": true,
         "build_timestamp": "",
-        "healthDataServerURL": "https://healthdev.brackets.io/healthDataLog"
+        "healthDataServerURL": "https://health.brackets.io/healthDataLog"
```
zaggino: you need to revert this change (it applies automatically after doing npm install, but we don't want it in PRs)

src/view/ThemeSettings.js
```diff
@@ -100,7 +100,24 @@ define(function (require, exports, module) {
                 var attr = $target.attr("data-target");
                 newSettings[attr] = $target.is(":checked");
             })
-            .on("input", "[data-target]:text", function () {
+            .on("input", "#font-size-input", function () {
+                var $target = $(this);
+                var $input = $target.val();
+                var validInput = /^[\d\.]+(p|px|e|em){0,1}$/g;
+                var btn = document.getElementById("done");
```
zaggino: add `document` to the globals please: https://github.com/adobe/brackets/blob/master/.eslintrc.json#L52

src/htmlContent/themes-settings.html
```diff
@@ -30,20 +30,20 @@ <h1 class="dialog-title">{{Strings.THEMES_SETTINGS}}</h1>
...
-                    <input type="text" data-target="fontFamily" value="{{settings.fontFamily}}">
+                    <input type="text" id="font-family-input" data-target="fontFamily" value="{{settings.fontFamily}}">
                 </div>
             </div>
         </form>
     </div>
     <div class="modal-footer">
         <button class="dialog-button btn" data-button-id="cancel">{{Strings.CANCEL}}</button>
-        <button class="dialog-button btn primary" data-button-id="save">{{Strings.DONE}}</button>
+        <button class="dialog-button btn primary" id="done" data-button-id="save">{{Strings.DONE}}</button>
```
zaggino: id `done` is not very good, you should be more specific when introducing id's, something like `theme-settings-done-btn` would be more appropriate

src/htmlContent/themes-settings.html
```diff
@@ -30,20 +30,20 @@ <h1 class="dialog-title">{{Strings.THEMES_SETTINGS}}</h1>
...
-                    <input type="text" data-target="fontSize" value="{{settings.fontSize}}">
+                    <input type="text" id="font-size-input" data-target="fontSize" value="{{settings.fontSize}}">
                 </div>
             </div>
 
             <div class="control-group">
                 <label class="control-label">{{Strings.FONT_FAMILY}}:</label>
                 <div class="controls">
-                    <input type="text" data-target="fontFamily" value="{{settings.fontFamily}}">
+                    <input type="text" id="font-family-input" data-target="fontFamily" value="{{settings.fontFamily}}">
```
zaggino: don't need to introduce id's here, you can use `.on("input", "[data-target=fontFamily]", function () {`
https://www.w3schools.com/jquery/sel_attribute_equal_value.asp

src/view/ThemeSettings.js
```diff
@@ -100,7 +100,24 @@ define(function (require, exports, module) {
                 var attr = $target.attr("data-target");
                 newSettings[attr] = $target.is(":checked");
             })
-            .on("input", "[data-target]:text", function () {
+            .on("input", "#font-size-input", function () {
+                var $target = $(this);
+                var $input = $target.val();
```
zaggino: `$` prefix for variables indicates that it's a jquery (or a DOM) element
in this case `$input` is a string and should really be named `input` or even better `targetValue`

src/view/ThemeSettings.js
```diff
@@ -100,7 +100,24 @@ define(function (require, exports, module) {
...
+                var validInput = /^[\d\.]+(p|px|e|em){0,1}$/g;
+                var btn = document.getElementById("done");
+
+                // Make sure that the font size is expressed in terms we can handle (px or em). If not, 'done' button disabled until input corrected.
+
+                if ($input.search(validInput) === -1) {
+                    ($input.length === 0) ? btn.disabled=false : btn.disabled=true;
```
zaggino: `btn.disabled = $input.length !== 0` is better

src/view/ThemeSettings.js
```diff
@@ -100,7 +100,24 @@ define(function (require, exports, module) {
...
+                    return false;
+                }else{
```
zaggino: `}else{` after `return` doesn't make sense, just close the if with `}` and continue normally

zaggino (changes requested):
just a first quick look without testing, let me know when you'll look at the comments

zaggino:
also make sure command `grunt eslint` in your branch doesn't throw any errors

.eslintrc.json
```diff
@@ -57,6 +57,7 @@
         "$": false,
 
         "window": false,
+        "document": false,
```
ficristo: Sorry for mixed signals, but please revert this.
`document` is a generic word and in Brackets usually refer to the Document object.
I think you can go with jquery here, can't you? `$("#theme-settings-done-btn")`

ficristo:
Out of curiosity, any reason to not use the pattern attribute?

mohanad1980:
hello can you contact me mohanad.rahimi1@gmail.com

.eslintrc.json
nyteksf: Sorry for the delay--I've been busy with moving, although wanted to finish and reply. That said,, yes, I can use jQuery here. Good point. And thanks for clarifying. This is my first pull request; let alone open source patch. I wasn't sure if I should use jQuery as little as possible as noobish as that sounds. This turned out to be a simple fix.  :)

nyteksf:
Thanks, guys, for your speedy replies. To reply to you now myself: that HTML5 pattern attribute was actually wholly new to me. I've learned a bunch in a few reviewer comments. I have to agree with you both that "pattern" is much better. Thanks again. Please feel free to let me know if anything else would be problematic.

zaggino (approved):

zaggino:
fine with me, @ficristo ?

ficristo:
It seems fine for me.
Ideally we could expose this regex: https://github.com/adobe/brackets/blob/d6ae906810b8e6afb43c88edcf0397b1dc1b0119/src/view/ViewCommandHandlers.js#L300
and use it on the html.

zaggino:
@nyteksf to do what @ficristo wrote, you'd have to edit this file https://github.com/nyteksf/brackets/blob/9a6f12cf259276ebd163e70266030377df67eb2b/src/view/ViewCommandHandlers.js and make this https://github.com/adobe/brackets/blob/d6ae906810b8e6afb43c88edcf0397b1dc1b0119/src/view/ViewCommandHandlers.js#L300 an exported variable. 

Then here https://github.com/nyteksf/brackets/blob/9a6f12cf259276ebd163e70266030377df67eb2b/src/view/ThemeSettings.js#L82 you'd pass the `ViewCommandHandlers.validFontRegex` (or however you call it) to Mustache and render it here https://github.com/nyteksf/brackets/blob/9a6f12cf259276ebd163e70266030377df67eb2b/src/htmlContent/themes-settings.html#L33 like `{{ fontPattern }}` or something

Hope that's clear enough, @ficristo correct me if I'm wrong

petetnt:
LGTM, but we should accept at least `rem` values and percentages too, no?

zaggino:
Current regexp which is used deeper ( https://github.com/adobe/brackets/blob/d6ae906810b8e6afb43c88edcf0397b1dc1b0119/src/view/ViewCommandHandlers.js#L300 ) seems to only allow `px|em` ... do we want to change this?

nyteksf:
Thank you, guys. I have had to fiddle around with the RegExp a little to make the prescribed changes work. It seems backslashes need to be escaped--having tried the various remedies in the docs--and this causes the RegExp constructor used for the non-exported segment to be passed something unintended. Likewise, to use RegExp literally, it seems would require `.slice()`ing off the foreslashes. That said, I have switched it to `"[0-9.]"` instead of `"\d"`. But everything else should be as asked for to my awareness. Please let me know what you think in any case. Thanks very much. :)

src/htmlContent/themes-settings.html
```diff
@@ -30,7 +30,7 @@ <h1 class="dialog-title">{{Strings.THEMES_SETTINGS}}</h1>
             <div class="control-group">
                 <label class="control-label">{{Strings.FONT_SIZE}}:</label>
                 <div class="controls">
-                    <input type="text" data-target="fontSize" value="{{settings.fontSize}}">
+                    <input type="text" pattern="{{ settings.validFontSizeRegExp}}" data-target="fontSize" value="{{settings.fontSize}}">
```
zaggino: missing space before `}}`

src/view/ThemeSettings.js
```diff
@@ -68,9 +68,10 @@ define(function (require, exports, module) {
 
         result.fontFamily = ViewCommandHandlers.getFontFamily();
         result.fontSize   = ViewCommandHandlers.getFontSize();
+        result.validFontSizeRegExp = ViewCommandHandlers.validFontSizeRegExp;
         return result;
     }
-
+    var validFontSizeRegExp = ViewCommandHandlers.validFontSizeRegExp;
```
zaggino: constant definitions should always be on top

src/view/ThemeSettings.js
```diff
@@ -81,7 +82,7 @@ define(function (require, exports, module) {
         var template        = $("<div>").append($settings).html();
         var $template       = $(Mustache.render(template, {"settings": currentSettings, "themes": themes, "Strings": Strings}));
 
-        // Select the correct theme.
+// Select the correct theme.
```
zaggino: invalid whitespace change

src/view/ViewCommandHandlers.js
```diff
@@ -295,16 +295,19 @@ define(function (require, exports, module) {
      * @param {number} adjustment  Negative number to make the font smaller; positive number to make it bigger
      * @return {boolean} true if adjustment occurred, false if it did not occur
      */
-    function _adjustFontSize(adjustment) {
-        var fsStyle   = prefs.get("fontSize"),
-            validFont = /^[\d\.]+(px|em)$/;
-
-        // Make sure that the font size is expressed in terms we can handle (px or em). If not, simply bail.
-        if (fsStyle.search(validFont) === -1) {
+     var validFontSizeRegExpStr = "^[0-9.]+(px|em)$"; // Need RegExp string to be exported for compatibility with HTML5 pattern attribute.
```
zaggino: constant definitions should always be on top (after import declarations) and not between functions, also needs a comment, something like:
```
/*
 * Font sizes should be validated by this regexp
 */
```

zaggino (changes requested):
just a few small changes that are of stylistic nature and I believe we're good to merge this

nyteksf:
Whoops! Well, okay: there you go. Thanks again for a speedy reply!

@zaggino

zaggino:
Testing this, it allows input like:
`12.00.00px`
can we please change regexp to be maybe:
`"^[0-9]+(\.[0-9]+)?(px|em)$"`

zaggino:
Done with testing, just change this one and I'm merging this (promise 😄 )

nyteksf:
Okay. I'm going out on a limb here: please do correct me if I am wrong. I added your change to the original deeper RegExp. Then in doing my own testing, I had found that it worked almost perfectly. The RegExp in question, however, didn't handle cases like "`.5em`", without first adding a zero to make it like "`0.5em`". It isn't a problem, I'm sure. My change seems like a small change indeed--but I figured I could change it. And now it should handle cases like "`0.5em`" as well as "`.5em`". Please let me know if it works as it seems to on my end. And thanks again for all your help.  :)))

@zaggino

nyteksf:
Nevermind. I found a flaw. Changing back to original.

nyteksf:
Okay, @zaggino. My fingers are crossed this time. I'm sending my final changes over your way if you'll please take another look now.  :)

zaggino:
Oh, github new features are superb, did a small change to your regexp through the web - when inside a string you need to escape `\.` like this `\\.` - otherwise an expression `12a5px` would be valid, I've fixed that

zaggino:
Looks good to me now. I think all comments above have been considered so I'm merging this.
Thanks a lot for the PR a for sticking with it to the end, hope we'll meet again on some other one.
Cheers!

PR #13133, opened 2017-03-01
Fixed #11291 - Added a delay of 200ms when searching for extensions and now we are updating the current view only instead of updating all 3 views

### LINKED ISSUE #11291: Add small delay when searching for extensions
Hello there.

Now that Brackets has a lot of extensions (and there will be even more and more), using the text-field for searching for them in absolutely real-time is quite annoying, specially in those low-power computers... What about add a small delay (300-400 ms should be enough) between typing and filtering?

Type -> wait a few milliseconds -> run filter

This should bring the feeling of a more responsive and natural typing!

saurabh95:
Now we are updating the current view only instead of updating all 3 views as when we switch between views, then we call clear search(.filter("") is called) so there is no need of updating the other views on key-up.

### DIFF
--- a/src/extensibility/ExtensionManagerDialog.js
+++ b/src/extensibility/ExtensionManagerDialog.js
@@ -459,11 +459,11 @@ define(function (require, exports, module) {
             });
 
             // Filter the views when the user types in the search field.
-            var searchTimeOut = 0;
+            var searchTimeoutID;
             $dlg.on("input", ".search", function (e) {
-                clearTimeout(searchTimeOut);
+                clearTimeout(searchTimeoutID);
                 var query = $(this).val();
-                searchTimeOut = setTimeout(function () {
+                searchTimeoutID = setTimeout(function () {
                     views[_activeTabIndex].filter(query);
                     $modalDlg.scrollTop(0);
                 }, 200);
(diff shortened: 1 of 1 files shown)


petetnt (approved):
LGTM

src/extensibility/ExtensionManagerDialog.js
```diff
@@ -459,12 +459,14 @@ define(function (require, exports, module) {
             });
 
             // Filter the views when the user types in the search field.
+            var searchTimeOut = 0;
```
ficristo: Could you rename this to `searchTimeoutID` and remove the default value?

src/extensibility/ExtensionManagerDialog.js
```diff
@@ -459,12 +459,14 @@ define(function (require, exports, module) {
...
             $dlg.on("input", ".search", function (e) {
+                clearTimeout(searchTimeOut);
                 var query = $(this).val();
-                views.forEach(function (view) {
-                    view.filter(query);
+                searchTimeOut = setTimeout(function () {
+                    views[_activeTabIndex].filter(query);
                     $modalDlg.scrollTop(0);
-                });
+                }, 200);
```
ficristo: Could this be something like 100 or better 50 ms instead?

saurabh95: I tried those values, but there is still interruption in typing if it is 100 ms or 50 ms. I tried higher values also (like 150), but there is still interruption in typing, 200 ms worked just fine for me.

zaggino: 👍 I don't think 200ms is too much

zaggino (approved):

PR #13136, opened 2017-03-02
add a tab with default extensions to extension manager

zaggino:
Adds a new tab that allows disabling/enabling default extensions (this functionality is already available, so this PR is mostly about displaying the list of default extensions). 

Maybe we could use a different icon but I'm not capable of producing one and this can be replaced in a separate PR after this is merged.

Tagging @swmitra @redmunds to review as per discussion in #13108 but others feel free to chime in.

image


ficristo:
I like how things come back https://github.com/adobe/brackets/pull/11618
I gave a try and seems I cannot disable the default Dark theme or JSLint.

zaggino:
Interesting, did you try it on Mac or Windows @ficristo ?

zaggino:
The thing it came back again probably means we should finish it this time 😄

petetnt:
Nice @ficristo! I remembered that PR but couldn't find it myself.

👍 for the idea from me. Adding descriptions for the default extensions that don't have them would be nice.

redmunds:
I think disabling the Default Theme should not be allowed. Isn't that what gets used when you _Restart Without Extensions_ ?

Also, JSLint extension is enabled by default. Maybe code needs to verify existence of linting module before trying to use it?

zaggino:
Disabling themes (as it makes little sense) has been forbidden in https://github.com/adobe/brackets/pull/13136/commits/2d197fba7fcaf1281eba73146459efcbef78592d

zaggino:
Ok guys, did a few changes, please have a look. I've tested disabling/enabling JSLint and it works fine.
I've also tested disabling all the default extensions (except themes which now can't be disabled) and Brackets did startup just fine. I've added a warning to the console logs about the default extension being disabled. As it might help with bug reports.

redmunds:
@zaggino Built code, did minimal testing, and seems to work as advertised. Made a pass through code -- looks good. Awesome!

ficristo:
I tryed again and JSLint is disabled by default and I cannot enabled. Is it normal or it is maybe related to my system?
In the preferences I have: `"extensions.default.disabled": []`

zaggino:
@ficristo if you have tried previous version, look at JSLint's directory if there's not a `.disabled` filename present and delete it or run `git clean -fdx && npm install` ... I had the same issue and this fixed it.

ficristo:
Running `git clean` fixed it.
I wanted to disable JSLint only for Brackets so I added the preference to disable it in `.brackets.json`.
But JSLint is still available. Is this configuration not supported?

zaggino:
I don't think this could work per project @ficristo if that's what you're trying to do. The place where the extensions are being loaded does not consider which project you're in and neither does this get executed when you switch projects. While it wouldn't be hard to enable an extension when you switch to other project, it would be very difficult to disable one when you switch to a project which has it disabled in `.brackets.json` because disabling an extension basically means reloading whole editor. Now I know VSCode does that, maybe even Atom (reload whole editor when switching projects) but we currently don't and that'd be potentially huge change how Brackets work.

zaggino:
Any more comments @adobe/brackets-committers ?

zaggino:
@swmitra can anyone in your team review this please?

justinrusso:
With PR #13245 merged, the @extension-manager-min-width may need adjusting

swmitra:
Very nice @zaggino 👍. 
Tested this branch just now and everything seems to be working fine. I just have one small comment (But that would not block me to merge this PR now :smile), If we can cook some description for the default extensions to provide some meaningful info to the user based on which they can decide whether to disable any default extension!

swmitra (approved):
LGTM 👍

swmitra:
@zaggino Can you please increase the min-width of the EM Dialog to ~760px?

zaggino:
@swmitra where's this min-width set?

I can see that EM has the width hard set to 760px here: https://github.com/adobe/brackets/blob/master/src/styles/brackets_patterns_override.less#L1026-L1029

zaggino:
I'll add descriptions to the default extensions in another PR, to prevent ongoing discussions what it should be :)

zaggino:
ah, found it @swmitra @justinrusso please check if https://github.com/adobe/brackets/pull/13136/commits/d6cf7c07241fd5e39dc7fe0cc09a1bec7db954b8 is correct

swmitra (approved):

swmitra:
@zaggino Yes that's the place. I was about to reply but you are fast 😄 I have tested this feature and works as expected. Merging ...

PR #13137, opened 2017-03-02
Fixed #9642 - Added extension to the filename in SaveAs Dialog using the language specified in language Switcher

### LINKED ISSUE #9642: Brackets should save untitled files using the language switcher (or text content?) as a hint for file extensions
Save as option doesn’t save the file with .html extension even though the file type is set as .html in the file type drop down in the status bar. After saving as html, even if we type the valid html - html specific features like code coloring, code hinting doest work.

saurabh95:


src/document/DocumentCommandHandlers.js
```diff
@@ -952,6 +953,11 @@ define(function (require, exports, module) {
                 saveAsDefaultPath = FileUtils.getDirectoryPath(origPath);
             }
             defaultName = FileUtils.getBaseName(origPath);
+            var language = LanguageManager.getLanguageForPath(origPath);
+            var fileExtensions = language.getFileExtensions();
```
zaggino: I believe `getLanguageForPath` can return `null` which should be checked here.

zaggino (changes requested):

zaggino (approved):

swmitra (approved):

PR #13147, opened 2017-03-03
Remove normalizeFilename function

ficristo:
Quick hack for https://github.com/adobe/brackets/issues/12744, I doubt this is the correct way to fix.
I am only interested to see if this could solve the crashes people are experiencing.

/cc @zaggino

### DIFF
--- a/src/extensions/default/JavaScriptCodeHints/tern-worker.js
+++ b/src/extensions/default/JavaScriptCodeHints/tern-worker.js
@@ -113,6 +113,13 @@ var config = {};
                 return fileName;
             }
 
+            function _getDenormalizedFilename(fileName) {
+                if (ternServer.projectDir && fileName.indexOf(ternServer.projectDir) === 0) {
+                    fileName = fileName.slice(ternServer.projectDir.length);
+                }
+                return fileName;
+            }
+
             /**
              * Create a new tern server.
              *
@@ -588,10 +595,7 @@ var config = {};
              * @param {string} path     - the path of the file
              */
             function handlePrimePump(path) {
-                var fileName = path;
-                if (ternServer.projectDir && fileName.indexOf(ternServer.projectDir) === 0) {
-                    fileName = fileName.slice(ternServer.projectDir.length);
-                }
+                var fileName = _getDenormalizedFilename(path);
                 var fileInfo = createEmptyUpdate(fileName),
                     request = buildRequest(fileInfo, "completions", {line: 0, ch: 0});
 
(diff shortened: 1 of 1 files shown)


zaggino:
This construct
```
var fileName = fileInfo.name;
if (ternServer.projectDir && fileName.indexOf(ternServer.projectDir) === -1) {
  fileName = ternServer.projectDir + fileName;
}
```
repeats itself if I have counted right and not missed any difference 5 times over the diff, perhaps
```
function getNormalizedFilename(fileInfo, ternServer) {
  return ...
}
```
and use `var fileName = getNormalizedFilename(fileInfo, ternServer);` would be a bit nicer

ficristo:
Sure! If this fixes the crashes I'll refactor like that.

zaggino:
Anyway, were you able to replicate anything, or is this just a hinch? If it doesn't break any tests I think we could merge this for time being as https://github.com/adobe/brackets/pull/11948 doesn't seem to be moving much...

ficristo:
Since people repored crashes in 1.8 that weren't in 1.8 this is the only thing that comes to my mind.
So yes, this is only an hinch...
The tests pass.

zaggino:
I have a project which, while working on it, definitely crashes Brackets, but I can't replicate this. Sometimes after 5 minutes, sometimes after an hour. So it'd vote the approach of merging it and seeing if there's any change in the reports.

zaggino (changes requested):
refactor before merge

ficristo:
Refactored. After running tests again, they are quite flaky. But I'm not sure if it is because our testsuite or because my PR...

src/extensions/default/JavaScriptCodeHints/tern-worker.js
```diff
@@ -587,14 +588,24 @@ var config = {};
              * @param {string} path     - the path of the file
              */
             function handlePrimePump(path) {
-                var fileInfo = createEmptyUpdate(path),
+                var fileName = path;
+                if (ternServer.projectDir && fileName.indexOf(ternServer.projectDir) === 0) {
```
zaggino: if you don't mind one last change, for future readability, I'd extract this to `_getDenormalizedFilename` as this basically reverts `_getNormalizedFilename` if I'm not mistaken

zaggino:
Hi @ficristo , I've run the tests and they all pass on my machine so that thing is fine.
Just one minor thing (commented above) and I'd merge this if nobody doesn't want to add anything.

src/extensions/default/JavaScriptCodeHints/tern-worker.js
ficristo: Done

zaggino:
👍

Issue #13061, opened 2017-01-23
Suggestion: Add a keybind command to swap the files currently open in upper and lower pane.


holyRa:
I like to have the upper pane much larger than the lower, so I can take a quick look at another file than the one I'm working on without losing much space on the screen. However, if I decide to work on the file I have open in the lower pane while quick-referencing the one on my upper pane, I have to use the mouse to swap the two files around.

Needless to say, this is an atrocity. I'd like to suggest adding a command to swap the two.

petetnt:
👍 Calling it an atrocity is a _bit_ too harsh, but I agree that it could be much better. Would you like to take a stab, checking out https://github.com/adobe/brackets/blob/e8bbf5ef8ce08c0d37f8e0218ed0127a1217fe26/src/view/Pane.js#L241 would be a good place to start.

holyRa:
I'm aware it was a bit too harsh, it was intended sarcastically as it's kind of a petty complaint in the big picture (oh no, I have to move my mouse). Sorry if it sounded rude, it wasn't intended as such :)

spiritson:
@petetnt  Should it be an extension instead? or would you like it be part of core? I see something very very close to this request already part of core from this #12853

spencerflem:
I'm thinking of taking this bug. Would you recommend moving the function to MainViewManager or leaving it in Pane?

swmitra:
I guess if we are making this part of core, it should be enabled only by a preference which by default should be false.

petetnt:
IMHO the fix should be flipping files between panes via buttons: the functionality is already there, but there's no keybindings for the flipview functionality. If we had the flipview functionality mapped to keybindings, SameDocMultiplePanes + FlipView would solve this problem (instead of just flipping two documents around).

swmitra:
Yes @petetnt that makes complete sense.

SteveMieskoski:
I'm thinking about picking this up if there are no objections.  @petetnt, I looked into mapping the keybindings to the functionality you mentioned but didn't find the 'SameDocMultiplePanes' functionality you mentioned.  I have something partially together which swaps the views but will wait to hear back.

zaggino:
@SteveMieskoski go ahead and reference your PR here so we can track the progress, thanks!

genjurosama:
Hello everyone i'm new here and wondering why this issue is still open .

natraj-s:
[...]

PR #12853, opened 2016-10-26
Add a feature to toggle between panes in the core #10555

arthur801031:
- I'm not sure if I need to add a menu item for this feature inside "Navigate".
- I'm also not sure if I placed the code inside the correct files since I'm new to open source project contributions. 

Thank you!

PR #13150, opened 2017-03-05
command to swap visible files between panes per issue #13061

SteveMieskoski:
per issue #13061 , I added a function to swap visible files between panes, keybinding to Ctrl + Alt + F , and a preference with a disabled default state. 
 The function itself is currently located within MainViewManager.js with the command and preference manager registration located in ViewCommandhandlers.js as it seemed the logical place.  There are three tests  covering the basic functionality, the case where the file to move already exists in the destination pane, and where only one pane contains open/visible files.  In the latter case I just call moveView.  

If its felt this should not exist in the core, but in an extension I don't think that will be much of an issue. I'm interested in any Comments and/or Remarks.

### DIFF
--- a/src/config.json
+++ b/src/config.json
@@ -20,7 +20,7 @@
         "extension_url": "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default": true,
         "build_timestamp": "",
-        "healthDataServerURL": "https://health.brackets.io/healthDataLog"
+        "healthDataServerURL": "https://healthdev.brackets.io/healthDataLog"
     },
     "name": "Brackets",
     "version": "1.9.0-0",
--- a/src/config.json
+++ b/src/config.json
@@ -20,7 +20,7 @@
         "extension_url": "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default": true,
         "build_timestamp": "",
-        "healthDataServerURL": "https://healthdev.brackets.io/healthDataLog"
+        "healthDataServerURL": "https://health.brackets.io/healthDataLog"
     },
     "name": "Brackets",
     "version": "1.9.0-0",
@@ -74,7 +74,8 @@
     },
     "scripts": {
         "postinstall": "grunt install",
-        "test": "grunt cla-check-pull test"
+        "test": "grunt test cla-check-pull",
+        "eslint": "grunt eslint"
     },
     "licenses": [
         {
--- a/src/view/MainViewManager.js
+++ b/src/view/MainViewManager.js
@@ -25,7 +25,7 @@
  * MainViewManager manages the arrangement of all open panes as well as provides the controller
  * logic behind all views in the MainView (e.g. ensuring that a file doesn't appear in 2 lists)
  *
- * Each pane contains one or more views wich are created by a view factory and inserted into a pane list.
+ * Each pane contains one or more views which are created by a view factory and inserted into a pane list.
  * There may be several panes managed by the MainViewManager with each pane containing a list of views.
  * The panes are always visible and the layout is determined by the MainViewManager and the user.
  *
@@ -139,6 +139,7 @@ define(function (require, exports, module) {
      * @private
      */
     var SECOND_PANE         = "second-pane";
+
 
     /*
      * NOTE: The following commands and constants will change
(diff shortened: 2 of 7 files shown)


src/config.json
```diff
@@ -20,7 +20,7 @@
         "extension_url": "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default": true,
         "build_timestamp": "",
-        "healthDataServerURL": "https://healthdev.brackets.io/healthDataLog"
```
zaggino: revert this change, it happens automatically after `npm install` but shouldn't be committed

zaggino:
Looks pretty good to me, I should be able to test tomorrow.

src/config.json
SteveMieskoski: Corrected the change, and that's good to know, thanks.

src/view/MainViewManager.js
```diff
@@ -846,6 +847,62 @@ define(function (require, exports, module) {
...
+            }
+
+        } else {
+
+            // check if currently viewed file is present in opposing pane.  If it is open the file.
+            // if it is not add it to the opposite pane and open.
+            activeFileInactiveView = inactivePane.getViewForPath(activeFile.fullPath);
+            inactiveFileActiveView = activePane.getViewForPath(inactiveFile.fullPath);
+
+            if (inactiveFileActiveView) {
```
petetnt: These could probably be a method like `_moveView` is

SteveMieskoski: @petetnt I looked into moving most of what is now in MainViewManager onto the pane prototype (see below), and I can push up the commit including it, however, with this setup I can't seem to get the active pane to consistently stay in place.  It flips from one pane to the other every third swap. I've tried promises and a few other ideas, but Maybe I'm missing something obvious.  Any insights are welcome.   
```javascript
    Pane.prototype.swapPaneContent = function(oppositeFile) {
        var self = this,
            oppositePane = self.id === FIRST_PANE ? SECOND_PANE : FIRST_PANE;

        if(!oppositeFile){
            var file = self.getCurrentlyViewedFile();
            MainViewManager._moveView( self.id, oppositePane.id, file , 0);
        } else {
            if (self.getViewForPath(oppositeFile.fullPath)) {
                CommandManager.execute(Commands.FILE_OPEN, {
                    fullPath: oppositeFile.fullPath,
                    paneId: self.id
                });
            } else {
                CommandManager.execute(Commands.CMD_ADD_TO_WORKINGSET_AND_OPEN, {
                    fullPath: oppositeFile.fullPath,
                    paneId: self.id
                });
            };
        };
    };

```
```javascript
    function swapPaneContent() {

        var activePaneId = getActivePaneId(),
            inactivePaneId = activePaneId === FIRST_PANE ? SECOND_PANE : FIRST_PANE,
            activePane = _getPane(activePaneId),
            inactivePane = _getPane(inactivePaneId),
            activeFile = activePane.getCurrentlyViewedFile(),
            inactiveFile = inactivePane.getCurrentlyViewedFile();

            inactivePane.swapPaneContent(activeFile);   // addition 
            activePane.swapPaneContent(inactiveFile);  // addition
    };
```

zaggino: I don't like this `rewrite` ... `Pane.prototype` calling `MainViewManager` methods doesn't seem right. Maybe @petetnt can specify little more what he meant, but looking at the code I prefer the original version.

zaggino:
I like the code, I like the feature, works quite well, but two things:
1) Why do I have to enable the feature in preferences?
2) This breaks https://github.com/adobe/brackets/pull/12853 in a way that if I use `ctrl+alt+f` to switch panes, `alt+w` stops to work until I mouse-click to the second pane, can you have a look why is that?

src/view/MainViewManager.js
petetnt: Yeah, my idea was something in line with the rewrite but just calling the method with `pane, file` arguments instead of adding it to the prototype. 

The original code is fine by me too if you think so too, you call @SteveMieskoski 👍

SteveMieskoski:
@zaggino having it as a default disabled preference was a mention in the original issue, and wouldn't take much to implement/remove so I went in that direction.  As such there isn't any necessary reason for it to be so.  I'll look into why alt+w stops working, and while I'm at it also change the keybindings to something+w because I agree it makes more sense.

PR #13151, opened 2017-03-06
add husky to prevent pushing failing code

zaggino:
this is just two lines, but it makes a significant change, consider it a proposal

basically, when you do `git push`, husky runs `prepush` script and if it fails, it won't allow you to push unless you override it with a command line parameter

advantages are clear, people won't push any code that doesn't fit our settings, while disadvantage is that it slows down the push (but how often you push anyway?)

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
         "fs-extra": "0.5.0",
         "glob": "7.0.6",
         "grunt": "0.4.5",
+        "husky": "0.13.2",
         "jasmine-node": "1.11.0",
         "grunt-jasmine-node": "0.1.0",
         "grunt-cli": "0.1.9",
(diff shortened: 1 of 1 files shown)


ficristo:
I like to prevent bad commits. I only wonder if this could be a contributor barrier.
I 'm fine either way so I defer this one to the others.

ingorichter:
I think everything that keeps the code base clean and tidy is worth a look and try. I don't think that the additional check doesn't introduce a barrier for contributors.

petetnt:
I just ran into `lint-staged` which can be used to lint (and `--fix`) files before committing which might be of use: https://github.com/okonet/lint-staged

swmitra:
I just love the idea of this check @zaggino 👍. I don't think it's going to slow down development as any sanity check failure has to be worked on immediately to pass the build. So either we work after a bad commit or while doing it. But this check can at least save a bad commit. I am all in for this 😄 .

swmitra (approved):

petetnt (approved):

PR #13153, opened 2017-03-07
add support for universal (language independent) code inspector

zaggino:
Hi guys, need this for https://github.com/zaggino/brackets-eslint/issues/54

Basically, eslint can be configured to lint any file extension through plugins (`.vue` in this case) and there're methods to check if a file path is lintable with the current eslint configuration, but to get there I need to be able to register the linter as language independent one which would be invoked for all potential files.

The change to Brackets is I believe quite trivial, also added a test-case.


petetnt (approved):

petetnt:
Didn't test this myself, but the change set LGTM.

PR #13201, opened 2017-03-21
Added a check of Untitled doc in _getNormalizedFilename and _getDenormalizedFilename

saurabh95:
Actually in the normalize path earlier there was no check of Untitled doc and it was appending the project dir to the Untitled path, so I added a check for Untitled doc before appending the project root.
Due to this JS codehints were not working for Untitled documents on changing the mode to JS.

All JavaScriptCodeHints tests are passing

### DIFF
--- a/src/extensions/default/JavaScriptCodeHints/ScopeManager.js
+++ b/src/extensions/default/JavaScriptCodeHints/ScopeManager.js
@@ -927,10 +927,11 @@ define(function (require, exports, module) {
          * @param {string} path - full path of file
          * @return {jQuery.Promise} - the promise for the request
          */
-        function primePump(path) {
+        function primePump(path, isUntitledDoc) {
             _postMessageByPass({
-                type        : MessageIds.TERN_PRIME_PUMP_MSG,
-                path        : path
+                type            : MessageIds.TERN_PRIME_PUMP_MSG,
+                path            : path,
+                isUntitledDoc   : isUntitledDoc
             });
 
             return addPendingRequest(path, OFFSET_ZERO, MessageIds.TERN_PRIME_PUMP_MSG);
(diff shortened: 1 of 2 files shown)


swmitra (approved):

ficristo:
Since this code will be refactored probably it is better to add a test.
Also you need to retarget the branch to release (if there aren't plans to merge again master to release of course).
Looking at the surrounding code of `tern-worker.js` usually parameters are passed to the functions instead of using a global variable like `isUntitledDoc` (see `handleUpdateFile` for example) but I don't think it is a big deal.
I added these two functions in https://github.com/adobe/brackets/pull/13147 in order to be able to remove `normalizeFilename` override. It's not pretty and if there are ways to improve the situation I would only be happy.

Overall this seems ok to me.

saurabh95:
@ficristo I will definitely add test after the release activities. Also will try to refactor the code as well

ficristo:
You can go ahead and merge for me.
Since I think you should retarget this PR to release branch I leave the merge to you.

swmitra:
@saurabh95 I am fine with the changes to unblock us for the release now, but would like to see some tests being added in general for InMemory document use cases (not just JS code hints). Also note that, eventually we have to refactor and migrate this code to Tern on node as well so that we can merge that PR on master after brackets 1.9 release. We may need more changes to leverage on new features/APIs added in Tern.

swmitra:
Merging.

petetnt (approved):
LGTM

swmitra:
@petetnt We both were typing together I guess 😄

swmitra:
@ficristo Let him do cherrypick after merging this to master, to keep master and release even for this module 👍

PR #13215, opened 2017-03-23
Update grunt-contrib-watch to 1.0.0 and clean a bit the task

ficristo:
The eslint task can be really slow, especially the eslint:src one, so to me the watch task (which run eslint) in its current form doesn't seem really useful.
Probably a better strategy would be to use watch only to run eslint on the changed file.
I was interested only in upgrading the dep so I didn't investigated more.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -268,21 +268,20 @@ module.exports = function (grunt) {
             ]
         },
         watch: {
-            all : {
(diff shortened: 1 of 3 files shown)


ficristo:
I also thought to simply remove this task since I don't believe it is useful in its current state.
But on the other hand I don't think this change hurts us in any way.
Merging.

PR #13216, opened 2017-03-23
Add inputStyle preference

ficristo:
Looking around it seems that setting `inputStyle` to `contenteditable` could improve the situation for screen readers but from what I understood is not stable enough.
After setting it I had some problems with the usage of the context menu so is not stable for Brackets.
I'm not sure what is the policy is in this case for adding new preferences,
but I think it could be useful to at least do some tests.


ficristo:
Rebase to add in the description that the new option is unstable.

zaggino (approved):

PR #13237, opened 2017-03-27
Now search history is stored in state and can be traversed using arrow keys while focus is on search text

saurabh95:
This implementation doesn't have any UI as per this PR. We can have a discussion regarding its UI. There is also a trello card regarding the same with UI also https://trello.com/c/IzACddEv/344-s-find-recent-search-history .
Will have UI implementation(as per the discussion) in next PR.


src/search/FindBar.js
```diff
@@ -310,9 +311,21 @@ define(function (require, exports, module) {
...
+                var maxCount = PreferencesManager.get("maxSearchHistory");
                 if (e.keyCode === KeyEvent.DOM_VK_RETURN) {
                     e.preventDefault();
                     e.stopPropagation();
+                    var searchQueryIndex = searchHistory.indexOf($('#find-what').val());
+                    if (searchQueryIndex !== -1) {
+                        searchHistory.splice(searchQueryIndex, 1);
+                    } else {
+                        if (searchHistory.length === maxCount) {
+                            searchHistory.splice(maxCount - 1, 1);
```
zaggino: isn't `.pop()` the same thing?

src/search/FindBar.js
```diff
@@ -333,6 +346,15 @@ define(function (require, exports, module) {
                         // if Shift is held down).
                         self.trigger("doFind", e.shiftKey);
                     }
+                    currentIndex = 0;
+                } else if (e.keyCode === KeyEvent.DOM_VK_DOWN) {
+                    currentIndex = (currentIndex - 1 + Math.min(maxCount, searchHistory.length)) % Math.min(maxCount, searchHistory.length);
```
zaggino: line too long (140 chars)

src/search/FindBar.js
```diff
@@ -267,6 +267,7 @@ define(function (require, exports, module) {
         FindBar._addFindBar(this);
 
         var $root = this._modalBar.getRoot();
+        var currentIndex = 0;
```
zaggino: I'd rename this to `historyIndex` or something little more specific

src/search/FindBar.js
```diff
@@ -333,6 +346,15 @@ define(function (require, exports, module) {
                         // if Shift is held down).
                         self.trigger("doFind", e.shiftKey);
                     }
+                    currentIndex = 0;
+                } else if (e.keyCode === KeyEvent.DOM_VK_DOWN) {
+                    currentIndex = (currentIndex - 1 + Math.min(maxCount, searchHistory.length)) % Math.min(maxCount, searchHistory.length);
+                    $("#find-what").val(searchHistory[currentIndex]);
+                    self.trigger("queryChange");
+                } else if (e.keyCode === KeyEvent.DOM_VK_UP) {
+                    currentIndex = (currentIndex + 1 + Math.min(maxCount, searchHistory.length)) % Math.min(maxCount, searchHistory.length);
```
zaggino: can you wrap `if (e.keyCode === KeyEvent.DOM_VK_DOWN)` and `if (e.keyCode === KeyEvent.DOM_VK_UP) {` codes into a function taking an argument up or down? code seems to repeat itself

zaggino (approved):

swmitra (approved):
Implementation looks good to me. We might want to add some UI at a later point of time.

swmitra:
@saurabh95 Great job! This is a very handy feature for developers. Keep these coming... 👍

Issue #12892, opened 2016-11-10
Coding in JSX files — javascript not fully supported in *.jsx


mrbodich:
### Prerequisites

* [x] Can you reproduce the problem with `Debug -> Reload Without Extensions`?
* [x] Did you perform a cursory search to see if your bug or enhancement is already reported?
* [x] Did you read the [Troubleshooting guide](https://github.com/adobe/brackets/wiki/Troubleshooting)?

For more information on how to write a good bug report read [here](https://github.com/adobe/brackets/wiki/How-to-Report-an-Issue)
For more information on how to contribute read [here](https://github.com/adobe/brackets/blob/master/CONTRIBUTING.md)

### Description

Hi. JSX is basically the same as JS but with some additional Adobe functions support. Please make Brackets helping when coding in *.jsx files just the same as in *.js files. Copy the behavior please. Thanks.

### Steps to Reproduce

1. Copy the behavior from *.js files to *.jsx files.
2. [Second Step]
3. [and so on...]

**Expected behavior:** Brackets should behave in JSX files just the same like in JS files.

**Actual behavior:** There is no suggestions of functions or variables etc. in JSX files like in JS.

### Versions

Brackets 1.7.0
MacOS 10.12.1

mussakhojayeva:
Can I work on it?

mrbodich:
What do you mean, mussakhojayeva?

mussakhojayeva:
I am talking about this: petetnt added Starter bug PICK ME UP labels 30 minutes ago

zaggino:
Hi, feel free to open a pull request. No need to ask for permission.

zaggino:
Fixed in https://github.com/adobe/brackets/pull/13243

PR #13243, opened 2017-03-28
Add JavaScript code hinting in JSX files

justinrusso:
#12892 Add support for JavaScript code hinting in JSX files.


justinrusso:
I've accepted the agreement since this test was run. I'm new to dealing with CI's - do I need to make it run again, and if yes, how?

zaggino:
Thanks @justinrusso !

Issue #13209, opened 2017-03-22
Core: There is a layout issue on extension manager panel when narrow down the window


Zhanxin:
Steps:
1. Open the Brackets.
2. Click the icon about Extension Manager.
3. Narrow down the window.
4. Observe the sort toggle and the search bar.

Result:
The sort toggle and the search bar are misplaced.

Expected:
The sort toggle and the search bar should be displayed in the correct position.
![layout](https://cloud.githubusercontent.com/assets/2245179/24198353/89b8c22c-0f40-11e7-9073-dab355a250b3.png)

Zhanxin:
Could reproduce on release 1.8 build.

petetnt:
Also reported by me at https://github.com/adobe/brackets/pull/13080#issuecomment-278798445

lkcampbell:
I will be glad to fix this bug.

@Zhanxin, would you mind providing a fake screenshot graphic depicting what the phrase "displayed in the correct position" means?

petetnt:
@lkcampbell there's a PR fixing this alteady if you want to review it https://github.com/adobe/brackets/pull/13245 :+1:

lkcampbell:
@petetnt okay thanks I will look at it this weekend.

zaggino:
Fixed in https://github.com/adobe/brackets/pull/13245

PR #13245, opened 2017-03-29
Fix layout on narrow extension manager

justinrusso:
I've added queries to stop the search and filter drop-down from overlapping the tabs when the window becomes too small, and from there, the search bar element will adjust in size

Ref issue: #13209

### DIFF
--- a/src/htmlContent/extension-manager-dialog.html
+++ b/src/htmlContent/extension-manager-dialog.html
@@ -16,8 +16,10 @@
                    <option value="downloadCount">{{Strings.EXTENSIONS_DOWNLOADS}}</option>
                </select>
             </div>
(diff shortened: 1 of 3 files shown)


saurabh95:
Just tested it out, works for me. The implementation also looks good. 👍

saurabh95 (approved):

lkcampbell:
@saurabh95 and @petetnt agreed. Works well, not a huge change. Has a nice little conditional that puts the right-sided elements under the tabs so everything is still usable regardless of the width.

lkcampbell (approved):

zaggino:
👍 merged, thanks @justinrusso

justinrusso:
No problem, glad you liked it!

PR #13250, opened 2017-03-30
Add support for TypeScript

justinrusso:
Added support for TypeScript files, and added code hinting using JavaScript's code hinting extension.


lkcampbell:
@zaggino didn't you already add Typescript support in an extension? What do you think?

lkcampbell:
@justinrusso I'm not saying it's a bad idea, was just wondering what @zaggino thought.

Looks like the extension is working for you though, so i will add a label.

zaggino:
Yes, this is available through my extension, along with code hints and linting.

Issue #13169, opened 2017-03-11
Block Comments don't take indentation into account


recursionbane:
### Prerequisites

* [Y] Can you reproduce the problem with `Debug -> Reload Without Extensions`?
* [Y] Did you perform a cursory search to see if your bug or enhancement is already reported?
* [Y] Did you read the [Troubleshooting guide](https://github.com/adobe/brackets/wiki/Troubleshooting)?

### Description

The dynamic indentation of Brackets is great, especially for cleaning up HTML.
When one or more indented lines are selected, and Edit->"Toggle Line Comment" is clicked, the indentation is also commented.

### Steps to Reproduce

1. Select one or more lines of indented code
2. Click Edit -> Toggle Line Comment

**Input:**
```HTML
                    <md-content class="md-padding ">
```

**Expected behavior:** 
```HTML
                    <!-- <md-content class="md-padding "> -->
```

**Actual behavior:**
```HTML
<!--                    <md-content class="md-padding ">-->
```

### Versions

Please include the OS and what version of the OS you're running: Windows 7 SP1
Please include the version of Brackets.: Release 1.8 build 1.8.0-17108 (alf_localization_release_1.8 3af64fae4); build timestamp: Wed Nov 02 2016 12:07:29 GMT+0000

blairdow:
I would love a fix for this as well!

redmunds:
This is a duplicate of #8012 which has been fixed. It looks like the fix is to add the `indentLineComment` preference and set it to `true`.

ficristo:
Reopening: `indentLineComment` does work only for languages which defines single line comments.
Html language defines only block comments.

PR #13254, opened 2017-04-01
Add new comments preference. Support indent block comments on line co…

ficristo:
…mment command

Fixes https://github.com/adobe/brackets/issues/13169

I created a new `comments` preference with an indent option with the idea of deprecate the `indentLineComment` preference. Is there a way to deprecate a preference?
I used a new object with the idea to add also a `padding` option in the future (like the CodeMirror addon), but maybe is better to simply create a new `indentBlockComment` preference without deprecate anything.
Any thought?

### DIFF
--- a/src/editor/Editor.js
+++ b/src/editor/Editor.js
@@ -2713,7 +2713,7 @@ define(function (require, exports, module) {
     };
 
     /**
-     * Returns true if blockCommentIndent is enabled for the specified or current file
+     * Sets indentBlockComment option.
      * Affects any editors that share the same preference location.
      * @param {boolean} value
      * @param {string=} fullPath Path to file to get preference for
--- a/src/editor/Editor.js
+++ b/src/editor/Editor.js
@@ -81,25 +81,24 @@ define(function (require, exports, module) {
 
     /** Editor preferences */
 
-    var CLOSE_BRACKETS       = "closeBrackets",
-        CLOSE_TAGS           = "closeTags",
-        DRAG_DROP            = "dragDropText",
-        HIGHLIGHT_MATCHES    = "highlightMatches",
-        LINEWISE_COPY_CUT    = "lineWiseCopyCut",
-        SCROLL_PAST_END      = "scrollPastEnd",
-        SHOW_CURSOR_SELECT   = "showCursorWhenSelecting",
-        SHOW_LINE_NUMBERS    = "showLineNumbers",
-        SMART_INDENT         = "smartIndent",
-        SOFT_TABS            = "softTabs",
-        SPACE_UNITS          = "spaceUnits",
-        STYLE_ACTIVE_LINE    = "styleActiveLine",
-        TAB_SIZE             = "tabSize",
-        UPPERCASE_COLORS     = "uppercaseColors",
-        USE_TAB_CHAR         = "useTabChar",
-        WORD_WRAP            = "wordWrap",
-        INDENT_LINE_COMMENT  = "indentLineComment",
-        INDENT_BLOCK_COMMENT = "indentBlockComment",
-        INPUT_STYLE          = "inputStyle";
+    var CLOSE_BRACKETS      = "closeBrackets",
+        CLOSE_TAGS          = "closeTags",
+        DRAG_DROP           = "dragDropText",
+        HIGHLIGHT_MATCHES   = "highlightMatches",
+        LINEWISE_COPY_CUT   = "lineWiseCopyCut",
+        SCROLL_PAST_END     = "scrollPastEnd",
+        SHOW_CURSOR_SELECT  = "showCursorWhenSelecting",
+        SHOW_LINE_NUMBERS   = "showLineNumbers",
+        SMART_INDENT        = "smartIndent",
+        SOFT_TABS           = "softTabs",
+        SPACE_UNITS         = "spaceUnits",
+        STYLE_ACTIVE_LINE   = "styleActiveLine",
+        TAB_SIZE            = "tabSize",
+        UPPERCASE_COLORS    = "uppercaseColors",
+        USE_TAB_CHAR        = "useTabChar",
+        WORD_WRAP           = "wordWrap",
+        INDENT_LINE_COMMENT = "indentLineComment",
+        INPUT_STYLE         = "inputStyle";
 
 
     /**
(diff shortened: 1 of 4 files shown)


redmunds:
@ficristo I think it's more trouble that it's worth to deprecate the `indentLineComment` preference. I vote for simply adding a new `indentBlockComment` preference.

redmunds:
@ficristo Also, I think #13169 can be fixed without adding a new preference as follows:

For languages that do not have a different line comment format (such as HTML), the block comment format is used on a single line. In that case, the `indentLineComment` preference could be checked.

But I can also see the value of having a separate indenting preference for block comments.

ficristo:
I would find `indentLineComment` a bit misleading if used also for block comments.
(In fact I wanted a single `indent` option to be used for line and block comments)
While I think the new object preference would be cleaner, I'll keep it simple and go with adding a new pref.

ficristo:
I still need to test it a bit more

ficristo:
It should be ready.
@redmunds do you have time to review?

redmunds:
Seeing weird results for block comment, indentBlockComment=true, and cursor at start of line (i.e. line not selected). Before:

```html
  <p>lkj wlekjr wlkerj lwkejr wlke lqkwje lkqjwe lkqwje lwkqje lq</p>
```

After:

```html
--<!-->  <p>lkj wlekjr wlkerj lwkejr wlke lqkwje lkqjwe lkqwje lwkqje lq</p>
```

src/editor/EditorCommandHandlers.js
```diff
@@ -467,12 +485,29 @@ define(function (require, exports, module) {
...
             if (canComment) {
                 var completeLineSel = sel.start.ch === 0 && sel.end.ch === 0 && sel.start.line < sel.end.line;
+                var startCh = _firstNotWs(doc, sel.start.line);
                 if (completeLineSel) {
-                    editGroup.push({text: suffix + "\n", start: sel.end});
-                    editGroup.push({text: prefix + "\n", start: sel.start});
+                    if (indentBlockComment) {
+                        var endCh = _firstNotWs(doc, sel.end.line - 1);
+                        editGroup.push({
+                            text: _.repeat(" ", endCh) + suffix + "\n",
```
redmunds: Can't always insert spaces -- need to use indent preference here in case it's Tabs.

src/editor/Editor.js
```diff
@@ -2709,6 +2713,27 @@ define(function (require, exports, module) {
     };
 
     /**
+     * Returns true if blockCommentIndent is enabled for the specified or current file
```
redmunds: First line of this comment is true for "get", but not "set".

redmunds:
@ficristo Done with review. Found a couple cases that don't work as expected, but mostly working as advertised. Tests pass for me on Mac 10.12.

redmunds:
Also, all Preferences should be listed in wiki here:
https://github.com/adobe/brackets/wiki/How-to-Use-Brackets#preferences

Please add `indentBlockComment` and `indentLineComment` (which was never added). Thanks!

src/editor/EditorCommandHandlers.js
ficristo: Great catch! Fixed.

ficristo:
I cannot reproduce the issue above about the html. Is that the full snippet?

redmunds:
@ficristo To repro [this bug](https://github.com/adobe/brackets/pull/13254#issuecomment-293049669), make sure cursor is at start of line, or in middle of indent. Doesn't seem to happen if cursor is immediately before start tag.

I also notice weird results when cursor is in middle of tag contents. Comment is added from start of tag to cursor positioin, which causes invalid HTML:

```html
  <!--<p>lkj wlekjr wlkerj lwkejr wlke--> lqkwje lkqjwe lkqwje lwkqje lq</p>
```

redmunds:
@ficristo Here's a simple page where I'm also seeing problem on `<p>` tag:

```html
<!doctype html>
<html>
<head>
<title></title>
</head>
<body>
  <p>one two three</p>
</body>
</html>
```

ficristo:
Odd, I still cannot reproduce. I tryed opening a new folder with only a `test.html` file with the snippet you posted above.
Some tests I added put the cursor at the beginning of the line. Do all the tests pass for you?

ficristo:
Initially I thought that `Edit \ Toggle Block Comment` was equal to `Edit \ Toggle Line Comment` but added block comments. But instead it add a block comment starting and ending at the exact position of the cursors.
I don't want to change that.
I want to support indentation for line comments also for languages with only block delimiters (like html).
I also removed the new preference and reused indentLineComment, I think the new preference would be misleading.
@redmunds could you take a look?

redmunds:
@ficristo

> But instead it add a block comment starting and ending at the exact position of the cursors.

Yes, that's the case when either start or end cursor is in middle of line. I agree that behavior should not change.

But the other case is when the selection only contains full lines (both start and end cursors are at start of line). In that case, a new line is added above selection with opening block delimiter, and another line is added after selection with closing block delimiter. This is the case where I was expecting the new `indentBlockComment` to take effect and delimiters would be indented to match code. Indenting for first line of selection might not match indenting of last line -- not sure how to handle that case.

Does that make sense? Let me know if you want to take that on or not -- I'll review code either way.

ficristo:
Oh, I think we had a bit different things in mind.
(Now that I know what it does...) I prefer to leave `Edit \ Toggle Block Comment` as is.
While interesting, it would seem odd to match the indentatation when the cursors are exactly at the beginning but not if they aren't, or so it appears to me at the moment.
If possible I prefer to not implement this part of the feature.

src/editor/EditorCommandHandlers.js
```diff
@@ -297,6 +297,22 @@ define(function (require, exports, module) {
...
+     *
+     * @private
+     * @param {!Document} doc
+     * @param {number} lineNum
+     * @returns {number} the column index or null
+     */
+    function _firstNotWs(doc, lineNum) {
+        var text = doc.getLine(lineNum);
+        if (text === null || text === undefined) {
+            return null;
```
redmunds: The return value from this function is not validated -- it gets used directly as a char position. Seems like it would be more consistent with CodeMirror APIs to return `-1` here (but my CM API knowledge is rusty...).

redmunds:
@ficristo Done with review. Looks good except for maybe 1 change.

ficristo:
I seems to remember I hit a case where the API could return null, but I don't remeber which...
And looking around the codebase I didn't find any check at all.
I changed it to return 0, I hope its fine for now.

ficristo:
@redmunds could you take a final look (and eventually merge)?

redmunds:
LGTM. Merging.

ficristo:
Thank you.

PR #13256, opened 2017-04-02
Fix #13099: Disallow user to create files/folder using relative path

### LINKED ISSUE #13099: User is able to create file/folders using relative path
### Prerequisites

* [x] Can you reproduce the problem with `Debug -> Reload Without Extensions`?
* [x] Did you perform a cursory search to see if your bug or enhancement is already reported?
* [x] Did you read the [Troubleshooting guide](https://github.com/adobe/brackets/wiki/Troubleshooting)?

### Description

This bug was first found by @pomax(mozilla#434) and then fixed by @tallandroid (#11862).

"It's possible to right-click in the file tree and create a file (or folder) name it `../something`, and the path isn't resolved, but rather the .. and / characters are treated as literals to be used in the filename itself." - @humphd(#11609)

It looks like this bug wasn't fully fixed. I'm using the latest version of Brackets and I'm still able to reproduce it.

screen shot 2017-02-12 at 12 30 43 pm
screen shot 2017-02-12 at 12 31 40 pm

### Steps to Reproduce

1. Create/Rename a file
2. Give it a valid relative path with a unique file name (Eg. `../../file1`)
3. Save it
4. File is now created in whatever path you gave it

**Expected behavior:** An error message should show complaining about the relative path.

**Actual behavior:** File is renamed/created without an error message.

### Versions
OS: macOS sierra 11.2.3
Brackets Version: 1.8 build 1.8.0-17108

sdalmeida:
This is a Work In Progress PR referring to original issue adobe#13099
Before this PR can be landed, it would be nice to know which chars are actually invalid.

These are the chars that I'm invalidating:

- `\?`
- `\\`
- `\*`
- `\.{2,}` <-- .. or more should not be allowed, but allow .
- `\.$` <-- allow a file to have . but not end with .
- `\/`
- `\|`
- `\>`
- `\<`

I also have to add a unit test for this case.
Let me know if there are any chars that need to be added/removed.


swmitra:
@simon66 I still have the same confusion regarding invalid character handling code. Part of the reason is the way file names get validated by the OS, it's quite different in case of windows and osx. Not sure whether we can have a common regex validator as that would limit the file naming capability in either of the platform.

swmitra:
What we can think about is, predominantly the projects being worked upon are web projects in Brackets. Eventually these files will be served and loaded in browser shells. What kind of restrictions do we have in the file names whole serving them over network. Just thinking loud, may be not valid in this context.

simon66:
Another confusion of mine is why are these filenames considered "bad"

screen shot 2017-04-02 at 1 23 23 pm

Why is `com` a valid filename and `com1` not?
Also, I don't see why any of these file names are labelled as invalid.
Any ideas @tallandroid?

simon66:
@swmitra I think that no matter where this code lives (MacOS, Windows, Linux), those chars should be invalid. I'm not too sure why we should allow `<` or `>` on a linux env but not on Windows (Even though that char should be invalid on both OS.
But then again, which chars are invalid and which should not be.

simon66:
I've changed the function `isValidFilename` to disallow any file name that contains folder structure (like `tmp\file.log`) and I created a new function to check if full pathname is valid. I've also added some test cases.

src/project/ProjectModel.js
```diff
@@ -70,41 +70,48 @@ define(function (require, exports, module) {
...
-        // Checks for valid Windows filenames:
-        // See http://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx
+    function isValidFilename(filename) {
+        // Fix issue adobe#13099
+        // See https://github.com/adobe/brackets/issues/13099
         return !(
-            new RegExp("[" + invalidChars + "]+").test(filename) ||
-            _illegalFilenamesRegEx.test(filename)
+            filename.match(_invalidChars)|| filename.match(_illegalFilenamesRegEx)
+            //filename.match(_invalidChars) || filename.match(_illegalFilenamesRegEx)
```
humphd: This got left in by accident, please remove.

simon66:
@humphd Done 👍

swmitra (approved):

simon66:
Hi @swmitra 
I want to follow up on the status of this PR. I see that the code was approved but the PR is still open.
Thanks :)

nethip:
Thanks for your contributions @simon66 .

PR #13268, opened 2017-04-05
Scrapped QuickDocs data for css and html from MDN to css.json and html.json

saurabh95:
The values are webscrapped from MDN docs and summary, URLs and logo are derived from https://github.com/mozilla/brackets/blob/master/src/extensions/extra/MDNDocs/css.json

CSS Example

image

HTML Example

image


swmitra (approved):
LGTM 👍

ficristo:
Can instead go with merging https://github.com/adobe/brackets/pull/10036 and then look if https://github.com/mozilla/brackets/blob/master/src/extensions/extra/MDNDocs is good for our case?
And if so, replace the references of WebPlatformDocs to MDN?
It feels odd to me to see MDN things in an extension called WebPlatformDocs

saurabh95:
Actually I have merged https://github.com/mozilla/brackets/blob/master/src/extensions/extra/MDNDocs  with brackets in my local (HTML docs are also working in it), I haven't done a blind merge as in the mozilla's extension they only show summary for css properties also and we show possible values of a given css property along with small description. I was thinking of creating another PR for it, I can have those changes in this PR itself with folder renaming also.

src/extensions/default/MDNDocs/main.js
```diff
@@ -0,0 +1,160 @@
...
+     * rejected if the file is missing/corrupt.
+     * @param {string} fileName JSON file to load
+     * @return {!$.Promise}
+     */
+    function getDocs(fileName) {
+        if (!promiseCache[fileName]) {
+            var result = new $.Deferred();
+
+            // XXXBramble - load from server vs. fs
+            $.ajax({
```
ficristo: This change seems very specific of mozilla/adobe, should we instead read from a file like before?

src/extensions/default/MDNDocs/README.md
```diff
@@ -1,5 +1,5 @@
 # Updating the Docs
 Use the Node script [update-wpd-docs](https://github.com/MarcelGerber/update-wpd-docs) to update the `css.json` contents:
```
ficristo: If we are using MDN we should also use https://github.com/mozilla/brackets/tree/master/src/extensions/extra/MDNDocs/tools/scrape-mdn instead of @MarcelGerber tool.

src/extensions/default/MDNDocs/InlineDocsViewer.html
```diff
@@ -1,4 +1,4 @@
-<div class="css-prop-defn" tabIndex="0"> <!-- tabIndex needed: otherwise click focuses CodeMirror scroller and Esc won't work -->
+<div class="css-prop-defn  {{#fullscreenSummary}}fullscreen-summary{{/fullscreenSummary}}" tabIndex="0"> <!-- tabIndex needed: otherwise click focuses CodeMirror scroller and Esc won't work -->
```
petetnt: Nit: extra space

src/extensions/default/MDNDocs/main.js
```diff
@@ -56,34 +61,39 @@ define(function (require, exports, module) {
...
      * @return {?$.Promise} resolved with an InlineWidget; null if we're not going to provide anything
      */
     function inlineProvider(hostEditor, pos) {
-        var langId = hostEditor.getLanguageForSelection().getId();
-        // Only provide docs when cursor is in CSS content
-        if (langId !== "css" && langId !== "scss" && langId !== "less") {
+        var jsonFile, propInfo,
+            propQueue = [], // priority queue of propNames to try
+            langId = hostEditor.getLanguageForSelection().getId(),
+            supportedLangs = ["css", "scss", "less", "html"],
```
petetnt: Maybe this could be an `Object` or similar so we wouldn't have to rely on the index.

petetnt (commented):
Didn't test it out, but for a quick look the change set looks good to me with couple of nits

src/extensions/default/MDNDocs/README.md
```diff
@@ -0,0 +1,3 @@
+# Updating the Docs
+* Use the Node script [scrape-mdn](https://github.com/mozilla/brackets/tree/master/src/extensions/extra/MDNDocs/tools/scrape-mdn) to update the `css.json` and `html.json` contents:
+* Both the JSON files will have URL and Summary, but for css.json we need to have possible values of each css property along with small description, so use this Node script [MDNDocsScrapper](https://github.com/saurabh95/MDNDocsScrapper), this takes css.json as input and then creates a newcss.json with the desired content.
```
ficristo: I think things like MDNDocsScrapper should belong in https://github.com/brackets-userland

src/extensions/default/MDNDocs/README.md
```diff
@@ -0,0 +1,3 @@
+# Updating the Docs
+* Use the Node script [scrape-mdn](https://github.com/mozilla/brackets/tree/master/src/extensions/extra/MDNDocs/tools/scrape-mdn) to update the `css.json` and `html.json` contents:
```
ficristo: Probably we need to change grunt to avoid to have MDNDocs/tools in the distribution.

src/extensions/default/MDNDocs/README.md
saurabh95: Ok I will add it in brackets-userland

ficristo (approved):

src/extensions/default/MDNDocs/README.md
saurabh95: MDN/tools is not generated in dist in grunt task

ficristo: That's great 😄

Issue #13264, opened 2017-04-04
Setting saveFoldStates preference to false breaks code folding


lkcampbell:
**OS:** MacS 10.12.3 Sierra

**Brackets Version:** Release 1.9 build 1.9.0-17312 (release 189f6d39a) 

**Repro  Steps:**
1. Add the line `"code-folding.saveFoldStates": false` to the brackets.json preference file.
1. Reload Brackets

**Observed Results:**
Code folding stops working. Error messages in the Dev Console:
```
/utils/EventDispatcher.js:231 Exception in 'activeEditorChange' listener on Object {_eventHandlers: Object, _EventDispatcher: true} TypeError: Cannot convert undefined or null to object TypeError: Cannot convert undefined or null to object
    at _isCurrentlyFolded (file:///Applications/Brackets.app/Contents/www/extensions/default/CodeFolding/foldhelpers/foldgutter.js:68:31)
    at updateFoldInfo (file:///Applications/Brackets.app/Contents/www/extensions/default/CodeFolding/foldhelpers/foldgutter.js:92:22)
    at file:///Applications/Brackets.app/Contents/www/extensions/default/CodeFolding/foldhelpers/foldgutter.js:137:13
    at runInOp (/thirdparty/CodeMirror/lib/codemirror.js:3837:26)
    at CodeMirror.operation (/thirdparty/CodeMirror/lib/codemirror.js:8131:35)
    at updateInViewport (file:///Applications/Brackets.app/Contents/www/extensions/default/CodeFolding/foldhelpers/foldgutter.js:136:12)
    at CodeMirror.<anonymous> (file:///Applications/Brackets.app/Contents/www/extensions/default/CodeFolding/foldhelpers/foldgutter.js:381:17)
    at eval (/thirdparty/CodeMirror/lib/codemirror.js:3847:20)
    at CodeMirror.setOption (/thirdparty/CodeMirror/lib/codemirror.js:7748:50)
[...]

PR #13269, opened 2017-04-05
Addresses #13264 where setting `saveFoldStates` preference to false caused folding problems

thehogfather:
This PR properly initialises the `_lineFolds` cache to the empty object when the saveFoldStates preferences is set to false. Previously, this was not being done and it caused issues down stream, i.e., when rendering gutters as the active editor changed.

Also updated the test cases to accommodate the requirement.

cc @lkcampbell @ficristo


lkcampbell:
@thehogfather, the changes look good to me. Initialize the uninitialized property with an empty object and get rid of the undefined/null error message down the road. 

Is `_codeMirror._lineFolds` where you are storing all of the fold state information for each instance of CodeMirror opened in brackets?

The only problem I see is the Travis build is failing. @ficristo or @petetnt any ideas what might be causing the build fail?

thehogfather:
@lkcampbell 
>Is _codeMirror._lineFolds where you are storing all of the fold state information for each instance of CodeMirror opened in brackets?

Yes that is correct.

With respect to Travis failing, I'll pull upstream and push to trigger the build again. The error messages show it did not get very far into the build process.

thehogfather:
That did no seem to help. Let me know if there is anything I can do.

petetnt:
Restarted the CI by hand, I'll look into it if it still fails

swmitra:
I was just trying out with the OAuth token and couldn't exceed the limit in 60 attempts. Not sure why this one failed!

ficristo:
Fixes #13264 (adding this comment so GitHub creates a link to the issue)

ficristo (approved):

lkcampbell:
Okay, I check that the fix works and ran the unit tests: no errors.

Merging.

PR #13271, opened 2017-04-06
Now search history is stored at project level context instead of global search history

saurabh95:


ficristo:
I tryed the search on Sublime and the history is the same for all the projects I opened.
Was done some kind of UX research for this? I'm asking because I'm not sure if this is the right move.

saurabh95:
Even Visual Studio also has global summary, but I thought this would be make searchHistory more contextual in terms of project. Anyway we can have some input from others as well, then we can decide whether to go with this or not.
@zaggino @petetnt @redmunds @swmitra ^?

redmunds:
@saurabh95 This stores history in project-level prefs file (i.e. `.brackets.json` in project root), correct? If so, then if that file does not already exist, then it will be created. Some people might not like files being created in their project folder. Maybe there should be a preference controlling history scope.

saurabh95:
@redmunds The history is stored in state.json with path of project as key in APPDATA, so it won't interfere with project root directory.
Something like this

`"project": {
        "C:/Program Files (x86)/Brackets/dev/samples/root/Getting Started/": {
            "code-folding.folds": {
                "C:/Users/kathpali/AppData/Roaming/Brackets/state.json": {}
            },
            "openFiles": [
                {
                    "file": "C:/Program Files (x86)/Brackets/dev/samples/root/Getting Started/index.html",
                    "paneId": "first-pane",
                    "cursor": {
                        "line": 0,
                        "ch": 0,
                        "sticky": null
                    }
                }
            ],
            "searchHistory": [
                "searchH21",
                "searchH",
                "state"
            ]
        }
    }`

saurabh95:
Closing this, since global history makes more sense rather than having history at project context

petetnt:
Let's see if this gets requested, would make sense as a pref

Issue #13240, opened 2017-03-28
Unable to change font family AND font size


brunnopleffken:
### Prerequisites

* [x] Can you reproduce the problem with `Debug -> Reload Without Extensions`?
* [x] Did you perform a cursory search to see if your bug or enhancement is already reported?
* [x] Did you read the [Troubleshooting guide](https://github.com/adobe/brackets/wiki/Troubleshooting)?

### Description

Go to View > Themes.
Change the font size, the font family is automatically changed to Source Code Pro.
Now go to View > Themes and change the font family, the font size is automatically reseted to the default (12 or 13 I guess), ignoring the one I saved before.

It's just one or another...

### Versions

Release 1.9 build 1.9.0-17312 (release 189f6d39a) 
build timestamp: Tue Mar 21 2017 03:04:51 GMT-0700

macOS Sierra 10.12.3

guillecro:
Watch this issue
https://github.com/adobe/brackets/issues/13221
It has been reported

swmitra:
Closing this as fixed by #13279 and available as part of 1.10 pre-release 1 build https://github.com/adobe/brackets/releases/tag/release-1.10-prerelease-1

Issue #13234, opened 2017-03-27
ligatures not working


MandarSant:
### Description

[ligatures of fira code not working in newest version of brackets 1.9 refer to the following [issue](https://github.com/tonsky/FiraCode/issues/393#issuecomment-289385186) ] 

**Actual behavior:**

 [When I change font line, it applies the changes and shows everything(ligatures ) in Fira Code. But when I restart Brackets it starts in default font]

### Versions

windows 8.1 
Brackets 1.9

ficristo:
Closing as a dupe of #13221. (I hope to have understood the issue correctly)

Issue #13232, opened 2017-03-27
Brackets 1.9, can't configure my theme font


AgamlaRage:
Brackets 1.9, Windows 10

I have to setup my font-family again, each time I quit Brackets.

Here the related part of my **AppData/Roaming/Brackets/brackets.json**, that is valid json.

>     "fonts.fontSize": "15px",
>     "fonts.fontFamily": "Liberation Sans",

My theme configuration is the same as my configuration in Brackets 1.8.
i did the normal update, running **Brackets.Release.1.9.msi**
It should works ...

clementi:
This appears to be the same as #13221

AgamlaRage:
Oh yes, sorry for duplicate :(

ficristo:
Closing as a dupe of https://github.com/adobe/brackets/issues/13221

Issue #13221, opened 2017-03-24
Custom font is ignored


thany:
### Prerequisites

* [x] Can you reproduce the problem with `Debug -> Reload Without Extensions`?
* [x] Did you perform a cursory search to see if your bug or enhancement is already reported?
* [x] Did you read the [Troubleshooting guide](https://github.com/adobe/brackets/wiki/Troubleshooting)?

### Description

A custom font set in View->Themes is ignored.

### Steps to Reproduce

1. Open Brackets
2. View->Themes
3. Set any other font. In my case, Fira Code, so: `Fira Code, FiraCode, 'SourceCodePro-Medium', ＭＳ ゴシック, 'MS Gothic', monospace`

**Expected behavior:**
Custom font should be displayed in the editor.

**Actual behavior:**
Custom font is set, but only once, after committing the setting. After zooming in/out, after a restart, or a reload, the default font is displayed again.

### Versions

Brackets 1.9
Windows 10

This is a regression from 1.8, where setting a custom font worked just fine.

kobezhu:
‘I’ve been through the same situation!

prasadmudedla:
same situation... tried changing the defaultPreferneces.json.... Still no luck...

clementi:
I'm seeing this also.

guillecro:
This issue is already posted here,
https://github.com/adobe/brackets/issues/13205

Its not working even in Ubuntu (in my case Ubuntu 16.04 LTS x64)

thany:
That appears to be a different issue, as it also reports the problem in 1.8.
This particular case is a *regression* from 1.8, meaning the feature worked fine in 1.8, but it broke as of 1.9.

gxnie:
I have modified the brackets.min.css.

thany:
Can this please be fixed with a maintenance update?

It happens too often that we need to wait 3 months or longer, for an issue to get solved in the next release. And any new release may introduce other breaking bugs, causing us to have to wait *even* longer.

guillecro:
> Can this please be fixed with a maintenance update?

I +1 this comment. Its such annoying bug..

jpdupont:
+1

swmitra:
@thany We are working on this issue and hopefully have a fix by next week. Once the fix is in, we will trigger a pre-release which would be available under Brackets GitHub releases.

vahid-sanati:
I have this bug too.

swmitra:
[...]

PR #13279, opened 2017-04-07
Fix font family name preference

swmitra:
This PR fixes the custom font-family issues reported in #13205, #13221, #13232, #13234 and #13240.   
It was introduced as part of #13091 commit 013c8c5624c1622d2ce0efc34fe23cae044ed86e.

Ping @ficristo @petetnt for a quick review.


petetnt (approved):
LGTM

PR #13288, opened 2017-04-09
Fix #13274 - make Brackets margin/padding remoteHighlight work like Chrome one

### LINKED ISSUE #13274: [Live preview] Improve the appearance of transformed elements when showing margins/paddings
### Steps to Reproduce

1. Enable highlighting elements in Live Preview
2. Reenable transform to custom-highlightning divs in `RemoteFunctions.js` (it is not applied atm)
2. Transform it via CSS, ex. `rotate:90deg`.
3. As you can see, the margin/paddings showed differ from those in Chrome DevTools

**Expected behavior:** Elements are highlighted the same way as DevTools does

**Actual behavior:** Their size differs due to `window.getClientBoundingRect` (it includes transform and causes the bugs in calculations) 

See https://github.com/adobe/brackets/pull/12949#issuecomment-268806531 for more details.

Worie:
This resolves the issue with highlighting a CSS transformed element.

Probably could be a little bit more efficient (do not calculate the real position via `do {} while loop` when the element isn't transformed by CSS) etc. But it does the job. From what I tested it acts just like the Chrome devtools inspector.

@petetnt @saurabh95

### DIFF
--- a/src/LiveDevelopment/Agents/RemoteFunctions.js
+++ b/src/LiveDevelopment/Agents/RemoteFunctions.js
@@ -425,11 +425,11 @@ function RemoteFunctions(config, remoteWSPort) {
             offsetTop  = 0;
             
             // Probably the easiest way to get elements position without including transform
-            do{
+            do {
                 offsetLeft += el.offsetLeft;
                 offsetTop  += el.offsetTop;
                 el = el.offsetParent;
-            } while( el );
+            } while(el);
 
             var stylesToSet = {
                 "left": offsetLeft + "px",
--- a/src/LiveDevelopment/Agents/RemoteFunctions.js
+++ b/src/LiveDevelopment/Agents/RemoteFunctions.js
@@ -296,8 +296,15 @@ function RemoteFunctions(config, remoteWSPort) {
               bottom: elementStyling.getPropertyValue('border-bottom-width')
             };
             
-            var innerWidth = elementBounds.width - parseInt(realElBorder.left) - parseInt(realElBorder.right);
-            var innerHeight = elementBounds.height - parseInt(realElBorder.top) - parseInt(realElBorder.bottom);
+            var borderBox = elementStyling.boxSizing === 'border-box';
+            
+            var innerWidth = parseInt(elementStyling.width) - parseInt(realElBorder.left) - parseInt(realElBorder.right);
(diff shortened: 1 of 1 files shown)


src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -413,12 +419,23 @@ function RemoteFunctions(config, remoteWSPort) {
             highlight.className = HIGHLIGHT_CLASSNAME;
 
             var offset = _screenOffset(element);
+            
+            var el = element,
+            offsetLeft = 0,
+            offsetTop  = 0;
+            
+            // Probably the easiest way to get elements position without including transform
+            do{
```
petetnt: Style nit: space between `do` and `{`

petetnt (approved):
Change set LGTM

swmitra (approved):
LGTM. Good that the 'transform' simulation will happen now on the highlight div.

swmitra:
Good work @Worie 👍 . We want more features like this from you.

swmitra:
@Worie I had to revert this PR as this breaks the positioning of the highlight div in case the target element is transformed with rotate function.

saurabh95:
The highlight area of div containing "ARCHITECTURE" is not positioned correctly.

image

Worie:
Thanks for the feedback. 

I know that there was an issue with `borders` when transforming an element, but I haven't seen any problems with rotate, nor I'm sure how  the site provided by @saurabh95 was implemented. Can you guys please give me a snippet with code which generates those issues? 

I've got a solution for border issue locally, too, so perhaps this would solve those two - shall I create another PR with my current implementation, or what?

saurabh95:
`transform-origin` is used in the above site

` #architecture .section-title {
    position: absolute;
    z-index: 5;
    -webkit-transform: rotate(90deg);
    -webkit-transform-origin: right bottom;
    transform: rotate(90deg);
    transform-origin: right bottom;
    right: 6.5%;
    top: 43.5%;
    left: auto;
    padding: 0 0 0 212px;
    margin: 0;
    text-transform: uppercase; }`

Worie:
Right, this property wasn't copied from original element to the highlight div so it had to be misaligned, now it should look good (at least on OSX Sierra). 
I've pushed my local changes, so I guess I'll create another PR, since this is closed/merged.

swmitra:
@Worie Great 👍 . You can use this branch itself, but raise a new PR.

PR #13295, opened 2017-04-10
CSS AtRules, Pseudo elements and Pseudo selector code hints

swmitra:
This PR adds CSS restricted block and pseudo selector/element code hints. In the implementation, priority of the hint providers have been kept at 0 while the priority of the default CSS code hint provider has been increased to 1. This has been done to ensure no additional overhead in case of CSS property:value code hints. Attached some screen shots of the corresponding code hints for illustration.
* @rules 
atrules

* Pseudo selectors 
pseudoselector

* Pseudo elements 
pseudoelements

Copying @ingorichter @petetnt @ficristo @zaggino @peterflynn  for code review.

Note: Additional description have been captured in the json def files to enable description tooltip at a later point of time.


src/extensions/default/CSSCodeHints/AtRuleCodeHints.js
```diff
@@ -0,0 +1,119 @@
...
+        var pos     = this.editor.getCursorPos(),
+            token   = this.editor._codeMirror.getTokenAt(pos);
+
+        this.filter = token.string;
+        this.token = token;
+
+        if (this.filter) {
+
+            // Filter the property list based on the token string
+            var result = $.map(Object.keys(AtRules), function (key, value) {
```
petetnt: Would prefer `Array.prototype.map` here, maybe with something like:

``` js
Object.keys(AtRules).map(function (key) {
  if (key.indexOf(token.string) === 0) {
    return key
  }
}).filter(function (key) {
  return key;
}).sort();
```

src/extensions/default/CSSCodeHints/PseudoSelectorHints.js
```diff
@@ -0,0 +1,171 @@
...
+    // As we are only going to provide :<pseudo> name hints
+    // we should claim that we don't have hints for anything else
+    PsudoSelectorHints.prototype.hasHints = function (editor, implicitChar) {
+        var pos = editor.getCursorPos(),
+            token = editor._codeMirror.getTokenAt(pos);
+
+        this.editor = editor;
+
+        // Check if we are at ':' pseudo rule or in 'variable-3' 'def' context
+        if (token.state.state === "pseudo" || token.type === "variable-3") {
```
petetnt: `return token.state.state === "pseudo" || token.type === "variable-3"`

src/extensions/default/CSSCodeHints/PseudoSelectorHints.js
```diff
@@ -0,0 +1,171 @@
...
+
+        // validate and keep the context in scope so that it can be used while getting description
+        this.context = _getPseudoContext(token, lineTillCursor);
+
+        this.token = token;
+
+
+        if (this.context !== -1) {
+            // Filter the property list based on the token string
+            var result = $.map(this.context === TOKEN_TYPE_PSEUDO_SELECTOR ? Object.keys(PseudoRules.selectors) : Object.keys(PseudoRules.elements), function (value, key) {
```
petetnt: Prefer `Array.prototype.map`, maybe do the context check first.

``` js
var keys = this.context === TOKEN_TYPE_PSEUDO_SELECTOR ? Object.keys(PseudoRules.selectors) : Object.keys(PseudoRules.elements);

var result = keys.map(...)
```

src/extensions/default/CSSCodeHints/PseudoSelectorHints.js
```diff
@@ -0,0 +1,171 @@
...
+
+            return {
+                hints: result,
+                match: filter,
+                selectInitial: true,
+                defaultDescriptionWidth: true,
+                handleWideResults: false
+            };
+        } else {
+            return false;
```
petetnt: The other similar function returns `null` but I guess it doesn't matter

src/extensions/default/CSSCodeHints/PseudoSelectorHints.js
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        this.editor.document.replaceRange(completion, startPos, endPos);
+
+        if (completion.slice(-1) === ")") {
+            cursor = this.editor.getCursorPos();
+            this.editor.setCursorPos({line: cursor.line, ch: cursor.ch - 1});
+        }
+
+        return false;
```
petetnt: Does this need to return anything?

src/extensions/default/CSSCodeHints/PseudoSelectorHints.js
```diff
@@ -0,0 +1,171 @@
...
+    };
+
+    function _init() {
```
petetnt: Empty method

src/extensions/default/CSSCodeHints/PseudoSelectors.json
```diff
@@ -0,0 +1,44 @@
+{
+    "selectors" :
+    {
+        "active" : { "desc": "Selects the active link"},
```
petetnt: Normalize whitespace in this file between the keys and values

src/extensions/default/CSSCodeHints/main.js
```diff
@@ -468,7 +472,7 @@ define(function (require, exports, module) {
 
     AppInit.appReady(function () {
         var cssPropHints = new CssPropHints();
-        CodeHintManager.registerHintProvider(cssPropHints, ["css", "scss", "less"], 0);
+        CodeHintManager.registerHintProvider(cssPropHints, ["css", "scss", "less"], 1);
```
petetnt: Q: What's the second parameter?

petetnt (changes requested):
Great job, some initial comments! The feature itself seems to work just fine!

ficristo:
This fixes #12486

src/extensions/default/CSSCodeHints/AtRuleCodeHints.js
```diff
@@ -0,0 +1,119 @@
...
+        this.editor = editor;
+
+        if (token.state.base && token.state.base.localState) {
+            cmState = token.state.base.localState;
+        } else {
+            cmState = token.state;
+        }
+
+        // Check if we are at '@' rule 'def' context
+        if ((token.type === "def" && cmState.context.type === "at") || (token.type === "variable-2" && (cmState.context.type === "top" || cmState.context.type === "block"))) {
```
ficristo: Split in two lines

src/extensions/default/CSSCodeHints/AtRuleCodeHints.js
```diff
@@ -0,0 +1,119 @@
...
+    };
+
+    AtRuleHints.prototype.getHints = function (implicitChar) {
+        var pos     = this.editor.getCursorPos(),
+            token   = this.editor._codeMirror.getTokenAt(pos);
+
+        this.filter = token.string;
+        this.token = token;
+
+        if (this.filter) {
```
ficristo: I prefer an early return here:
```js
if (!this.filter) {
  return null;
}

var result...
```

src/extensions/default/CSSCodeHints/AtRuleCodeHints.js
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * Inserts a given @<rule> hint into the current editor context.
+     *
+     * @param {string} hint
+     * The hint to be inserted into the editor context.
+     *
+     * @return {boolean}
+     * Indicates whether the manager should follow hint insertion with an
+     * additional explicit hint request.
```
ficristo: We should use an uniform style to doc comments: the description should be always inline or always at new line.
I think the current style is on the same line.

src/extensions/default/CSSCodeHints/AtRuleCodeHints.js
```diff
@@ -0,0 +1,119 @@
...
+        } else {
+            return null;
+        }
+    };
+
+
+    /**
+     * Inserts a given @<rule> hint into the current editor context.
+     *
+     * @param {string} hint
```
ficristo: Wrong param name

src/extensions/default/CSSCodeHints/AtRulesDef.json
```diff
@@ -0,0 +1,10 @@
+{
+    "@charset" : "Defines the character set used by the style sheet.",
+    "@import" : "Tells the CSS engine to include an external style sheet.",
+    "@namespace" : "Tells the CSS engine that all its content must be considered prefixed with an XML namespace.",
+    "@media" : "A conditional group rule which will apply its content if the device meets the criteria of the condition defined using a media query.",
+    "@supports" : "A conditional group rule which will apply its content if the browser meets the criteria of the given condition.",
+    "@page" : "Describes the aspect of layout changes which will be applied when printing the document.",
+    "@font-face" : "Describes the aspect of an external font to be downloaded.",
+    "@keyframes"  : "Describes the aspect of intermediate steps in a CSS animation sequence."
```
ficristo: In general I prefer no space before `:` and one after, but I think, in general, we use vertical align?

src/extensions/default/CSSCodeHints/PseudoSelectorHints.js
```diff
@@ -0,0 +1,171 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+
+/*jslint vars: true, plusplus: true, devel: true, nomen: true, indent: 4, maxerr: 50 */
+/*global define, brackets, $ */
```
ficristo: As before, remove.

src/extensions/default/CSSCodeHints/PseudoSelectorHints.js
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    /**
+     * @constructor
+     */
+    function PsudoSelectorHints() {
+    }
+
+    // As we are only going to provide :<pseudo> name hints
+    // we should claim that we don't have hints for anything else
```
ficristo: Use JSDoc and add the params

src/extensions/default/CSSCodeHints/PseudoSelectorHints.js
```diff
@@ -0,0 +1,171 @@
...
+            return null;
+        }
+
+        // validate and keep the context in scope so that it can be used while getting description
+        this.context = _getPseudoContext(token, lineTillCursor);
+
+        this.token = token;
+
+
+        if (this.context !== -1) {
```
ficristo: Early return.

src/extensions/default/CSSCodeHints/PseudoSelectorHints.js
ficristo: AFAICT unused, remove.

src/extensions/default/CSSCodeHints/main.js
ficristo: And why we need this change?

src/extensions/default/CSSCodeHints/AtRulesDef.json
ficristo: Curiosity: is it possible to translate these descriptions? I think it would be nice.

ficristo (changes requested):

ficristo:
I briefly tested, but I cannot use the pseudo selectors in a SCSS file, but I can in a CSS one.

Do you have thought of using multiple extensions like suggested by @redmunds in https://github.com/adobe/brackets/issues/12486#issuecomment-257583129?
I don't like much the `require`s with side effects in main.js.
Do you think is possible to have AtRuleCodeHints, PseudoSelectorHints and CssPropHints files which exports their class and a main.js file with a single `AppInit.appReady` function that create an instance for every of the three classes?

src/extensions/default/CSSCodeHints/main.js
swmitra: I have mentioned it in the opening note of the PR. 

> In the implementation, priority of the hint providers have been kept at 0 while the priority of the default CSS code hint provider has been increased to 1. This has been done to ensure no additional overhead in case of CSS property:value code hints.

To explain, the second parameter is for priority. CodeHintManager keeps the providers for a particular mode in order of priority, If there are priority collisions, then the lists are sorted based on loading order(this is not in our control). CodeHintManager first checks the current doc mode and then iterates over this sorted providers by calling 'hasHints', once it gets affirmative answer from a particular provider, it breaks there and selects that provider as the current session hint provider. By increasing the priority of the existing provider I am trying to ensure that, it gets the opportunity first to provide hints ( as hasHints() of this provider will be called first). This will ensure that, there is zero overhead on the current property:value hint provider.

petetnt: @swmitra Sorry I already forgot about the opening note of PR 😸 thanks for the detailed explanation!

ficristo: Same here...

swmitra:
@petent @ficristo I have addressed all the review comments in my last commit. Also extracted these providers as separate default extensions. Need your review once more 😄

src/extensions/default/CSSCodeHints/main.js
swmitra: 👍

ficristo:
In general LGTM, but I still cannot see pseudo selector hints for SCSS code.
Also add some tests.

I've use this snippet to test:
```scss
@charset "utf8";
/*comment*/

.test1: {
}

.test2:: {
}
```

swmitra:
@ficristo You probably have 'brackets-saas' extension installed. That extension registers 'SaasHints' provider with high priority and claims to provide hints ( 'hasHints' call returns true) even when it can't. 
Please disable that extension and try. It should work.

ficristo:
I disabled all my extensions and I can still reproduce.
If you write `.test:` in a SCSS file the hints work.
If you write `.test:: {}`, move the cursor after the last `:` and use Ctrl-space to ask hints, nothing happen. But the same works in a CSS file.

swmitra:
@ficristo WIth the SCSS tokenizer in CM2, the token identifier is not able to detect the pseudo context in the second scenario you have mentioned. I have filed an issue with CM https://github.com/codemirror/CodeMirror/issues/4688 just now. Lets wait for a response from @marijnh.

swmitra:
@ficristo Please refer to the CM issue and the discussion at https://github.com/codemirror/CodeMirror/issues/4688, we might have to live with this issue for sometime 😞 . But fortunately the moment you put the first char after `:`, you will get hints if the typed characters matches any of them. I tried to get some workaround with regex, but it's not going to work anyway as CM matches `/:\s*\{/` and combines them to a single token. So distinguishing between `/:\s*\{/` and `/::\s*\{/` is virtually impossible for me (unless someone can figure out a hack not dependent on the CM token).

marijnh:
Just figuring out what the SCSS mode is doing and fixing it is going to be a lot easier than trying to work around this from the outside.

swmitra:
Thanks @marijnh for joining in 👍 
I was just going through some of the syntax's in SCSS and most likely need to spend some more time to abstractly make sure that I don't break anything while fixing the CM CSS mode parser 😄

ficristo:
According to http://sass-lang.com/documentation/file.SASS_REFERENCE.html#nested_properties you can "split" properties like this:
```scss
.test {
  font: {
    family: serif;
  }
}
```
instead of writing
```scss
.test {
  font-family: serif;
}
```
I don't know what we can do about it for our case.

@swmitra is the "problem" also present with `::` ?

swmitra:
yes @ficristo. The regex used in the mode parser matches both the cases. This is what it does for punctuation `:`  handling in mode/css/css.js @line 752 - 
```javascript
":": function(stream) {
        if (stream.match(/\s*\{/))
          return [null, "{"];
        return false;
      }
```
Now if I remove this or try to provide a punctuation handler without wrapping the `/\s{/` , our property:value handling will go for a toss 😞 in the following SCSS construct 

```scss
.test {
  font: {
    family: serif;
  }
}
```
@marijnh Badly need your input here...

swmitra:
@marijnh I see that the following https://github.com/codemirror/CodeMirror/commit/2c46d1b45af0baba87804f451c566de256c8d390 fixes the issue 👍

swmitra:
@ficristo The issue in SCSS mode is fixed now using CM https://github.com/codemirror/CodeMirror/commit/2c46d1b45af0baba87804f451c566de256c8d390 and my last commit. Can you please check it once?

ficristo:
To see the last changes in action we need to wait for a CodeMirror upgrade, is it correct?
Also this PR needs some tests.

swmitra:
Yes @ficristo, I will upgrade CM once a release is available with the mentioned commit. I am writing tests for both the extensions...

src/extensions/default/CSSAtRuleCodeHints/main.js
```diff
@@ -0,0 +1,117 @@
...
+
+        this.filter = token.string;
+        this.token = token;
+        
+        if (!this.filter) {
+            return null;
+        }
+
+        // Filter the property list based on the token string
+        var result = Object.keys(AtRules).map(function (key) {
```
ingorichter: I might be wrong, but I think `filter` would be more appropriate than `map` for the first loop. This would also make the following `filter` loop unnecessary.

src/extensions/default/CSSPseudoSelectorHints/PseudoSelectors.json
```diff
@@ -0,0 +1,44 @@
+{
+    "selectors" :
+    {
+        "active":              {"desc": "Selects the active link"},
+        "checked":             {"desc": "Selects every checked <input> element"},
+        "disabled":            {"desc": "Selects every disabled <input> element"},
+        "empty"	:              {"desc": "Selects every element that has no children/text (including text nodes)"},
```
ingorichter: nit: formatting is a bit off

src/extensions/default/CSSPseudoSelectorHints/PseudoSelectors.json
```diff
@@ -0,0 +1,44 @@
...
+        "read-write":          {"desc": "Selects input elements with the 'readonly' attribute NOT specified"},
+        "required":            {"desc": "Selects input elements with the 'required' attribute specified"},
+        "root":                {"desc": "Selects the document's root element"},
+        "target":              {"desc": "Selects the current active element (clicked on a URL containing that anchor name)"},
+        "valid":               {"desc": "Selects all input elements with a valid value"},
+        "visited":             {"desc": "Selects all visited links"}
+    },
+    "elements" :
+    {
+        "after":        {"desc": "Insert something after the content of each element identified by this selctor"},
```
ingorichter: nit: `selector`

src/extensions/default/CSSPseudoSelectorHints/PseudoSelectors.json
```diff
@@ -0,0 +1,44 @@
...
+        "before":       {"desc": "Insert something before the content of each element identified by this selctor"},
```
ingorichter: nit: same `selector` spelling

ingorichter: basically all instances of `selctor` in this object ;-)

src/extensions/default/CSSPseudoSelectorHints/main.js
```diff
@@ -0,0 +1,176 @@
...
+
+        this.token = token;
+
+        // Filter the property list based on the token string
+        var result = Object.keys(this.context === TOKEN_TYPE_PSEUDO_SELECTOR ? PseudoRules.selectors : PseudoRules.elements).map(function (key) {
+            if (key.indexOf(filter) === 0) {
+                return key;
+            }
+        }).filter(function (key) {
+            return key;
```
ingorichter: replacing the previous `map` call with `filter` should make this obsolete

src/extensions/default/CSSPseudoSelectorHints/main.js
```diff
@@ -0,0 +1,176 @@
...
+    // As we are only going to provide :<pseudo> name hints
+    // we should claim that we don't have hints for anything else
+    PsudoSelectorHints.prototype.hasHints = function (editor, implicitChar) {
+        var pos = editor.getCursorPos(),
+            token = editor._codeMirror.getTokenAt(pos);
+
+        this.editor = editor;
+
+        // Check if we are at ':' pseudo rule or in 'variable-3' 'def' context
+        return token.state.state === "pseudo" || token.type === "variable-3" || token.string === PUNCTUATION_CHAR;
```
ingorichter: nit: this could be a function determining the state. The same statement (negated) is used in `getHints` and could be reused there. Plus, it makes testing this function in isolation easier.

ingorichter (commented):
A couple of nits. Otherwise LGTM

swmitra:
@petetnt @ingorichter @ficristo Can you guys please have a look at this PR now? 
CM dependency has been resolved with #13331 and also I have addressed all review comments and added unit tests for all style modes (css, less, scss).

ficristo:
There is still the nit of @ingorichter to fix. No time to look again.

ficristo (approved):

petetnt (approved):

zaggino (approved):

PR #13302, opened 2017-04-13
Added UI for search History

saurabh95:
The UI is inspired from https://trello.com/c/IzACddEv/344-s-find-recent-search-history with minor tweaks.

Dark Theme

image

Light Theme

image

The UI can be invoked using arrow up and down keys or clicking on the down arrow on the right of the search input.
The UI is reused from QuickSearchField.js which is also used by Quick Open.

### DIFF
--- a/src/search/FindBar.js
+++ b/src/search/FindBar.js
@@ -408,6 +408,10 @@ define(function (require, exports, module) {
         this.focusQuery();
     };
 
+    /**
+     * @private
+     * Shows the search History in dropdown.
+     */
     FindBar.prototype.showSearchHints = function () {
         var self = this;
         this.$searchField = $("input#find-what");
--- a/src/search/FindBar.js
+++ b/src/search/FindBar.js
@@ -491,8 +491,8 @@ define(function (require, exports, module) {
                 //self.trigger("queryChange");
                 self.searchField.destroy();
             },
-            onHighlight: function (selectedItem, query, explicit) {}
-            //verticalAdjust: 44
+            onHighlight: function (selectedItem, query, explicit) {},
+            highlightZeroIndex: false
         });
         this.searchField.setText($("#find-what").val());
     };
--- a/src/search/FindBar.js
+++ b/src/search/FindBar.js
@@ -332,7 +332,7 @@ define(function (require, exports, module) {
                 if (e.keyCode === KeyEvent.DOM_VK_RETURN) {
                     e.preventDefault();
                     e.stopPropagation();
-                    var searchVal = self.$("#find-what").val()
+                    var searchVal = self.$("#find-what").val();
                     var searchQueryIndex = searchHistory.indexOf(searchVal);
                     if (searchQueryIndex !== -1) {
                         searchHistory.splice(searchQueryIndex, 1);
--- a/src/search/FindBar.js
+++ b/src/search/FindBar.js
@@ -495,7 +495,7 @@ define(function (require, exports, module) {
             //verticalAdjust: 44
         });
         this.searchField.setText($("#find-what").val());
-    }
+    };
 
     /**
      * Closes this Find bar. If already closed, does nothing.
--- a/src/search/FindBar.js
+++ b/src/search/FindBar.js
@@ -434,11 +434,7 @@ define(function (require, exports, module) {
             maxResults: 20,
             resultProvider: function (query) {
                 var asyncResult = new $.Deferred();
-                var searchHistory = PreferencesManager.getViewState("searchHistory");
-                var filteredResults = _.filter(searchHistory, function (s) {
-                    return s.indexOf(query) !== -1;
-                });
-                asyncResult.resolve(filteredResults);
+                asyncResult.resolve(PreferencesManager.getViewState("searchHistory"));
                 return asyncResult.promise();
             },
             formatter: function (item, query) {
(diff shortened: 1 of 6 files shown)


ficristo:
I leave the review for someone else here.
I only have one question: when you select an item from the dropdown list and then reopen the list, only the item selected is visible. Is this expected?
If I need to change the item, I have to clear the content of the search box and only then I'm able to see other items in the dropdown list, which is a bit of a pain.

Do you still have an UI / UX designer that can review these kind changes ?

saurabh95:
@ficristo Actually I have added a filtering in the dropdown, if there is some text in the search field and then we try to open dropdown, then only elements in dropdown which have that string as its substring would be visible.

saurabh95:
@ficristo discussed it with Rajnish, he also feels that removing the filtering would be better, so now whole search history is visible irrespective of content in the search field

src/search/FindBar.js
```diff
@@ -281,6 +285,17 @@ define(function (require, exports, module) {
                     self.trigger("doFind");
                 }
             })
+            .on("click", ".dropdown-icon", function (e) {
+                if (!self.searchField) {
+                    self.showSearchHints();
+                } else if ($(".quick-search-container").is(':visible')) {
+                    $(".quick-search-container").hide();
+                } else if (!$(".quick-search-container").is(':visible')) {
```
petetnt: This can be just `else`, right?

src/search/FindBar.js
```diff
@@ -324,7 +339,9 @@ define(function (require, exports, module) {
                             searchHistory.pop();
                         }
                     }
-                    searchHistory.unshift($('#find-what').val());
+                    if ($('#find-what').val()) {
+                        searchHistory.unshift($('#find-what').val());
```
petetnt: Maybe cache the `#find-what` value to a variable

src/search/FindBar.js
```diff
@@ -281,6 +285,17 @@ define(function (require, exports, module) {
                     self.trigger("doFind");
                 }
             })
+            .on("click", ".dropdown-icon", function (e) {
+                if (!self.searchField) {
+                    self.showSearchHints();
+                } else if ($(".quick-search-container").is(':visible')) {
```
petetnt: `${".quick-search-container"}` could be cached to a variable

src/search/FindBar.js
```diff
@@ -348,13 +365,11 @@ define(function (require, exports, module) {
                     }
                     historyIndex = 0;
                 } else if (e.keyCode === KeyEvent.DOM_VK_DOWN || e.keyCode === KeyEvent.DOM_VK_UP) {
-                    if (e.keyCode === KeyEvent.DOM_VK_DOWN) {
-                        historyIndex = (historyIndex - 1 + searchHistory.length) % searchHistory.length;
-                    } else {
-                        historyIndex = (historyIndex + 1 + searchHistory.length) % searchHistory.length;
+                    if (!self.searchField) {
+                        self.showSearchHints();
+                    } else if (!$(".quick-search-container").is(':visible')) {
```
petetnt: Use the cached variable here too (see the previous review comment)

src/search/FindBar.js
```diff
@@ -407,6 +422,41 @@ define(function (require, exports, module) {
     };
 
     /**
+     * @private
+     * Shows the search History in dropdown.
+     */
+    FindBar.prototype.showSearchHints = function () {
+        var self = this;
+        this.$searchField = $("input#find-what");
```
petetnt: This part (L430-L432) with nested `this.$searchField` looks a bit scary, maybe refactor it to use some named variables instead.

src/search/FindBar.js
```diff
@@ -281,6 +285,17 @@ define(function (require, exports, module) {
...
+            .on("click", ".dropdown-icon", function (e) {
+                if (!self.searchField) {
+                    self.showSearchHints();
+                } else if ($(".quick-search-container").is(':visible')) {
+                    $(".quick-search-container").hide();
+                } else if (!$(".quick-search-container").is(':visible')) {
+                    self.searchField.setText($("#find-what").val());
+                    $(".quick-search-container").show();
+                }
+                self.$("#find-what").focus();
```
petetnt: What's the difference between `self.$("#find-what")` and `$("#find-what")`?

petetnt (changes requested):
Added a few comments, mostly looks really great 👍

src/search/FindBar.js
swmitra: I second @petetnt. Value of `#find-what` should be assigned to a variable while initiating the search session. Whole session should refer to the same value for navigation.

src/search/FindBar.js
saurabh95: Actually there are chances of having multiple find bars present (one is closing, other might be opening), so there can be multiple `$("#find-what")`, so to add context to it we use `self.$("#find-what")`

swmitra (approved):
LGTM 👍

swmitra:
@petetnt Once you approve, I will merge this PR 😄

petetnt (approved):
LGTM codewise, haven't had the time to retest the functionality though.

swmitra:
@saurabh95 Great work :+1

swmitra:
Merging...

Issue #13282, opened 2017-04-07
Navigating between large, code folded files causes performance problems


lkcampbell:
**OS:** MacS 10.12.3 Sierra and Windows 10

**Brackets Version:** Release 1.9 build 1.9.0-17312 (release 189f6d39a) 

**Repro  Steps:**
1. Make sure `code-folding.enabled` preference is set to `true`.
1. Make sure `code-folding.saveFoldStates` preference is set to `true`.
1. Set `code-folding.maxFoldLevel` preference to a high number like `100`. Collapse all should really be collapsing *all* folds.
1. Create two decent sized xml test files. I'm using [this file](https://raw.githubusercontent.com/jmandel/sample_ccdas/master/HL7%20Samples/CCD.sample.xml) for both of my test files.
1. Open both files so they are in the `Working Files` section of the Project Sidebar.
1. `Collapse All` both files so there is plenty of fold state data.
1. Navigate back and forth between the two files continuously, showing one, then the other, in the editor.

**Observed Results:**
After three or four loops of navigating to each file, the time it takes for the contents of each file to show up in the editor will get longer and longer, roughly doubling in time for each loop. Eventually, the delay is so long the files can no longer be navigated in a reasonable amount of time.

**Expected Results:**
The time to navigate to each test file should be about the same regardless of how many time you loop between them.

**Workaround:**
Make sure `code-folding.saveFoldStates` preference is set to `false`. You will need PR https://github.com/adobe/brackets/pull/13269 before you can use this workaround.

**Other Notes:**
The problem occurs with all extensions disabled.

Since this problem goes away when fold states are turned off, and the time seems to roughly be doubling each time, maybe new fold states are being appended to old fold states instead of replacing them. Feels like there might be some redundant, duplicate data being saved somewhere, just a theory, haven't tested it.

lkcampbell:
Pinging @thehogfather, wanted to make sure you saw this issue, Patrick.

thehogfather:
@lkcampbell got it 👍 .

lkcampbell:
@thehogfather let me know if you need help. I am curious if my theory is correct. Either way, let me know.

thehogfather:
@lkcampbell your theory is in the right direction - but we could nail it down better. The problem is not actually with the routine for saving fold states. The issue occurs when we attempt to restore the saved fold states for the editor gaining focus.

Here is a quick overview of what I have found so far.
[...]

PR #13310, opened 2017-04-18
Addressed performance problems when navigating between large code folded files

thehogfather:
Addresses #13282 by preventing reinitialisation of the fold gutter for editors whose fold gutters have already been initialised.

cc @lkcampbell


swmitra (approved):
Changeset looks good to me. Good catch @thehogfather to prevent re-initialization 👍

Issue #13189, opened 2017-03-19
Brackets 1.9 and earlier: If you have multiple stylesheets in Quick Edit panel the code is rendered with Grayscale antialiasing (a bit blurry text) instead of ClearType


r4p7or:
Windows 7 64bit
**Brackets Release 1.9 build 1.9.0-17305**
(font rendering on my system is enhanced with MacType (for sharper text than default Win7 ClearType) - but this should be irrelevant)

Hi, I've noticed that Brackets in some cases uses inferior Grayscale antialiasing (a bit blurry text not comfortable for the eyes) instead of sharper ClearType.
The most serious case is that Quick Edit switches to Grayscale antialiasing if you have more than one stylesheet in the QE panel (when a list with multiple stylesheets is displayed on the right): 

![quickedit_fontrendering_1](https://cloud.githubusercontent.com/assets/26348847/24081689/ae8637f4-0cb8-11e7-9700-d6815277d5dd.png)

One interesting observation - if you fold **all** the stylesheets in the stylesheet list, font rendering inside the Quick Edit panel switches back to ClearType.
I have made an animated GIF illustrating this strange behavior:

![quickedit_fontrendering_2](https://cloud.githubusercontent.com/assets/26348847/24081690/b81d831c-0cb8-11e7-98e6-c0139d0bca17.gif)

This bug also manifests itself in the right click menu:

![quickedit_fontrendering_3](https://cloud.githubusercontent.com/assets/26348847/24081693/be346d6a-0cb8-11e7-87ad-556bcc77b684.png)

Another examples of a text with Grayscale (blurry) antialiasing in the Brackets application:

![quickedit_fontrendering_4](https://cloud.githubusercontent.com/assets/26348847/24081694/c2e25516-0cb8-11e7-9f57-4c650436d9e8.png)

Blackweda:
This looks like something I'll look at.

r4p7or:
Hi, thank you for trying to fix this issue.

> So my question is, "Is this something I should try to fix.. or its not considered a real 'problem' for brackets right now?"

I think the blurry text in the Quick Edit panel is very real problem and it would be great if it was fixed ASAP.

As for the other cases of Grayscale AA inside the application - it's not exactly a show stopper for me right now but one day it would be great if developers finally solved all the font rendering issues once and for all (Brackets has always had various issues of font rendering from the start). 

> &.platform-mac {

[...]

PR #13318, opened 2017-04-21
Altered blurry (antialiased) for crisper (subpixel-antialiased) settings

Blackweda:
This PR is a fix for issue #13189 

For Brackets 1.9 and earlier, there were instances where the Quick Edit panel and the Right-Click pop up box would render the wording in a blurred Grayscale format, which looked unsettling for the user.

I read through all the relevant /src code files, and found instances of antialiasing in 3 files:

src/view/ViewCommandHandlers.js
src/styles/brackets.less
src/styles/brackets_patterns_override.less

In attempting to test this bug with the current Brackets 1.10, I found that I couldn't replicate the issue as I could when I had 1.9. Here are some of my thoughts on why:

In src/styles/font/**SourceSansPro**, we can see this font is preferred to be used with grayscale antialiasing. However, when we go over to src/view/ViewCommandHandlers, we see that the default font of choice is actually " 'SourceCodePro-Medium', ＭＳ ゴシック, 'MS Gothic', monospace ".

I believe that Brackets is no longer using a font that is coded to display itself in grayscaling anymore.

Furthermore, in **brackets.less** as well as **ViewCommandHandlers.js**, both of these files are actually referring to the Mac platform's ability to choose antialiasing as a preference, however, its default is actually already subpixel-antialiasing (which we want). Therefore I left them alone as they are.

The changes I made are taking the brackets pattern overrides which explicitly call for antialiasing, and changing them back to subpixel-antialiasing for a cleaner, crisper look. 

These changes targeted non-editable input fields, input placeholders, and buttons.

With this PR, my hope is that users will no longer have any issues of seeing blurred displays of text in Brackets (unless they are a Mac user and really want it).

There is no where else in the entire /src folder where I found any instance of affecting the display of the Brackets editor or its components except for the files I mentioned above.

For a more in-depth explanation of font-smoothing and antialiasing techniques, feel free to read:
http://szafranek.net/blog/2009/02/22/font-smoothing-explained/

### DIFF
--- a/src/styles/brackets_patterns_override.less
+++ b/src/styles/brackets_patterns_override.less
@@ -1430,7 +1430,7 @@ input[type="color"],
     color: @bc-text-thin;
     font-size: (@baseFontSize + 1px);
     font-weight: normal;
-    -webkit-font-smoothing: antialiased;
+    -webkit-font-smoothing: subpixel-antialiased;
     box-shadow: inset 0 1px 0 @bc-shadow-small;
     border-radius: @bc-border-radius;
     .transition(~"border linear 180ms, box-shadow linear 180ms");
@@ -1445,7 +1445,7 @@ input[type="color"],
     &::-webkit-input-placeholder {
         color: @bc-text-thin-quiet;
         font-weight: normal;
-        -webkit-font-smoothing: antialiased;
+        -webkit-font-smoothing: subpixel-antialiased;
 
         .dark & {
             color: @dark-bc-text-thin-quiet;
@@ -1572,7 +1572,7 @@ input[type="color"],
     border: 1px solid @bc-btn-border;
     border-radius: @bc-border-radius;
     box-shadow: inset 0 1px @bc-highlight-hard;
-    -webkit-font-smoothing: antialiased;
+    -webkit-font-smoothing: subpixel-antialiased;
     text-shadow: none;
     margin: 3px 0 3px 3px;
 


petetnt:
Cool stuff @Blackweda! 

Could you add some before & after screenshots? Have you had the change to test this change on other platforms than your preferred one? 👍

Blackweda:
I actually meant to ask you guys.. I followed the steps for 'Hacking on Brackets'.
The problem is, no matter what version of Brackets I downloaded (say Brackets 1.9), when I setup_for_hacking, it will use the most current version of Brackets source code (i.e. Brackets 1.10).
How do I play with older versions of source code so that I can genuinely test the effectiveness of the code on older versions where the issues were made??

I realize this might be a silly question, since 'WHY would anybody want to test backwards when the current version works better??"

petetnt:
You can go backwards by either:

1. Running `restore_installed_build` in the same tools folder to return it to the currently installed version
2. Navigating to an older commit with `git checkout [revision] .`. The revision is a `commit hash` , for example for `1.9` it's `189f6d39abf1651454feabd8ed01d371eadf2628` so `git checkout 189f6d39abf1651454feabd8ed01d371eadf2628 .` should work (notice the dot in the end, it's important). You can see the commit hashes for releases with https://github.com/adobe/brackets/releases.

Blackweda:
I'm going to redo this PR and resubmit soon.

Issue #11814, opened 2015-10-13
LOC: Japanese: The CC-BY 3.0 Unported link doesn't redirect to the Japanese page


Zhanxin:
Steps:
1. Launch the Brackets
2. Click Help->About Brackets
3. Click the link "CC-BY 3.0 Unported"

Result: The link would redirect to the English page
Expect: It should redirect to the Japanese page
Not reproduce on French
![linknotlocalized](https://cloud.githubusercontent.com/assets/2245179/10453925/112f25a4-71e3-11e5-8a87-56b1c2d21f02.png)

Blackweda:
Took care of it.

Zhanxin:
Verified as fixed. It links to https://creativecommons.org/licenses/by-sa/2.5/deed.ja now

PR #13321, opened 2017-04-21
Changed url to access Japanese translated page

Blackweda:
This PR is to fix issue #11814 

The issue was regarding the **Help -> About Brackets** when using the Japanese language. The 'CC-BY 3.0 Unported' link was set to the default. I got the link for the Japanese page and changed the urls.js in order to reflect that change. 

This solves the issue and I think its safe to close it.


ficristo:
Recently we switched to MDN docs, it seems to me the same license. Probably we should rename all those keywords to reflect the current state. @saurabh95 any input here?

But since it doesn't hurt, merging.
Thanks @Blackweda

saurabh95:
@ficristo Yeah we should change those keywords.
Also I was thinking of having docs available in French and Japanese as well, since the translations are provided by MDN only, it seems possible by changing URLs.

PR #13334, opened 2017-04-24
add: flow-root as value for display

praveenpuglia:
`display` now has a new value `flow-root` which is also getting browser support. Both Chrome and Firefox support it. Should be handy to have as an autocompletion value.


praveenpuglia:
Signed up the CLA.

petetnt (approved):
LGTM, thanks!

PR #13342, opened 2017-04-28
Now initial query is also added to searchHistory queue

saurabh95:
If the search bar in invoked and with some string already filled, then it will add that string in search history.


ficristo:
Could you clarify what it fixes? Is it possible to add a test?

saurabh95:
@ficristo Added some explanation in description

swmitra (approved):

petetnt (approved):

zaggino (approved):

PR #13353, opened 2017-05-04
replace parseInt with parseFloat in RemoteFunctions (LP boxmodel fix)

Worie:
Just realized that using parseInt when calculating box-model visualization position/size has sometimes caused weird visual glitches (especially with custom config of it). So this basicly replaces all parseInt used to calculate the box-model to parseFloat and solves this issue.


petetnt (approved):
LGTM

Issue #13356, opened 2017-05-05
Visual regression testing for Live Preview boxmodel


Worie:
My recent changes in `RemoteFunctions.js` related to implementation of boxmodel visualization in Live Preview made this feature break in several edge cases. We should probably create some visual regression tests for this feature to prevent such cases to happen more often.

What do you think? Screenshot comparsion for specific set of properties could work.

PR #13357, opened 2017-05-05
Fix box model regression (transform)

Worie:
Hi. In my recent commit https://github.com/adobe/brackets/commit/ad915c721ffe379d47e95335eaf2ddd1f6838a6f for whatever reason I've removed a piece of code that made showing the box-model of transformed elements appear OK. I must have discarded the wrong hunk and did not refresh Brackets afterwards, so I didn't see it was broken. I'm very sorry for that.

This PR fixes this regression. However, we should consider visual testing to this feature to prevent such cases #13356 .

For now, could you ensure that it looks good for transformed elements, too (with or without borders, etc). I have tested it, but as you can see you can never be too cautios.

### DIFF
--- a/src/LiveDevelopment/Agents/RemoteFunctions.js
+++ b/src/LiveDevelopment/Agents/RemoteFunctions.js
@@ -431,19 +431,37 @@ function RemoteFunctions(config, remoteWSPort) {
             highlight.className = HIGHLIGHT_CLASSNAME;
 
             var offset = _screenOffset(element);
+            		
+            var el = element,		
+            offsetLeft = 0,		
+            offsetTop  = 0;		
+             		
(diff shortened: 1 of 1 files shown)


petetnt (approved):

petetnt:
Meanwhile visual tests you could write a test project that showcases the possible options too 🤔 I think I reviewed that commit, sorry for not catching it myself.

Worie:
What about this one? Shall we merge it? There's no reason to keep it hanging, as the master still has  broken version of RemoteFunctions.js. 

I remember about the showcases with the options but I'm quite busy recently, so it'll have to wait.

petetnt:
Thanks @Worie

PR #13368, opened 2017-05-10
Fixed Project Manager failing tests

saurabh95:
There are some issues in `copy` api, I am trying to fix that. Until then we can use `copyPath` to run tests. Please note that `copyPath` cannot copy binary files, so we can't use the same API in `LowLevelFileIO` and `FindInFiles` Tests which are also failing as they use test files involving some images.
Once copy API is fixed, then we can fix the failing tests in `LowLevelFileIO` and `FindInFiles` as well.

### DIFF
--- a/test/spec/ProjectManager-test.js
+++ b/test/spec/ProjectManager-test.js
@@ -50,7 +50,7 @@ define(function (require, exports, module) {
 
             // copy files to temp directory
             runs(function () {
-                waitsForDone(SpecRunnerUtils.copy(testPath, tempDir), "copy temp files");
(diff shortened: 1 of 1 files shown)


petetnt:
Changeset LGTM, but what happened to the `copy` API, it used to work right? 🤔

saurabh95:
It is always failing in test domain. I am yet to find a reason, when I last debugged it was sending message through websocket, there might be an issue with the shell side.

zaggino (approved):

zaggino:
It'd be good to figure out what the problem is, but LGTM

petetnt (approved):

swmitra:
Merging.
Failed tests are passing on win environment. Not tested on mac though...

PR #13371, opened 2017-05-10
Filled out Portuguese menu options

Blackweda:
This is a PR to solve issue #4938 .
I have added the necessary menu items that were missing as far as I can tell.
Translation checks may be needed, but its a start at filling out the entire menu for Portuguese.

### DIFF
--- a/src/nls/pt-pt/strings.js
+++ b/src/nls/pt-pt/strings.js
@@ -114,7 +114,8 @@ define({
      * ProjectManager
      */
 
-    "UNTITLED" : "Sem título",
+    "UNTITLED" 		                : "Sem título",
+    "WORKING_FILES"                     : "Arquivos de trabalho",
 
     /**
      * Keyboard modifier names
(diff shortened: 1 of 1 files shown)


src/nls/pt-pt/strings.js
```diff
@@ -114,7 +114,8 @@ define({
      * ProjectManager
      */
 
-    "UNTITLED" : "Sem título",
+    "UNTITLED" 		                : "Sem título",
```
ficristo: Use only spaces not tabs.

ficristo:
@Blackweda do you know someone who can review this?

PS: if you have problems with multiple forks, you can use `git remote` to help to track down the remote you want to use. I mention this because you said a couple of times you deleted your fork.

Blackweda:
@ficristo yes I can probably get someone to help confirm the translations... like I said, I'm new to using Git commands and stuff.. so I make mistakes and use the brute-force easiest way to fix things for now.

Issue #13214, opened 2017-03-23
Change the name of the Brackets Node executable


ficristo:
### Description

I had trouble to install some extensions on OSX. The problem was that the `node` command wasn't found.
I think the problem is the executable name of node process of Brackets: on Windows is `node.exe`, and on Windows 10 I don't have any problem, while on macOS and Linux is `Brackets-node`.
I don't have proof this is the real problem but to my eyes seems very likely.

On https://github.com/adobe/brackets/issues/10658 there were some discussion to rename the process on Windows. Maybe we should instead rename the others?

### Steps to Reproduce

1. Go to the extension manager
2. Try to install the npm registry extension

**Expected behavior:** the extension should be installed

**Actual behavior:** the installation fails

### Versions

OSX
1.9

Issue #12814, opened 2016-10-04
Windows 64-bit build target


timothydewolf:
No 64-bit build target on Windows

Attempting a 64-bit build on Visual Studio 2015 yields errors.  I updated my code locally to allow the 64-bit build to proceed, meaning fixing things like moving GetWindowLong=>GetWindowLongPtr in the NC painting code and numerous size_t <=> int casts.

The Gruntfile does not contain a 64-bit build target.  Would it be possible to get this 64-bit target into git?  I think it would be nice for those wanting to build their own (64-bit CEF) apps on brackets-shell.

ficristo:
While I agree, I think Adobe should upload a CEF for Windows 64-bit first. Then probably there is some work to do, as you already found.
If you really need a 64bit Brackets you can try https://github.com/zaggino/brackets-electron/

Denisov21:
+1

PR #13384, opened 2017-05-20
On Windows force a 32 bit build until nodejs 64 bit is supported

ficristo:
Since Brackets doesn't support nodejs 64 bit, I needed to force to compile my [extension](https://github.com/ficristo/brackets-terminal-x) to 32 bit by adding a .npmrc file with the right flags.
But with that, users on unix platform cannot install the extension from the extension manager anymore.
Adding those flags directly in Brackets should let me remove the .npmrc file.

### DIFF
--- a/src/extensibility/node/npm-installer.js
+++ b/src/extensibility/node/npm-installer.js
@@ -37,22 +37,13 @@ var Errors = {
  * Private function to run "npm install --production" command in the extension directory.
  *
  * @param {string} installDirectory Directory to remove
+ * @param {array} npmOptions can contain additional options like `--production` or `--proxy http://127.0.0.1:8888`
  * @param {function} callback NodeJS style callback to call after finish
  */
 function _performNpmInstall(installDirectory, npmOptions, callback) {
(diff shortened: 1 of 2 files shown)


petetnt (approved):
Changeset LGTM, what's stopping us from using the 64bit NodeJS?

ficristo:
AFAIK automation doesn't have 32bit builds (see also https://github.com/adobe/brackets-shell/pull/595) and we don't support win64 shell builds (https://github.com/adobe/brackets/issues/12814).

ficristo:
@petetnt did you test this on Windows?
If soù, is there someone from @adobe/brackets-committers that could test this on a non Windows platform and merge? Thanks.

saurabh95:
@ficristo I tested it on MAC, it is working for some extensions, but failing for Brackets Terminal X

ficristo:
What is the error ?
Two things come to my mind:
https://github.com/adobe/brackets/issues/13214
and my extension has still the .npmrc file

ficristo:
I tryed this branch on Linux and I was able to install some extensions.
Merging so to have more time to test it.

PR #13412, opened 2017-06-02
Saurabh95/encoding support

saurabh95:


petetnt:
Oh cool, this is very nice 🙌

src/document/Document.js
```diff
@@ -758,6 +758,22 @@ define(function (require, exports, module) {
         return this.file instanceof InMemoryFile;
     };
 
+    /**
+     *  Reloads the document from FileSystem
+     *  @return {promise} - to check if reload was successful or not
+     */
+    Document.prototype.reload = function () {
```
swmitra: You might want to use a separate promise like what we do in other places.

src/editor/EditorStatusBar.js
```diff
@@ -81,6 +89,22 @@ define(function (require, exports, module) {
...
 
     /**
+     * Update encoding
+     * @param {Editor} editor Current editor
+     */
+    function _updateEncodingInfo(editor) {
+        var doc = editor.document;
+
+        // Ensure width isn't left locked by a previous click of the dropdown (which may not have resulted in a "change" event at the time)
+        encodingSelect.$button.css("width", "auto");
```
swmitra: Define it once in CSS.

src/document/DocumentCommandHandlers.js
```diff
@@ -903,6 +919,7 @@ define(function (require, exports, module) {
 
             // First, write document's current text to new file
             newFile = FileSystem.getFileForPath(path);
+            newFile._encoding = doc.file._encoding;
```
swmitra: We can actually write the explicitly selected encoding for path even before we do a 'getFileForPath()'. In that way our existing logic of picking up a user preferred encoding will take care of the rest. With the current code we are actually missing the user preference completely for the new file and in next launch of the new file, encoding detection might result in a corruption where the user might have to select a suitable encoding again.

swmitra (changes requested):

swmitra (approved):

PR #13418, opened 2017-06-05
Sequential navigation in edit history

swmitra:
This feature is something which I missed a lot! Remembering most recently used (interesting?) cursor positions and navigating there manually is really cumbersome. 

This module adds the feature of remembering  cursors/selections across full editors inside a project root and aids the user by providing 'Navigate Back'(Alt+I) and 'Navigate Fwd'(Alt+Shift+I) using the captured navigation frames.

Ping @petetnt @ficristo for review.

### DIFF
--- a/src/extensions/default/NavigationAndHistory/NavigationProvider.js
+++ b/src/extensions/default/NavigationAndHistory/NavigationProvider.js
@@ -113,7 +113,7 @@ define(function (require, exports, module) {
                 } else {
                     deferred.reject();
                 }
-	    	});
+			});
 		}
 
         return deferred.promise();
@@ -158,7 +158,7 @@ define(function (require, exports, module) {
     
    /**
     * Function to create CM TextMarkers for the navigated positions/selections.
-    * This logic is required to ensure that the captured navaigation positions 
+    * This logic is required to ensure that the captured navigation positions 
     * stay valid and contextual even when the actual document text mutates.
     * The mutations which are handled here :
     * -> Addition/Deletion of lines before the captured position
@@ -166,8 +166,8 @@ define(function (require, exports, module) {
     */
     NavigationFrame.prototype._createMarkers = function (ranges) {
         var range,
-	    	index,
-	    	bookMark;
+			index,
+			bookMark;
 
         this.bookMarkIds = [];
         for (index in ranges) {
@@ -325,14 +325,14 @@ define(function (require, exports, module) {
                 CommandManager.execute(NAVIGATION_JUMP_BACK);
             }).always(function () {
 				_validateNavigationCmds();
-	    	});
+			});
         }
     }
     
    /**
     * Command handler to navigate forward
     */
-    function _navigateFwd() {
+    function _navigateForward() {
         var navFrame = jumpedPosStack.pop();
 		
 		// Check if the poped frame is the current active frame
@@ -375,7 +375,7 @@ define(function (require, exports, module) {
     */
     function _initNavigationCommands() {
         CommandManager.register(Strings.CMD_NAVIGATE_BACKWARD, NAVIGATION_JUMP_BACK, _navigateBack);
-        CommandManager.register(Strings.CMD_NAVIGATE_FORWARD, NAVIGATION_JUMP_FWD, _navigateFwd);
+        CommandManager.register(Strings.CMD_NAVIGATE_FORWARD, NAVIGATION_JUMP_FWD, _navigateForward);
         commandJumpBack = CommandManager.get(NAVIGATION_JUMP_BACK);
         commandJumpFwd = CommandManager.get(NAVIGATION_JUMP_FWD);
         commandJumpBack.setEnabled(false);
--- a/src/extensions/default/NavigationAndHistory/NavigationProvider.js
+++ b/src/extensions/default/NavigationAndHistory/NavigationProvider.js
@@ -49,7 +49,8 @@ define(function (require, exports, module) {
         NAVIGATION_JUMP_FWD       = "navigation.jump.fwd";
     
     // The latency time to capture an explicit cursor movement as a navigation frame
-    var NAV_FRAME_CAPTURE_LATENCY = 2000;
+    var NAV_FRAME_CAPTURE_LATENCY = 2000,
+		MAX_NAV_FRAMES_COUNT = 50;
     
     /*
     * Contains list of most recently known cursor positions.
@@ -66,16 +67,26 @@ define(function (require, exports, module) {
     var jumpedPosStack = [],
         activePosNotSynced = false,
         captureTimer,
+		currentEditPos,
         jumpInProgress,
         commandJumpBack,
         commandJumpFwd;
+		
+	/**
+    * Function to check if there are any navigatable frame backward.
+    * @private
+    */
+	function _hasNavBackFrames() {
+		return (jumpedPosStack.length > 0 && jumpToPosStack.length > 0)
+				|| (!jumpedPosStack.length && jumpToPosStack.length > 1);
+	}
      
    /**
     * Function to enable/disable navigation command based on cursor positions availability.
     * @private
     */
     function _validateNavigationCmds() {
-        commandJumpBack.setEnabled(jumpToPosStack.length > 0);
+        commandJumpBack.setEnabled(_hasNavBackFrames());
         commandJumpFwd.setEnabled(jumpedPosStack.length > 0);
     }
     
@@ -129,9 +140,20 @@ define(function (require, exports, module) {
     NavigationFrame.prototype._bindEditor = function (editor) {
         var self = this;
         editor.on("beforeDestroy", function () {
+			var updateCurrentPos;
+			
+			// Check if this frame is actually the current frame
+			// if true then we will update the current frame with serialized positions from CM
+			if (self === currentEditPos) {
+				updateCurrentPos = true;
+			}
             self._backupSelectionRanges();
             self.cm = null;
             self.bookMarkIds = null;
+			
+			if (updateCurrentPos) {
+				currentEditPos = self;
+			}
         });
     };
     
@@ -197,6 +219,23 @@ define(function (require, exports, module) {
             }
         }
     };
+	
+	/**
+    * Function to clean up the markers in cm
+    */
+    NavigationFrame.prototype._clearMarkers = function () {
+        if (!this.cm) {
+            return;
+        }
+        var self = this;
+		
+		// clear only the markers we used to mark selections/cursors
+        this.cm.getAllMarks().filter(function (entry) {
+            if (entry.className === self.uId || self.bookMarkIds.indexOf(entry.id) !== -1) {
+                entry.clear();
+            }
+        });
+    };
 
    /**
     * Function to actually navigate to the position(file,selections) captured in this frame
@@ -234,10 +273,19 @@ define(function (require, exports, module) {
             window.clearTimeout(captureTimer);
             captureTimer = null;
         }
+		
         // Ensure cursor activity has not happened because of arrow keys or edit
-        if (selectionObj.origin !== "+move" && (window.event && window.event.type !== "input")) {
+        if (selectionObj.origin !== "+move" && (!window.event || (window.event && window.event.type !== "input"))) {
             captureTimer = window.setTimeout(function () {
-                jumpToPosStack.push(new NavigationFrame(event.target, selectionObj));
+				// Check if we have reached MAX_NAV_FRAMES_COUNT
+				// If yes, control overflow
+				if (jumpToPosStack.length === MAX_NAV_FRAMES_COUNT) {
+					var navFrame = jumpToPosStack.splice(0, 1);
+					navFrame._clearMarkers();
+				}
+				
+				currentEditPos = new NavigationFrame(event.target, selectionObj);
+				jumpToPosStack.push(currentEditPos);
                 _validateNavigationCmds();
                 activePosNotSynced = false;
             }, NAV_FRAME_CAPTURE_LATENCY);
@@ -252,20 +300,33 @@ define(function (require, exports, module) {
     function _navigateBack() {
         if (!jumpedPosStack.length) {
             if (activePosNotSynced) {
-                jumpToPosStack.push(new NavigationFrame(EditorManager.getCurrentFullEditor(), {ranges: EditorManager.getCurrentFullEditor()._codeMirror.listSelections()}));
+				currentEditPos = new NavigationFrame(EditorManager.getCurrentFullEditor(), {ranges: EditorManager.getCurrentFullEditor()._codeMirror.listSelections()});
+                jumpedPosStack.push(currentEditPos);
             }
         }
+		
         var navFrame = jumpToPosStack.pop();
+		
+		// Check if the poped frame is the current active frame
+		// if true, jump again
+		if (navFrame === currentEditPos) {
+			jumpedPosStack.push(navFrame);
+			CommandManager.execute(NAVIGATION_JUMP_BACK);
+			return;
+		}
+		
         if (navFrame) {
             // We will check for the file existence now, if it doesn't exist we will jump back again
             // but discard the popped frame as invalid.
             _checkIfExist(navFrame).done(function () {
                 jumpedPosStack.push(navFrame);
                 navFrame.goTo();
+				currentEditPos = navFrame;
             }).fail(function () {
-                _validateNavigationCmds();
                 CommandManager.execute(NAVIGATION_JUMP_BACK);
-            });
+            }).always(function () {
+				_validateNavigationCmds();
+			});
         }
     }
     
@@ -274,16 +335,28 @@ define(function (require, exports, module) {
     */
     function _navigateFwd() {
         var navFrame = jumpedPosStack.pop();
+		
+		// Check if the poped frame is the current active frame
+		// if true, jump again
+		if (navFrame === currentEditPos) {
+			jumpToPosStack.push(navFrame);
+			CommandManager.execute(NAVIGATION_JUMP_FWD);
+			return;
+		}
+		
         if (navFrame) {
             // We will check for the file existence now, if it doesn't exist we will jump back again
             // but discard the popped frame as invalid.
             _checkIfExist(navFrame).done(function () {
                 jumpToPosStack.push(navFrame);
                 navFrame.goTo();
+				currentEditPos = navFrame;
             }).fail(function () {
                 _validateNavigationCmds();
                 CommandManager.execute(NAVIGATION_JUMP_FWD);
-            });
+            }).always(function () {
+				_validateNavigationCmds();
+			});
         }
     }
     
--- a/src/extensions/default/NavigationAndHistory/NavigationProvider.js
+++ b/src/extensions/default/NavigationAndHistory/NavigationProvider.js
@@ -58,14 +58,14 @@ define(function (require, exports, module) {
     * @private
     * @type {Array.<Object>}
     */
-    var jumpToPosStack = [];
+    var jumpBackwardStack = [];
     
    /**
     * Contains list of most recently traversed cursor positions using NAVIGATION_JUMP_BACK command.
     * @private
     * @type {Array.<Object>}
     */
-    var jumpedPosStack = [],
+    var jumpForwardStack = [],
         activePosNotSynced = false,
         captureTimer = null,
         currentEditPos = null,
@@ -78,8 +78,8 @@ define(function (require, exports, module) {
     * @private
     */
     function _hasNavBackFrames() {
-        return (jumpedPosStack.length > 0 && jumpToPosStack.length > 0)
-            || (!jumpedPosStack.length && jumpToPosStack.length > 1);
+        return (jumpForwardStack.length > 0 && jumpBackwardStack.length > 0)
+            || (!jumpForwardStack.length && jumpBackwardStack.length > 1);
     }
      
    /**
@@ -88,7 +88,7 @@ define(function (require, exports, module) {
     */
     function _validateNavigationCmds() {
         commandJumpBack.setEnabled(_hasNavBackFrames());
-        commandJumpFwd.setEnabled(jumpedPosStack.length > 0);
+        commandJumpFwd.setEnabled(jumpForwardStack.length > 0);
     }
     
    /**
@@ -97,10 +97,10 @@ define(function (require, exports, module) {
     */
     function _validateFrame(entry) {
         var deferred = new $.Deferred(),
-            fileEntry = FileSystem.getFileForPath(entry.file);
+            fileEntry = FileSystem.getFileForPath(entry.filePath);
 
         if (entry.inMem) {
-            var indexInWS = MainViewManager.findInWorkingSet(entry.paneId, entry.file);
+            var indexInWS = MainViewManager.findInWorkingSet(entry.paneId, entry.filePath);
             // Remove entry if InMemoryFile is not found in Working set
             if (indexInWS === -1) {
                 deferred.reject();
@@ -132,7 +132,7 @@ define(function (require, exports, module) {
     */
     function NavigationFrame(editor, selectionObj) {
         this.cm = editor._codeMirror;
-        this.file = editor.document.file._path;
+        this.filePath = editor.document.file._path;
         this.inMem = editor.document.file.constructor.name === "InMemoryFile";
         this.paneId = editor._paneId;
         this.fileStat = editor.document.file._stat;
@@ -163,23 +163,7 @@ define(function (require, exports, module) {
     NavigationFrame.prototype._reinstateMarkers = function (editor) {
         this.cm = editor._codeMirror;
         this.paneId = editor._paneId;
-        
-        var range,
-            index,
-            bookMark;
-
-        this.bookMarkIds = [];
-        for (index in this.selections) {
-            range = this.selections[index];
-            // 'markText' has to used for a non-zero length position, if current selection is 
-            // of zero length use bookmark instead.
-            if (range.start.line === range.end.line && range.start.ch === range.end.ch) {
-                bookMark = this.cm.setBookmark(range.start, range.end);
-                this.bookMarkIds.push(bookMark.id);
-            } else {
-                this.cm.markText(range.start, range.end, {className: (this.uId)});
-            }
-        }
+        this._createMarkers(this.selections);
     };
     
    /**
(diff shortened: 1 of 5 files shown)


src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,359 @@
...
+    /*
+    * Contains list of most recently traversed cursor positions using NAVIGATION_JUMP_BACK command.
+    * @private
+    * @type {Array.<Object>}
+    */
+    var jumpedPosStack = [],
+        captureTimer,
+        activePosNotSynched = false,
+        jumpInProgress,
+        command_JumpBack,
```
petetnt: The `camelcase` + `underscore` hybrid is a bit weird, would prefer just `commandJumpBack`.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,359 @@
...
+    var jumpToPosStack = [];
+    
+    /*
+    * Contains list of most recently traversed cursor positions using NAVIGATION_JUMP_BACK command.
+    * @private
+    * @type {Array.<Object>}
+    */
+    var jumpedPosStack = [],
+        captureTimer,
+        activePosNotSynched = false,
```
petetnt: Nit: values with defaults first

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,359 @@
...
+        command_JumpFwd,
+        cmMarkers = {};
+     
+   /**
+    * Function to enable/disable navigation command based on cursor positions availability.
+    * @private
+    */
+    function _validateNavigationCmds() {
+        if (jumpToPosStack.length > 0) {
+            command_JumpBack.setEnabled(true);
```
petetnt: Can we just do 
``` js
commandJumpBack.setEnabled(jumpToPosStack.length > 0);
commandJumpFwd.setEnabled(jumpedPosStack.length > 0);
```

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,359 @@
...
+    /**
+    * Function to check existence of a file entry
+    * @private
+    */
+    function _checkIfExist(entry) {
+        var deferred = new $.Deferred(),
+            fileEntry = FileSystem.getFileForPath(entry.file);
+
+        if (entry.inMem) {
+            var indxInWS = MainViewManager.findInWorkingSet(entry.paneId, entry.file);
```
petetnt: Nit: prefer `index` over `indx`

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,359 @@
...
+        } else {
+            command_JumpFwd.setEnabled(false);
+        }
+    }
+    
+    /**
+    * Function to check existence of a file entry
+    * @private
+    */
+    function _checkIfExist(entry) {
```
petetnt: We probably should null check for entry here

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,359 @@
...
+                deferred.resolve();
+            }
+        } else {
+            fileEntry.exists(function (err, exists) {
+                if (!err && exists) {
+                    deferred.resolve();
+                } else {
+                    deferred.reject();
+                }
+		    });
```
petetnt: Nit: indentation

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,359 @@
...
+        return deferred.promise();
+    }
+        
+   /**
+    * Prototype to capture a navigation frame and it's various data/functional attributues
+    */
+    function NavigationFrame(editor, selectionObj) {
+        this.cm = editor._codeMirror;
+        this.file = editor.document.file._path;
+        this.inMem = editor.document.file.constructor.name === "InMemoryFile" ? true : false;
```
petetnt: `this.inMem = editor.document.file.constructor.name === "InMemoryFile"`

is sufficent.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,359 @@
...
+        this.paneId = editor._paneId;
+        this.uId = (new Date()).getTime();
+        this.selections = [];
+        this.bookMarkIds = [];
+        this._createMarkers(selectionObj.ranges);
+        this._bindEditor(editor);
+    }
+    
+   /**
+    * Binds the lifecycle event listner of the editor for which this frame is captured
```
petetnt: Typo: `listener`

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,359 @@
...
+   /**
+    * Function to create CM TextMarkers for the navigated positions/selections.
+    * This logic is required to ensure that the captured navaigation positions 
+    * stay valid and contextual even when the actual document text mutates.
+    * The mutations which are handled here :
+    * -> Addition/Deletion of lines before the captured position
+    * -> Addition/Updation of characters in the captured selection
+    */
+    NavigationFrame.prototype._createMarkers = function (ranges) {
+        var range, index, bookMark;
```
petetnt: Nit: new variables on new lines

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,359 @@
...
+        this.bookMarkIds = [];
+        for (index in ranges) {
+            range = ranges[index];
+            // 'markText' has to used for a non-zero length position, if current selection is 
+            // of zero length use bookmark instead.
+            if (range.anchor.line === range.head.line && range.anchor.ch === range.head.ch) {
```
petetnt: Q: can `anchor` or `head` be null?

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,359 @@
...
+    * Function to actually convert the CM markers to CM positions which can be used to 
+    * set selections or cursor positions in Editor.
+    */
+    NavigationFrame.prototype._backupSelectionRanges = function () {
+        if (!this.cm) {
+            return;
+        }
+        
+        this.selections = [];
+        var marker, selection, index;
```
petetnt: Nit: new variables on new lines

src/extensions/default/NavigationAndHistory/NavigationProvider.js
petetnt: Nit: `activePosNotSynced`

petetnt (changes requested):
Some initial code review, good stuff @swmitra!

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: Done.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: Done.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: Done.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: It's already checked at the call location - 

```javascript
if (navFrame) {
            // We will check for the file existence now, if it doesn't exist we will jump back again
            // but discard the popped frame as invalid.
            _checkIfExist(navFrame).done(function () {
`

File: src/extensions/default/NavigationAndHistory/NavigationProvider.js (hunk 4 above)
Comment (swmitra) [inline src/extensions/default/NavigationAndHistory/NavigationProvider.js]: Done.

File: src/extensions/default/NavigationAndHistory/NavigationProvider.js (hunk 6 above)
Comment (swmitra) [inline src/extensions/default/NavigationAndHistory/NavigationProvider.js]: Done.

File: src/extensions/default/NavigationAndHistory/NavigationProvider.js (hunk 7 above)
Comment (swmitra) [inline src/extensions/default/NavigationAndHistory/NavigationProvider.js]: Done.

File: src/extensions/default/NavigationAndHistory/NavigationProvider.js (hunk 8 above)
Comment (swmitra) [inline src/extensions/default/NavigationAndHistory/NavigationProvider.js]: Done 😄

File: src/extensions/default/NavigationAndHistory/NavigationProvider.js (hunk 10 above)
Comment (swmitra) [inline src/extensions/default/NavigationAndHistory/NavigationProvider.js]: If there is a cursor or selection, it can never be null and if there are no selections/cursor then the ranges would be an empty array.

File: src/extensions/default/NavigationAndHistory/NavigationProvider.js (hunk 11 above)
Comment (swmitra) [inline src/extensions/default/NavigationAndHistory/NavigationProvider.js]: Done.

File: src/extensions/default/NavigationAndHistory/NavigationProvider.js (hunk 9 above)
Comment (swmitra) [inline src/extensions/default/NavigationAndHistory/NavigationProvider.js]: Done.

**Comment by swmitra** [comment]
@petetnt Thanks a lot for reviewing this PR. It was really quick 👍 
I have updated the code with relevant changes pertaining to review comments. All yours now 😄

File: src/extensions/default/NavigationAndHistory/NavigationProvider.js [hunk 12]
```diff
@@ -0,0 +1,360 @@
...
+    var KeyboardPrefs = JSON.parse(require("text!keyboard.json"));
+    
+    // Command constants for navigation history
+    var NAVIGATION_JUMP_BACK      = "navigation.jump.back",
+        NAVIGATION_JUMP_FWD       = "navigation.jump.fwd";
+    
+    // The latency time to capture an explicit cursor movement as a navigation frame
+    var NAV_FRAME_CAPTURE_LATENCY = 2000;
+    
+    /*
```
petetnt: Nit: missing * (/**)

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,360 @@
...
+    * Contains list of most recently known cursor positions.
+    * @private
+    * @type {Array.<Object>}
+    */
+    var jumpToPosStack = [];
+    
+    /*
```
petetnt: Nit: missing * (/**)

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,360 @@
...
+                deferred.resolve();
+            }
+        } else {
+            fileEntry.exists(function (err, exists) {
+                if (!err && exists) {
+                    deferred.resolve();
+                } else {
+                    deferred.reject();
+                }
+			});
```
petetnt: Nit: indentation: Tabs -> Spaces

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,360 @@
...
+    
+   /**
+    * Function to actually convert the CM markers to CM positions which can be used to 
+    * set selections or cursor positions in Editor.
+    */
+    NavigationFrame.prototype._backupSelectionRanges = function () {
+        if (!this.cm) {
+            return;
+        }
+
```
petetnt: Nit: L173-L177 indendation Tabs -> Spaces

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,360 @@
...
+        var self = this;
+		
+		// Collate only the markers we used to mark selections/cursors
+        var markers = this.cm.getAllMarks().filter(function (entry) {
+            if (entry.className === self.uId || self.bookMarkIds.indexOf(entry.id) !== -1) {
+                return entry;
+            }
+        });
+		
+		// Iterate over CM textmarkers and collate the updated(if?) positions
```
petetnt: Nit: Indentation: tabs -> spaces

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,360 @@
...
+		
+		var marker,
+			selection,
+			index;
+        
+		// Reset selections first.
+        this.selections = [];
+        var self = this;
+		
+		// Collate only the markers we used to mark selections/cursors
```
petetnt: Nit: Indentation: tabs -> spaces

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,360 @@
...
+    NavigationFrame.prototype._backupSelectionRanges = function () {
+        if (!this.cm) {
+            return;
+        }
+		
+		var marker,
+			selection,
+			index;
+        
+		// Reset selections first.
```
petetnt: Nit: Indentation: tabs -> spaces

petetnt (changes requested):
Few more nits, otherwise codewise starting to LGTM. Haven't had the time to test it yet.

ficristo:
I will not review, but I have a question.
If I understand correctly the navigation history is never emptyed but when you change project.
Doesn't it risk to consume too much memory? Usually this kind of things have a limit, like remembering only the last 100 navigation changes.

swmitra:
@ficristo You are right 👍 . 
I had now put a limit of 50 and control overflow beyond that as having infinite navigation frames won't help anyone. As a developer even to do 50 back navigation would be too much.

swmitra:
@petetnt Can you please have a look at the PR again and if possible use it. 
@MarcelGarber Can you please have a look at this PR?

swmitra:
@nethip Can you please have a look at this PR?

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,432 @@
...
+    * @type {Array.<Object>}
+    */
+    var jumpToPosStack = [];
+    
+   /**
+    * Contains list of most recently traversed cursor positions using NAVIGATION_JUMP_BACK command.
+    * @private
+    * @type {Array.<Object>}
+    */
+    var jumpedPosStack = [],
```
nethip: It is better to initialize all the variables. here.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,432 @@
...
+            }).always(function () {
+				_validateNavigationCmds();
+	    	});
+        }
+    }
+    
+   /**
+    * Command handler to navigate forward
+    */
+    function _navigateFwd() {
```
petetnt: Would prefer `_navigateForward()` here

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,432 @@
...
+			self.bookMarkIds = null;
+			if (updateCurrentPos) {
+				currentEditPos = self;
+			}
+        });
+    };
+    
+   /**
+    * Function to create CM TextMarkers for the navigated positions/selections.
+    * This logic is required to ensure that the captured navaigation positions
```
petetnt: Typo: navigation

petetnt (changes requested):
Looks mostly good to me codewise, one nit and one typo. Also there's now mixed tabs and spaces in new code. Haven't had the time to test the functionality yet but I'll try soon.

swmitra:
@petetnt Done. Fixed the last 2 nits and mixed tabs & spaces.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: Done 👍

swmitra:
@nethip Can you please have a look at this PR again?

nethip:
@swmitra I think the code part is fine. But here are few questions I had

* What will happen if the file, for which a `NavigationFrame` is present in your `JumpToStack` or `JumpedPosStack`, is changed externally  ?

* I am trying to understand why you have to use `beforeSelection` instead of selection changed? Had it been capturing after selection change, you won't have to rely on `activePosNotSynced` variable.

* I see that you have used `currentFullEditor` for switching between different files to restore a frame state. Does that mean selections captured for editors inside `Inline` editors will transform to full view when `back` or `front` command is fired?

* I see that you are relying on `cm.setBookMark`. What happens when any extension overrides the selection marks set, while selection being changed?

* Also for batch operations like `Replace All`, are you going to log all of the selection changes?

swmitra:
@nethip Thanks a ton for reviewing the PR 👍 

I will try to answer the questions, please let me know if you think anything in the assumptions can go wrong.

* If we categorize external changes, we can see there are broadly 2 different kind of mutations. First one is the file entry itself being changed (delete/rename) and secondly the content being changed. In first case when we actually try to navigate using the stale frame, we end up discarding the frame as the entry doesn't exist on the disk. In second case, Brackets calls _resetText() on the existing editor with modified content which removes all the CM markers. So we gracefully discard the frame.(Let me verify it once again).

* I have used CM beforeSelection to capture the selection change "origin". This property is not part of selection change event as that's a simulated event by Brackets. We need to check for the origin of the change to discard frame capture in case user has used keyboard navigation keys or cursor change has happened because of character input. We just want to capture programmatic selection change (jump to def) or explicit pointer click. 

* This part was getting really tricky to implement as showing the selection again in InlineEditor would be impossible for two reasons - First and the most prominent one is availability of "createInlineEditor" as a public API. I have seen multiple usage of this API even by extensions and then might show this in a different DOM container all together. That's the reason I listen only for full editors ( check for _paneId). So any navigation in inline editors are not captured at all.  

* Once a textmark is created in CM, it can't be overridden. It can be removed though. In case it is removed , we won't be able to find it ("find()" will return empty position) and hence we will not act on it. 

* Correct me if I am wrong, to highlight the matches in search/replace operation, we use CM text marker instead of actual selection and to actually replace the searched token we use replaceRange. So we don't use selection in these operations. 

Please let me know if you think we can go wrong in any of these or other situations. Thanks again for raising this questions, I could verify lot of the scenarios after looking at this 😄

swmitra:
@nethip I have added additional handling for external changes. Also, deciding on which pane the document should open on navigation is determined based on the current master editors association with pane. 

Please have a look at the additional changes.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,553 @@
...
+        }
+    }
+    
+   /**
+    * Command handler to navigate forward
+    */
+    function _navigateForward() {
+        var navFrame = jumpedPosStack.pop();
+        
+        // Check if the poped frame is the current active frame or doesn't have any valid marker information
```
MarcelGerber: We should have an early `return` here for the case when navFrame is null (`!navFrame`)

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,553 @@
...
+        commandJumpBack,
+        commandJumpFwd;
+        
+   /**
+    * Function to check if there are any navigatable frame backward.
+    * @private
+    */
+    function _hasNavBackFrames() {
+        return (jumpedPosStack.length > 0 && jumpToPosStack.length > 0)
+            || (!jumpedPosStack.length && jumpToPosStack.length > 1);
```
MarcelGerber: I don't quite get this code.
So, when you can't jump forward, there need to be at least two elements in `jumpToPosStack`? Why is that?

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,553 @@
...
+    
+   /**
+    * Function to create CM TextMarkers for the navigated positions/selections.
+    * This logic is required to ensure that the captured navigation positions 
+    * stay valid and contextual even when the actual document text mutates.
+    * The mutations which are handled here :
+    * -> Addition/Deletion of lines before the captured position
+    * -> Addition/Updation of characters in the captured selection
+    */
+    NavigationFrame.prototype._createMarkers = function (ranges) {
```
MarcelGerber: createMarkers and reinstateMarkers share a lot of code. Maybe we can consolidate them a little more?

As I see it, this should work:
``` js
NavigationFrame.prototype._reinstateMarkers = function (editor) {
    this.cm = editor._codeMirror;
    this.paneId = editor._paneId;
    this._createMarkers(this.selections)
}
```

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,553 @@
...
+        }
+        // Reset forward navigation stack if we are capturing a new event
+        jumpedPosStack = [];
+        if (captureTimer) {
+            window.clearTimeout(captureTimer);
+            captureTimer = null;
+        }
+        
+        // Ensure cursor activity has not happened because of arrow keys or edit
+        if (selectionObj.origin !== "+move" && (!window.event || (window.event && window.event.type !== "input"))) {
```
MarcelGerber: You can simplify the logic here: `!window.event || window.event.type !== "input"`

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,553 @@
...
+            captureTimer = window.setTimeout(function () {
+                // Check if we have reached MAX_NAV_FRAMES_COUNT
+                // If yes, control overflow
+                if (jumpToPosStack.length === MAX_NAV_FRAMES_COUNT) {
+                    var navFrame = jumpToPosStack.splice(0, 1)[0];
```
MarcelGerber: You can use `.shift()` here.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,553 @@
...
+    }
+    
+   /**
+    * Command handler to navigate backward
+    */
+    function _navigateBack() {
+        if (!jumpedPosStack.length) {
+            if (activePosNotSynced) {
+                currentEditPos = new NavigationFrame(EditorManager.getCurrentFullEditor(), {ranges: EditorManager.getCurrentFullEditor()._codeMirror.listSelections()});
+                jumpedPosStack.push(currentEditPos);
```
MarcelGerber: Suggestion:
Make the anonymous function inside `setTimeout` a "real" function and use it here, too.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,553 @@
...
+    // The latency time to capture an explicit cursor movement as a navigation frame
+    var NAV_FRAME_CAPTURE_LATENCY = 2000,
+        MAX_NAV_FRAMES_COUNT = 30;
+    
+   /**
+    * Contains list of most recently known cursor positions.
+    * @private
+    * @type {Array.<Object>}
+    */
+    var jumpToPosStack = [];
```
MarcelGerber: To make naming clearer, you might also use `jumpStackForward` and `jumpStackBackward`

src/extensions/default/NavigationAndHistory/NavigationProvider.js
```diff
@@ -0,0 +1,553 @@
...
+
+        return deferred.promise();
+    }
+        
+   /**
+    * Prototype to capture a navigation frame and it's various data/functional attributues
+    */
+    function NavigationFrame(editor, selectionObj) {
+        this.cm = editor._codeMirror;
+        this.file = editor.document.file._path;
```
MarcelGerber: Might call this `filepath` or just `path` to make it clear that this is _not_ a file handle.

marcelgerber (changes requested):
Didn't really get to actually testing it yet.
Just one note, and I hope you are aware of it: CodeMirror already has a kind of cursor/selection history using `Ctrl-U / Ctrl-Shift-U` (Windows), so maybe we can utilize that a little? Maybe they already take care of all the markers internally?

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: Done!

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: Done. "filePath" would be more logical.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: The problem was with range start and end placeholder names but handled it now to make use of same function.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: Refactored this code earlier but missed this condition completely. Changed it now.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: Changed.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: I had to use the anonymous function to create closure in the call back scope.

src/extensions/default/NavigationAndHistory/NavigationProvider.js
swmitra: Done.

swmitra:
@MarcelGerber Thanks a lot for reviewing this PR 👍 
I have tried CM's implicit way to handle navigation but the problem was to handle navigation between multiple instances of CM which is the case we want to support predominantly. Then if we add handling of external changes and re initiating the edit session of the same file in a different CM instance, maintaining the marker cache inside CM was impossible. I had to externalize the whole thing to handle such scenarios. 
Do let me know if you think it can be handled in a better way by supporting all the features what we are going to have with this implementation. 

I have addressed most of the review comments. Please have a look at this PR and use it if possible.

marcelgerber (approved):

saurabh95 (approved):

schroef:
Just downloaded the revision, but alt+i shows me this character ˆ??

petetnt:
Hi @schroef,

the OSX shortcuts were improved / fixed in https://github.com/adobe/brackets/pull/13527, which didn't land to 1.10. You can change your shortcuts to match the ones in 13527 from Debug -> Show User Key Map and adding the following:

```
{
    "documentation": "https://github.com/adobe/brackets/wiki/User-Key-Bindings",
    "overrides": {
        "Cmd-Alt-Left": "navigation.jump.back",
        "Cmd-Alt-Right": "navigation.jump.fwd".
    }
}
```

schroef:
Sorry but that doesnt do anything as well. Looking at that code, shouldn't that also be visible then in the navigation menu, if so it's not there in 1.10

screen shot 2017-07-08 at 9 39 11 pm

schroef:
Seems shortcuts work not normally like in other applications. I need to pres cmd+alt+ left or right in a sequence in order to let it do something. I seem to be able to jump through documents. It doesnt always jump back to the complete history of it locations though

swmitra:
@schroef Brackets have a slightly different implementation for navigation. We don't remember all the positions rather only explicit positions. For example when you do mouse click at a position and pause for a brief moment to read something, do a jump to def or switch file to work on some other context.

swmitra:
We are going to change the OSX shorcuts and provide a small UI place holder as well in the next pre release ( within a week's time!)

schroef:
Well it doesnt store all the position i think. I've tried it couple times and sometimes is seems to work and sometimes not. I definitely dont see that history of 30 states.

utkarshmehta (commented):
a

Issue #13381, opened 2017-05-18
File extension is outputted twice when saving a file created using DocumentManager.createUntitledDocument.


Grafluxe:
### Prerequisites

* [ ] Can you reproduce the problem with `Debug -> Reload Without Extensions`?
* [x] Did you perform a cursory search to see if your bug or enhancement is already reported?
* [x] Did you read the [Troubleshooting guide](https://github.com/adobe/brackets/wiki/Troubleshooting)?

### Description

Brackets v1.9.0 added support for creating untitled files with extensions (along with the ability to change language types in untitled files). See [Brackets 13086](https://github.com/adobe/brackets/pull/13086). When using `DocumentManager.createUntitledDocument(count, ext);`, everything works as expected until you save. The bug is that when you save an untitled file with an extension, the filename outputted into the save-as dialog treats the untitled documents extension as part of its filename and appends the extension again. In other words, the save-as dialogs filename has the extension twice.

### Steps to Reproduce

1. Create an untitled document with an extension using the DocumentManager.
    - To expedite this step, you can use my plugin [AnyTemplate](https://github.com/Grafluxe/any-template) (which creates an untitled file with an extension).
2. Save the file.

**Expected behavior:** The untitled files name should not include the extension twice.

**Actual behavior:** The untitled files name includes the extension twice when saving. It outputs like this: `Untitled-1.js.js`.

### Versions

- OS: macOS Sierra 10.12 (16A323)
- Brackets: Release 1.9 build 1.9.0-17312 (release 189f6d39a) 

This issue occurs on both Mac and Windows.

petetnt:
Hi @Grafluxe,

thanks for the detailed report! Would you be interested in fixing it in the Brackets core?

Grafluxe:
You're welcome @petetnt. I wouldn't mind fixing it. Anything I need to know that I can't find on the dev page?

I'm a **huge fan of Brackets** and use it daily.

petetnt:
Cool!

See the article [How to hack on Brackets](https://github.com/adobe/brackets/wiki/How-to-Hack-on-Brackets) to get started with the Brackets codebase which should cover all the basics. If you have any questions, feel free to post them here! We also have a [Slack channel](https://brackets.slack.com) for day-to-day communication. (You can join by sending a mail to admin@brackets.io with the subject line `slack registration request` specifying the email addresses you would like to register).

After you have a fix, send in a PR which we will then review and merge after any requested changes 🙌 !

[...]

PR #13426, opened 2017-06-06
Fix issue where file extension is outputted twice when saving…

grafluxe:
…a file created using DocumentManager.createUntitledDocument

Brackets v1.9.0 added support for creating untitled files with extensions (along with the ability to change language types in untitled files). See Brackets 13086. When using DocumentManager.createUntitledDocument(count, ext);, everything works as expected until you save. The bug is that when you save an untitled file with an extension, the filename outputted into the save-as dialog treats the untitled documents extension as part of its filename and appends the extension again. In other words, the save-as dialogs filename has the extension twice.

See: https://github.com/adobe/brackets/issues/13381

### DIFF
--- a/src/document/DocumentCommandHandlers.js
+++ b/src/document/DocumentCommandHandlers.js
@@ -955,8 +955,10 @@ define(function (require, exports, module) {
             defaultName = FileUtils.getBaseName(origPath);
             var file = FileSystem.getFileForPath(origPath);
             if (file instanceof InMemoryFile) {
-                var language = LanguageManager.getLanguageForPath(origPath);
-                if (language) {
+                var language = LanguageManager.getLanguageForPath(origPath),
+                    hasExtension = (FileUtils.getFileExtension(origPath).length > 0);
+
+                if (language && !hasExtension) {
                     var fileExtensions = language.getFileExtensions();
                     if (fileExtensions && fileExtensions.length > 0) {
                         defaultName += "." + fileExtensions[0];


petetnt (commented):
Changeset LGTM, a unit test would be nice.

petetnt:
Hi @Grafluxe,

thanks for this great PR! The changes LGTM, but a unit test would be ace.

In https://github.com/adobe/brackets/blob/master/test/spec/DocumentCommandHandlers-test.js there's a boatload of tests regarding DocumentCommandHandlers, which can be ran from Debug -> Run Tests -> Integration -> DocumentCommandHandlers. A tests that would be something like `it("should not append another extension if its already defined")` would be nice to cover this case not caught before by the unit tests.

If you haven't worked with unit tests on Brackets before, feel free to ask any questions and we will be more than happy to point you to the right direction! 👍

Grafluxe:
Hi @petetnt,

The reason I didn't send a PR sooner, is because I spent much time trying to add a new integration test. Trust me, I really wanted to include it in my commit.

I could not figure out a way to test this functionality. The logic that I would need to test is nested inside a private member (_doSaveAs) which was not reachable/testable in my attempts. If there is a way to test it, please let me know and I'll be happy to add.

petetnt:
@Grafluxe no problem, I bet we can find a way together! I have to think that about a bit but I'll get back to you ASAP. Or we can merge this and create an issue for adding the test too if the solution is something non-trivial

Grafluxe:
Hey @petetnt, was this branch ever merged into master?

PR #13430, opened 2017-06-11
Problems panel: add an icon per problem to tell which type it is

ficristo:
How it will look:
lint-icons


swmitra (approved):
Good work @ficristo 👍

ficristo:
@swmitra thank you.

PR #13447, opened 2017-06-15
Changed WebPlatform License to MDN License and respective strings

saurabh95:
We still need to get translations for some languages.

### DIFF
--- a/src/htmlContent/about-dialog.html
+++ b/src/htmlContent/about-dialog.html
@@ -14,7 +14,7 @@
                 </p>
 
                 <p class="dialog-message"><!-- $NON-NLS$ -->Copyright 2012 - present Adobe Systems Incorporated and its licensors. All rights reserved.</p>
-                <p class="dialog-message">{{{Strings.ABOUT_TEXT_WEB_PLATFORM_DOCS}}}</p>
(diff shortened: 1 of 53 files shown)


ficristo:
I skimmed through the changes and they seems good.
But I noticed the following things (at the left it is the locale):
- gl: missing urls.js file
- hu: missing urls.js file
- lv: missing urls.js file
- pt-pr: missing entry
- ro: missing urls.js file
- sk: missing urls.js file
- sr: missing urls.js file
- tr: missing entry ABOUT_TEXT_MDN_DOCS (old ABOUT_TEXT_WEB_PLATFORM_DOCS) in strings.js
- uk: missing urls.js file

saurabh95:
The above mentioned locales with no urls.js there is no getting started page defined in sample, so probably this might be the reason of not creating the urls.js

ficristo:
Yeah, what I was trying to say, if you need the text translated everywhere those languages will not be.

saurabh95:
@ficristo I just did replace of Web Platform to MDN, there are some languages which got missed out, since I am not an expert in those languages so I kept them as it is. Once this is merged we can ask Adobe Localization team to verify translations for French and Japanese and the rest of the languages can be verified/corrected by community over time.

swmitra (approved):

PR #13476, opened 2017-06-23
More Robust UUID Handling

nethip:
More efficient UUID Handling: We need to backing up existing uuid so that we can accurately get the new user count, as the uuid is changed now and will effect the existing users as well. Also added robust checking for shell API availability before calling the shell API and if not present handling olderUUID in a different way.

@swmitra Please review this PR.

### DIFF
--- a/src/extensions/default/HealthData/HealthDataManager.js
+++ b/src/extensions/default/HealthData/HealthDataManager.js
@@ -70,31 +70,61 @@ define(function (require, exports, module) {
                         oneTimeHealthData.bracketsTheme = bracketsTheme;
                     })
                     .always(function () {
-                        var userUuid = PreferencesManager.getViewState("UUID");
+                        var userUuid  = PreferencesManager.getViewState("UUID");
+                        var olderUuid = PreferencesManager.getViewState("OlderUUID");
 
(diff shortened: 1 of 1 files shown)


src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -70,31 +70,61 @@ define(function (require, exports, module) {
...
+                                    oneTimeHealthData.uuid      = userUuid;
+                                    oneTimeHealthData.olderuuid = olderUuid;
+                                    return result.resolve(oneTimeHealthData);
+                                });
+                            } else {
+                                // Probably running on older shell, in which case we will
+                                // assign the same uuid to olderuuid.
+                                if (!userUuid) {
+                                    oneTimeHealthData.uuid = oneTimeHealthData.olderuuid = uuid.v4();
+                                    PreferencesManager.setViewState("UUID", oneTimeHealthData.uuid);
```
swmitra: Nit: May be we can write this outside the if else branch along with the other setViewState call @line:123.

swmitra (approved):
Good idea @nethip 👍 
Approved with a small nit.

PR #13483, opened 2017-06-26
Fix lgtm alerts

xiemaisi:
This fixes a few minor issues found by [lgtm](https://lgtm.com); for more details see https://lgtm.com/projects/g/adobe/brackets/alerts/.

~~The fixes in the first three commits are fairly obvious, the last two commits involve some educated guessing as to the intended semantics of the code involved.~~

### DIFF
--- a/src/extensions/default/JavaScriptCodeHints/HintUtils.js
+++ b/src/extensions/default/JavaScriptCodeHints/HintUtils.js
@@ -145,7 +145,7 @@ define(function (require, exports, module) {
             t.kind = kind;
             t.origin = "ecmascript";
             if (kind === "string") {
-                if (/[\\\\]*[^\\]"/.test(t.value)) {
+                if (/[^\\]"/.test(t.value)) {
                     t.delimiter = SINGLE_QUOTE;
                 } else {
                     t.delimiter = DOUBLE_QUOTE;
--- a/src/LiveDevelopment/MultiBrowserImpl/language/HTMLSimpleDOM.js
+++ b/src/LiveDevelopment/MultiBrowserImpl/language/HTMLSimpleDOM.js
@@ -266,7 +266,7 @@ define(function (require, exports, module) {
      */
     Builder.prototype.build = function (strict, markCache) {
         var self = this;
-        var token, lastClosedTag, lastTextNode, lastIndex = 0;
+        var token, lastClosedTag, lastTextNode;
         var stack = this.stack;
         var attributeName = null;
         var nodeMap = {};
@@ -441,7 +441,6 @@ define(function (require, exports, module) {
                     newNode.update();
                 }
             }
-            lastIndex = token.end;
         }
 
         // If we have any tags hanging open (e.g. html or body), fail the parse if we're in strict mode,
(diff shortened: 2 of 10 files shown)


src/extensibility/Package.js
```diff
@@ -265,7 +265,7 @@ define(function (require, exports, module) {
             // Download the bits (using Node since brackets-shell doesn't support binary file IO)
             var r = extensionManager.downloadFile(downloadId, urlInfo.url, PreferencesManager.get("proxy"));
             r.done(function (result) {
-                d.resolve({ localPath: FileUtils.convertWindowsPathToUnixPath(result), filenameHint: urlInfo.filenameHint });
+                d.resolve({ localPath: FileUtils.convertWindowsPathToUnixPath(result), filenameHint: filename });
```
petetnt: This makes sense but I am not 100% sure either, @swmitra?

petetnt (commented):
Cool stuff @xiemaisi, LGTM but that last one is bit tricky so someone else should check it too.

xiemaisi:
I've removed the last two commits with the less obvious fixes; I'll put them up as a separate PR.

xiemaisi:
@petent, does this look safer now?

xiemaisi:
@petetnt, does this look safer now?

src/extensions/default/JavaScriptCodeHints/HintUtils.js
```diff
@@ -145,7 +145,7 @@ define(function (require, exports, module) {
             t.kind = kind;
             t.origin = "ecmascript";
             if (kind === "string") {
-                if (/[\\\\]*[^\\]"/.test(t.value)) {
+                if (/[^\\]"/.test(t.value)) {
```
petetnt: This looks bit scary to me, but didn't manage to find any differences after running some tests cases by hand 🤔

petetnt (approved):
LGTM with a comment, someone else should do a double check too. @saurabh95 could you do that 🙏

src/extensions/default/JavaScriptCodeHints/HintUtils.js
saurabh95: This one looks scary to me as well. Will do some testing around it.

xiemaisi: It's not as scary as you might think: the original code looks for a substring of `t.value` that contains zero or more backslashes, followed by something that isn't a backslash, followed by a double quote. This is the same as simply looking for a substring consisting of something that isn't a backslash followed by a double quote, which is what the new code does.

(Note that lgtm actually only highlighted the fact that `[\\\\]` is redundant and means the same as `[\\]` or indeed `\\`.)

saurabh95: Oh ok. seems good to me. We can merge this after 1.10 release

petetnt: Good stuff @xiemaisi, merging this in and will land in 1.11. Hopefully we get the other fixes found by `lgtm` in too.

PR #13492, opened 2017-06-28
Fix NavigationProvider throwing errors when doc.file is missing

### LINKED ISSUE #13491: NavigationProvider - _removeXFramesForFile missing a null check
```
  /**
    * Removes all frames from backward navigation stack for the given file only if the file is changed on disk.
    * @private
    */
    function _removeBackwardFramesForFile(file) {
        jumpBackwardStack = jumpBackwardStack.filter(function (frame) {
            return frame.filePath !== file._path && frame.stat !== file._stat;
        });
    }
    
    /**
    * Removes all frames from forward navigation stack for the given file only if the file is changed on disk.
    * @private
    */
    function _removeForwardFramesForFile(file) {
        jumpForwardStack = jumpForwardStack.filter(function (frame) {
            return frame.filePath !== file._path && frame.stat !== file._stat;
        });
    }
```

For some reason `file` can be null (for example when switching projects), raising errors in the stack filter calls.

@swmitra

petetnt:
Fixes #13491

~Need to investigate _why_ `doc.file` can be missing though.~

Signed-off-by: petetnt <pete.a.nykanen@gmail.com>

### DIFF
--- a/src/extensions/default/NavigationAndHistory/NavigationProvider.js
+++ b/src/extensions/default/NavigationAndHistory/NavigationProvider.js
@@ -535,7 +535,9 @@ define(function (require, exports, module) {
             _reinstateMarkers(editor, jumpForwardStack);
         });
         FileSystem.on("change", function (event, entry) {
-            _handleExternalChange(event, {file: entry});
+            if (entry) {
+                _handleExternalChange(event, {file: entry});
+            }
         });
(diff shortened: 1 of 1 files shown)


swmitra (approved):

swmitra:
Thanks a lot @petetnt for reporting and fixing this scenario 👍

PR #13514, opened 2017-07-04
Fix import url  and @rule hints in inline style blocks

swmitra:
@saurabh95 Please review and test the failing unit test.


src/language/CSSUtils.js
```diff
@@ -649,6 +649,10 @@ define(function (require, exports, module) {
         if (_isInPropValue(ctx)) {
             return _getRuleInfoStartingFromPropValue(ctx, ctx.editor);
         }
+
```
swmitra: This was accidentally removed by this https://github.com/adobe/brackets/commit/73430b6c467a718f5710fe8d68d3b0d510fb4a24#diff-205888d21d20aa2319d86062c2e86080L624

PR #13517, opened 2017-07-04
If user cancels save as dialog after changing the encoding of dirty f…

saurabh95:
…ile, then encoding should not be changed in the file data structure


swmitra:
Late catch but still better we fixed this :+1

swmitra (approved):

PR #13545, opened 2017-07-10
Fix for external change handling

swmitra:
Improved external change handling. Eventually frames were being discarded while re-opening a file, in order fix that, we had to change the faulty condition check. 

@petetnt @saurabh95 Please review.


petetnt (commented):
Code changes LGTM, can you fill out some reproduction steps so I can check that the issue is fixed 👍

swmitra:
@petetnt navigate in a file for a couple of times. That would create few navigation frames. Now when you close that file, try to navigate back to the previous position. Without this change set, navigation provider was clearing up all the frames for the reopened file although there were no external changes.

petetnt (approved):

saurabh95 (approved):

PR #13560, opened 2017-07-13
Fix CTRL+Space handling while the CodeHintList is open, fixes #13481

### LINKED ISSUE #13481: CodeHintManager: Pressing "Ctrl+Space" causes TypeError each time
### Description
I was working on hint extension, which uses CodeHintManager API. During tests I noticed, that each time when I'm pressing Ctrl+Space, TypeError appears in brackets console: 

Full message: 
>TypeError: Cannot set property type of #<Event> which has only a getter TypeError: Cannot set property type of #<Event> which has only a getter.

I tested it in different ways (by runs Brackets without extensions for example) and results was always the same. In fact problem concerns CodeHintList's method named **`callMoveUp`** (line 520). This method tries to change read only property of KeyboardEvent instance as message shows. 

Decided to do extra test in two ways:
- First, I tried to comment this fragment of code and see the results. After running unit tests I made manual tests and nothing wrong happened.
- Fixed this code by cloning instance to new fake event object and noticed strange behavior. When I pressed *Ctrl+Space* keys, hint list was shown, but selection jumped to second element and when I pressed it again selection moved to third element. 
Screenshot below:
codehintmanager

I think, that this part of code could be removed, but I don't know what was the finally purpose of this functionality.

### Steps to Reproduce

1. Open any JS, CSS, SCSS file and debug console
2. Set cursor near the CSS property (for example)
3. Press **`Ctrl+Space`**

**Expected behavior:** Show code hint list without any errors

**Actual behavior:** Show code hint list with TypeError

### Versions
Brackets 1.9 and 1.10 pre-release 2 (Windows 7)

petetnt:
This PR fixes regression in filed as #13481 by instead of relying on
modifying original event we add a boolean flag telling `_keydownHook` the
event is actually faked and should be treated as such.

Signed-off-by: Pete Nykanen <pete.a.nykanen@gmail.com>

### DIFF
--- a/src/editor/CodeHintList.js
+++ b/src/editor/CodeHintList.js
@@ -326,9 +326,9 @@ define(function (require, exports, module) {
      * Convert keydown events into hint list navigation actions.
      *
      * @param {KeyBoardEvent} keyEvent
-     * @param {bool} isFakeCallUp - True if faked call up (for example calling CTRL+Space while hints are open)
+     * @param {bool} isFakeKeydown - True if faked key down call (for example calling CTRL+Space while hints are open)
      */
-    CodeHintList.prototype._keydownHook = function (event, isFakeCallUp) {
+    CodeHintList.prototype._keydownHook = function (event, isFakeKeydown) {
         var keyCode,
             self = this;
 
@@ -390,13 +390,13 @@ define(function (require, exports, module) {
         }
 
         // (page) up, (page) down, enter and tab key are handled by the list
-        if (event.type === "keydown" && this.isHandlingKeyCode(event) || isFakeCallUp) {
+        if ((event.type === "keydown" || isFakeKeydown) && this.isHandlingKeyCode(event)) {
             keyCode = event.keyCode;
 
             if (event.keyCode === KeyEvent.DOM_VK_ESCAPE) {
                 event.stopImmediatePropagation();
                 this.handleClose();
-
+                
                 return false;
             } else if (event.shiftKey &&
                     (event.keyCode === KeyEvent.DOM_VK_UP ||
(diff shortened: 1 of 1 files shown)


src/editor/CodeHintList.js
```diff
@@ -326,8 +326,9 @@ define(function (require, exports, module) {
      * Convert keydown events into hint list navigation actions.
      *
      * @param {KeyBoardEvent} keyEvent
+     * @param {bool} isFakeCallUp - True if faked call up (for example calling CTRL+Space while hints are open)
      */
-    CodeHintList.prototype._keydownHook = function (event) {
+    CodeHintList.prototype._keydownHook = function (event, isFakeCallUp) {
```
swmitra: NIT: `isFakeKeyDown` might me more suitable in this case?

src/editor/CodeHintList.js
```diff
@@ -389,7 +390,7 @@ define(function (require, exports, module) {
         }
 
         // (page) up, (page) down, enter and tab key are handled by the list
-        if (event.type === "keydown" && this.isHandlingKeyCode(event)) {
+        if (event.type === "keydown" && this.isHandlingKeyCode(event) || isFakeCallUp) {
```
swmitra: Could be combined like this to make our intention more explicit ?
``` javascript
 if ((event.type === "keydown" || isFakeKeyDown) && this.isHandlingKeyCode(event))

File: src/editor/CodeHintList.js [hunk 3]
```diff
@@ -408,7 +409,7 @@ define(function (require, exports, module) {
             } else if (keyCode === KeyEvent.DOM_VK_UP) {
                 _rotateSelection.call(this, -1);
             } else if (keyCode === KeyEvent.DOM_VK_DOWN ||
-                    (event.ctrlKey && keyCode === KeyEvent.DOM_VK_SPACE)) {
+                    (event.ctrlKey && keyCode === KeyEvent.DOM_VK_SPACE) || isFakeCallUp) {
```
swmitra: Do we need to touch this condition? In case of a fake key down for Ctrl+Space `(event.ctrlKey && keyCode === KeyEvent.DOM_VK_SPACE)` will return true anyways.

swmitra (approved):
Looks good to me barring the NITs and relevancy of the inner condition check 👍

petetnt:
Addressed comments in 8bd7adb, thanks @swmitra

swmitra (approved):

swmitra:
@petetnt Good work. This was a nagging issue and used to clutter the console completely with errors. 
Thanks a lot for fixing this long pending problem 👍

PR #13574, opened 2017-07-21
FIX: Creating a resizable panel using an HTML code fragment starting …

khatastroffik:
…with a comment will break

As mentioned in #6252, the original error msg isn't really helpful. 
The modified msg could be improved as well, though:
Since the content of the panel may be a **HTML fragment** (usually starting with a DIV node), mentioning that ```the first (root) DOM node must have an ID that will be used as a pref key``` would help.

Nevertheless, the mangled code in WorkspaceManager.js isn't reliable since it will _break_ when the panel content **starts with a html comment**. Such a comment may be necessary when  a plugin like HTMLhint is used. On a "panel" code fragment, this plugin attests an error due to a missing doctype. To disable this error, one may put a comment at the top of the html code: ```<!--htmlhint doctype-first: false-->```. WorkspaceManager.js will fail creating the panel, then. The proposed modification fixes such an issue.

**sample test code for a resizable panel**
```
<!--htmlhint doctype-first: false-->
<!--This comment is of no use but for testing purpose-->
<div id="my-very-special-testing-panel"><span style="color: red">Elvis</span> was here!</div>
```

### DIFF
--- a/src/command/KeyBindingManager.js
+++ b/src/command/KeyBindingManager.js
@@ -1455,6 +1455,7 @@ define(function (require, exports, module) {
     exports.getKeyBindings = getKeyBindings;
     exports.addGlobalKeydownHook = addGlobalKeydownHook;
     exports.removeGlobalKeydownHook = removeGlobalKeydownHook;
+    exports.normalizeKeyDescriptorString = normalizeKeyDescriptorString;
 
     /**
      * Use windows-specific bindings if no other are found (e.g. Linux).
--- a/src/view/WorkspaceManager.js
+++ b/src/view/WorkspaceManager.js
@@ -166,6 +166,8 @@ define(function (require, exports, module) {
      */
     function Panel($panel, minSize) {
         this.$panel = $panel;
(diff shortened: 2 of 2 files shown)


petetnt (approved):
Hi @lbegue! Thanks for this contribution!

LGTM, maybe a inline comment on the change would be nice as it's not immediately obvious why the root node might not be what you expect it to be 👍

lbegue:
my pleasure :-)

ficristo:
This change seems to break my extension when I add a comment at the start of the panel's html.
The following change seem to fix it. @lbegue could you double check?
```js
// Since the panel source code may start with a html comment, 
// the 'root' DOM node must be targeted explicitly using a filter removing all comments.
this.$panel = $panel.filter("*");
var $rootNodeWithID = this.panel[0];
```

PS: In future I would prefer if you don't misk unrelated things in the same PR. So the normalizedKey thing should have been in another PR, but not a big deal in this case.

RobertDowny92:
http://support.apple.com/kb/HT5883

PR #13589, opened 2017-08-01
Upgraded npm version to 3.10.10

saurabh95:


ficristo (approved):
I thought we were much more behind...
But it doesn't hurt.

petetnt (approved):
Changeset is okay, but any reason why we are not on `npm@4`?

saurabh95:
I have used the npm version which is shipped with node 6.11.0

ficristo:
We can upgrade to npm 4 in another PR. Merging now.

PR #13608, opened 2017-08-11
Remove react and add preact

boopeshmahendran:


src/main.js
```diff
@@ -29,16 +29,21 @@ require.config({
     paths: {
         "text"              : "thirdparty/text/text",
         "i18n"              : "thirdparty/i18n/i18n",
-        "react"             : "thirdparty/react",
 
         // The file system implementation. Change this value to use different
         // implementations (e.g. cloud-based storage).
         "fileSystemImpl"    : "filesystem/impls/appshell/AppshellFileSystem"
     },
+    packages: [
```
petetnt: We should probably add `preact` as an `node_module` instead of inlining it to the `thirdparty` folder as we have been doing with new dependencies.  (related to #12006)

src/project/FileTreeView.js
```diff
@@ -31,8 +31,8 @@
 define(function (require, exports, module) {
     "use strict";
 
-    var React             = require("thirdparty/react"),
-        ReactDOM          = require("thirdparty/react-dom"),
+    var React             = require("thirdparty/preact"),
```
petetnt: We probably should just `find-and-replace` all the React and ReactDOM mentions to `Preact`

test/spec/FileTreeView-test.js
```diff
@@ -29,12 +29,31 @@ define(function (require, exports, module) {
...
+        React             = require("thirdparty/preact"),
+        ReactDOM          = require("thirdparty/preact"),
         Immutable         = require("thirdparty/immutable"),
-        RTU               = React.addons.TestUtils,
+        RTU               = require("thirdparty/preact/preact-test-utils"),
         _                 = require("thirdparty/lodash");
 
+    // Preact Test Utils doesn't have findRenderedDOMComponentWithTag method
+    // So create it
+    RTU.findRenderedDOMComponentWithTag = function(root, tagName) {
```
petetnt: Instead of extending the `RTU` object can we just add `findRenderedDOMComponentWithTag` as a helper method?

``` js
function findRenderedDOMComponentWithTag(root, tagName) { ... }

...

findRenderedDOMComponentWithTag(...,...)
```

Just in case `findRenderedDOMComponentWithTag` gets added to RTU in future

petetnt:
Nice work @boopeshmahendran!

boopeshmahendran:
@petetnt Thanks for reviewing. I'll address these comments.

src/main.js
boopeshmahendran: @petetnt Actually react is being replaced with preact to remove facebook dependencies due to the PATENT issues. But preact-compat uses prop-types which is by facebook and we do not use proptypes in the `FileTreeView` code, so I removed it manually from preact-compat. So only we are inlining it to the `thirdparty` folder.

petetnt: Fair enough, wasn't aware that `compat` uses `prop-types` package as-is

swmitra:
I would really like to highlight the fact that @petetnt had actually suggested to use preact which resulted in minimal code changes. Great work @petetnt and @boopeshmahendran 👍 . 
Merging this PR now. Checked unit tests and ran smoke tests.

ficristo:
https://github.com/mikaeljorhult/brackets-todo uses React through 
```js
var React = brackets.getModule('thirdparty/react');
var ReactDOM = brackets.getModule('thirdparty/react-dom');
```
with this PR it will be broken.

petetnt:
@ficristo I can send a PR there later

boopeshmahendran:
@ficristo @petetnt I have raised a PR https://github.com/mikaeljorhult/brackets-todo/pull/145 at https://github.com/mikaeljorhult/brackets-todo changing react to preact

ficristo:
I was wondering if instead we should have handled this thing in core with some path mapping through RequireJS and possibly with some warnings.
If there isn't any way to remap / add warnings it should clearly noted in release notes this is a breaking change.

I noticed BracketsTODO is broken because I use it, but maybe there are other extensions.

If I remember correctly for other libs we use the path `thirdparty/<lib name>` why for preact is only `preact-compat` when use with `brackets.getModule`?

PR #13611, opened 2017-08-12
Fixes wrong function call in ProjectManager.js

hkirat:
The createNewItem function in ProjectManager.js calls the createAtPath in Projectmodel.js.

The function call expects a single parameter, which is the path at which the file must be created.
And it expects a "/" in the end of the path in case it is a folder.

However, right now 2 parameters are being passed. One of them being the isFolder variable.

### DIFF
--- a/src/project/ProjectManager.js
+++ b/src/project/ProjectManager.js
@@ -1072,7 +1072,10 @@ define(function (require, exports, module) {
         baseDir = model.getDirectoryInProject(baseDir);
 
         if (skipRename) {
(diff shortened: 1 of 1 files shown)


saurabh95 (approved):

saurabh95:
Thanks for the contribution!
Could you please sign the agreement http://dev.brackets.io/brackets-contributor-license-agreement.html ?
This might be one of the reason because of cla check failing.

hkirat:
Signed it!

swmitra (approved):
Good job @hkirat 👍

PR #13625, opened 2017-08-21
Default code coloring change to distinguish variables from other foreground tokens like braces, operators

swmitra:
With this PR, default code coloring in Brackets ( Default Light and Dark) being changed to distinguish between variables and other tokens colored as fore ground. Consider the following snippet - 
image
With this change set, the same snippet will be rendered as - 
image

Along with this change we might have to implement some more changes to distinguish between call expressions and simple variable access. As suggested by @marijnh in CM issue [4906](https://github.com/codemirror/CodeMirror/issues/4906), I am looking into extending our JavaScript mode with [google-modes](https://github.com/codemirror/google-modes)

### DIFF
--- a/src/extensions/default/DarkTheme/main.less
+++ b/src/extensions/default/DarkTheme/main.less
@@ -67,7 +67,7 @@
 .cm-number, .cm-attribute, .cm-plus {color: #85a300;}
 .cm-def, .cm-property {color: #b77fdb;}
 .cm-operator, .cm-meta, .cm-bracket {color: @foreground;}
-.cm-variable, .cm-variable-2, .cm-variable-3 {color: #00b9c4;}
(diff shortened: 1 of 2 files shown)


petetnt (commented):
👍 for the idea, but would like if the light theme variables would pass WCAG AA for normal text sizes. With the changes wouldn't need to be drastic even http://webaim.org/resources/contrastchecker/

swmitra:
@petetnt Modified `cm-variable` colors to improve the contrast ratio to acceptable limit. PSB, the screenshot from http://webaim.org/resources/contrastchecker/
For DarkTheme 
darkthemecontrast
For LightTheme 
lightthemecontrast

petetnt (approved):

swmitra:
@petetnt That was really quick. Thanks a lot 👍

PR #13635, opened 2017-08-23
JSUtils modification to handle es6 constructs

swmitra:
This PR will replace the existing regex based function def extraction with Acorn and also adds the capability to identify classes and methods (es6). Our current implementations is having some limitations and issues. One issue which is noticed during my dev test is wrong location information for function definition. Consider the following snippet - 
``` javascript 
Handsome.prototype.some = function () {
}
```
From the function/type list when we select `some`, instead of highlighting the member "some" , "some" from `Handsome` gets highlighted. With the new implementation, this issue is resolved as we get location information from the AST generated by Acorn. 

Couple of other observations which are actually open questions to all reviewers, there are some odd syntax which is considered valid in our current regex based implementation. Consider the following snippets - 
``` javascript 
something: function () {
}
```
``` javascript 
function hello() {
     something: function () {
     }
}   
```
It has been handled in the new implementation to let the existing tests pass, should we consider this to be valid?

Second observation is regarding a broken syntax for which we have a negative test case(this single test fails in the new implementation), but in the new implementation, Acorn parses it successfully. The snippet in consideration is - 
``` javascript
function invalid name () {
}
```
How do we handle this case?

@zaggino @petetnt @ficristo @MarcelGerber Can you please have a look at this PR?

### DIFF
--- a/src/extensions/default/QuickOpenJavaScript/main.js
+++ b/src/extensions/default/QuickOpenJavaScript/main.js
@@ -65,9 +65,7 @@ define(function (require, exports, module) {
         var lines = docText.split("\n");
         var functions = JSUtils.findAllMatchingFunctionsInText(docText, "*");
         functions.forEach(function (funcEntry) {
-            var chFrom = funcEntry.columnStart;
-            var chTo = funcEntry.columnEnd;
-            functionList.push(new FileLocation(null, funcEntry.nameLineStart, chFrom, chTo, funcEntry.label || funcEntry.name));
+            functionList.push(new FileLocation(null, funcEntry.nameLineStart, funcEntry.columnStart, funcEntry.columnEnd, funcEntry.label || funcEntry.name));
         });
         return functionList;
     }
--- a/src/extensions/default/QuickOpenJavaScript/main.js
+++ b/src/extensions/default/QuickOpenJavaScript/main.js
@@ -65,9 +65,9 @@ define(function (require, exports, module) {
         var lines = docText.split("\n");
         var functions = JSUtils.findAllMatchingFunctionsInText(docText, "*");
         functions.forEach(function (funcEntry) {
-            var chFrom = lines[funcEntry.lineStart].indexOf(funcEntry.name);
-            var chTo = chFrom + funcEntry.name.length;
-            functionList.push(new FileLocation(null, funcEntry.lineStart, chFrom, chTo, funcEntry.name));
+            var chFrom = funcEntry.columnStart;
+            var chTo = funcEntry.columnEnd;
+            functionList.push(new FileLocation(null, funcEntry.nameLineStart, chFrom, chTo, funcEntry.label || funcEntry.name));
         });
         return functionList;
     }
--- a/src/language/JSUtils.js
+++ b/src/language/JSUtils.js
@@ -28,8 +28,9 @@ define(function (require, exports, module) {
     "use strict";
 
     var _          = require("thirdparty/lodash"),
-        AcornLoose = require("node_modules/acorn/dist/acorn_loose"),
-        ASTWalker  = require("node_modules/acorn/dist/walk");
+        Acorn      = require("thirdparty/acorn/dist/acorn"),
+        AcornLoose = require("thirdparty/acorn/dist/acorn_loose"),
+        ASTWalker  = require("thirdparty/acorn/dist/walk");
 
     // Load brackets modules
     var CodeMirror              = require("thirdparty/CodeMirror/lib/codemirror"),
@@ -55,7 +56,8 @@ define(function (require, exports, module) {
      * @return {Object.<string, Array.<{offsetStart: number, offsetEnd: number}>}
      */
     function _findAllFunctionsInText(text) {
-        var results = {},
+        var AST,
+            results = {},
             functionName,
             resultNode,
             memberPrefix,
@@ -63,7 +65,11 @@ define(function (require, exports, module) {
    
         PerfUtils.markStart(PerfUtils.JSUTILS_REGEXP);
         
-        var AST = AcornLoose.parse_dammit(text, {locations: true});
+        try {
+            AST = Acorn.parse(text, {locations: true});
+        } catch (e) {
+            AST = AcornLoose.parse_dammit(text, {locations: true});
+        }
         
         function _addResult(node, offset, prefix) {
             memberPrefix = prefix ? prefix + " - " : "";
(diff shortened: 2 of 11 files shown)


src/language/JSUtils.js
```diff
@@ -68,22 +57,102 @@ define(function (require, exports, module) {
     function _findAllFunctionsInText(text) {
         var results = {},
             functionName,
+            resultNode,
+            memberPrefix,
             match;
-
+
```
petetnt: Nit: whitespace

src/language/JSUtils.js
```diff
@@ -68,22 +57,102 @@ define(function (require, exports, module) {
...
+                    if (node.label) {
+                        _addResult(node.label);
+                    }
+                }
+            }
+        });
 
         PerfUtils.addMeasurement(PerfUtils.JSUTILS_REGEXP);
-
+
```
petetnt: Nit: whitespace

src/language/JSUtils.js
```diff
@@ -21,15 +21,15 @@
...
-/*jslint regexp: true */
-
 /**
  * Set of utilities for simple parsing of JS text.
  */
 define(function (require, exports, module) {
     "use strict";
 
     var _ = require("thirdparty/lodash");
+    var Acorn_Loose = require("node_modules/acorn/dist/acorn_loose");
```
petetnt: NIT: Should we be consistent and use `AcornLoose` and `ASTWalker`?

petetnt (commented):
The functionality works great and I only have couple of nits.

petetnt:
FWIW I think all of those three broken ones should be invalid

src/extensions/default/QuickOpenJavaScript/main.js
```diff
@@ -65,9 +65,9 @@ define(function (require, exports, module) {
         var lines = docText.split("\n");
         var functions = JSUtils.findAllMatchingFunctionsInText(docText, "*");
         functions.forEach(function (funcEntry) {
-            var chFrom = lines[funcEntry.lineStart].indexOf(funcEntry.name);
-            var chTo = chFrom + funcEntry.name.length;
-            functionList.push(new FileLocation(null, funcEntry.lineStart, chFrom, chTo, funcEntry.name));
+            var chFrom = funcEntry.columnStart;
+            var chTo = funcEntry.columnEnd;
+            functionList.push(new FileLocation(null, funcEntry.nameLineStart, chFrom, chTo, funcEntry.label || funcEntry.name));
```
saurabh95: Instead, you can write directly 
`functionList.push(new FileLocation(null, funcEntry.nameLineStart, funcEntry.columnStart, funcEntry.columnEnd, funcEntry.label || funcEntry.name));`

saurabh95 (approved):

swmitra:
@petetnt Addressed code review NITs and added test cases for classes, inheritance, getter/setter and static methods.

src/language/JSUtils.js
```diff
@@ -21,15 +21,15 @@
...
 /**
  * Set of utilities for simple parsing of JS text.
  */
 define(function (require, exports, module) {
     "use strict";
 
-    var _ = require("thirdparty/lodash");
+    var _          = require("thirdparty/lodash"),
+        AcornLoose = require("node_modules/acorn/dist/acorn_loose"),
+        ASTWalker  = require("node_modules/acorn/dist/walk");
```
ficristo: I'm quite sure this will not work on dist.
You should add some logic here:
https://github.com/adobe/brackets/blob/abdf55d51f7c4bb6160a4a7b20048d8d0ec31e08/Gruntfile.js#L143

Maybe going forward we should use more webpack, but in another PR.

src/language/JSUtils.js
```diff
@@ -68,19 +57,99 @@ define(function (require, exports, module) {
...
             match;
-
+   
         PerfUtils.markStart(PerfUtils.JSUTILS_REGEXP);
-
-        while ((match = _functionRegExp.exec(text)) !== null) {
-            functionName = (match[2] || match[5]).trim();
-
+        
+        var AST = AcornLoose.parse_dammit(text, {locations: true});
```
ficristo: I think this work by checking indentation of the code.
So you need
1) use always first Acorn.parse and if it fails use AcornLoose.parse_dammit
2) add some tests of code not indented correctly (and / or without indentation at all)

src/language/JSUtils.js
```diff
@@ -68,19 +57,99 @@ define(function (require, exports, module) {
...
+                }
+            );
         }
+        
+        ASTWalker.simple(AST, {
+            /*
+                function <functionName> () {}
+            */
+            FunctionDeclaration: function (node) {
+                if (node.id.name !== '✖') { // A hack to reject anonymous declarations
```
ficristo: If Acorn use '✖' I don't think the commit is really good.
Maybe something saying that Acorn use '✖' to detect anonymous declarations and we want to skip them.

ficristo (changes requested):

ficristo:
Is JSUtils supposed to work on code not completed? If so you need to add some tests.

ficristo:
Is it possible to detect arrow functions? i.e. `var foo = () => {}`
Does it work with async functions? i.e. `var foo = { async bar() {} };`

petetnt:
Arrow functions (and functionProperties) worked when I tested, I assume that it works with async ones too

ficristo:
> worked when I tested

Great! Make sure to add soem test (if there aren't)

swmitra:
This never got handled though - 
 ``` javascript 
var foo = () => {}
 ```
Adding a walk option for this now 😃

src/language/JSUtils.js
swmitra: Consider this snippet - 
``` javascript 
key: function () {
}
```
This is actually a broken syntax and acorn can't parse it, but acorn_loose does it. FYI the acorn_loose code which uses this as is - 
``` javascript 
LooseParser.prototype.dummyIdent = function dummyIdent () {
  var dummy = this.dummyNode("Identifier");
  dummy.name = "✖";
  return dummy
};
```
The problem here is `FunctionDeclaration` walk option catches this and if I don't have a check, ends up putting "✖" as a function name. To actually put `key` as function name we have a `LabeledStatement` walk option. For your reference this is how the node looks like - 
image


This is not exactly a normal case but rather a broken syntax handler where this gets initialized with this name. I am not sure of any better way to handle this.

ficristo: Sorry @swmitra I meant comm**ent** not comm**it** !
I simply want you to rephrase the commit to avoid the word hack, since is how acorn works.

swmitra: I am sorry @ficristo , misinterpreted the comment 😞 . I have changed the comment in the code now  and also added the ArrowFunctionExpression construct along with addressing other code review comments and test cases.

swmitra:
@ficristo Can you please have a look at the change list again? I have addressed all the comments .

ficristo:
Are [generator functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*) supposed to be handled? If so, please add a test.

I just want to note I actually didn't test this PR, but the code seems good.

ficristo (approved):

Issue #13307, opened 2017-04-17
How to disable the reverse inspect ?


Doremiska:
Hi, I'm starting to get bothered by the new feature "reverse inspect", it's moving me in my code and I spend time always to come back where I was and I didn't find any way to disable this feature. Can you please let me know if there is a way to disable this feature ?

swmitra:
We don't have a dedicated preference for this, but setting 'livedev.wsport' to null should do the trick as websocket for reverse inspect won't be created in that case.

swmitra:
Just realized it might not work. Let us work on this to expose a preference.

AntonioDeCasper:
Same problem here. I want to disable this feature but i can't...

onmyflow:
Same problem here,
its really starting to annoy me, when I'm working on a section **at the bottom of my website**, that has the same class as one at the top of my website, and then the browser **always jumps to the top**. 
But I also wouldn't want to work without live preview ... 
Well then. 
I'm downgrading again.

siuangie91:
Same here! I thought I was the only one super annoyed by it. I was annoyed within the first 5 minutes of using it. Shame, because I liked the other updated features.

G3-Logic:
Another request to PLEASE give us a way to turn reverse inspect off. Can't think of any situation where I would want to have to sift through my code to find out where I was, just because I happened to click on an element in the browser. SUPER annoying.

G3-Logic:
Any progress on this? Thought Enable/Disable default extensions in 1.10 would mean that I could turn this off. But not seeing it.

manchinesmith:
does this work with the livedev.multibrowser true

PR #13659, opened 2017-09-04
Now reverse inspect can be disabled by setting the 'livedev.enableReverseInspect' pref to false 

saurabh95:
Fixes: https://github.com/adobe/brackets/issues/13307

### DIFF
--- a/src/LiveDevelopment/LiveDevelopment.js
+++ b/src/LiveDevelopment/LiveDevelopment.js
@@ -201,6 +201,10 @@ define(function LiveDevelopment(require, exports, module) {
     PreferencesManager.definePreference("livedev.wsPort", "number", 8125, {
         description: Strings.DESCRIPTION_LIVEDEV_WEBSOCKET_PORT
     });
+    
+    PreferencesManager.definePreference("livedev.enableReverseInspect", "boolean", true, {
+        description: Strings.DESCRIPTION_LIVEDEV_ENABLE_REVERSE_INSPECT
+    });
 
     function _isPromisePending(promise) {
         return promise && promise.state() === "pending";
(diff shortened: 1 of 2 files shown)


src/LiveDevelopment/LiveDevelopment.js
```diff
@@ -1369,7 +1369,10 @@ define(function LiveDevelopment(require, exports, module) {
             // wait for server (StaticServer, Base URL or file:)
             prepareServerPromise
                 .done(function () {
-                    WebSocketTransport.createWebSocketServer(PreferencesManager.get("livedev.wsPort"));
+                    var wsPort = PreferencesManager.get("livedev.wsPort");
```
petetnt: Any reason not to use a different pref instead, such as `livedef.enableReverseInspect`

swmitra: @petetnt IMO, that might confuse a lot of users. Instead if an invalid port (0) is given by the user, it's implicit to assume that reverse inspect is not required.

petetnt: My 2 cents is that average user doesn't even know what a `port` is, even without considering what might be an invalid port and what isn't 🤔

swmitra: I am a pendulum now 😃 . I had asked the same initially to @saurabh95 but somehow got convinced with the explanation. But now I am convinced by your explanation again so would ask the author to include a new preference. @saurabh95 over to you...

saurabh95: Yeah, for a user who is not aware of ports, it would be a difficult way to disable reverse inspect. I will  add a new preference and will update this PR only.

petetnt: 👍

swmitra:
@petetnt Can you please have a look at the new pref? I have to merge this PR today and get the localized description strings as well.

petetnt (approved):

swmitra:
Great work @saurabh95 and @petetnt 👍

PR #13664, opened 2017-09-06
Change Preact require to thirdparty/preact

boopeshmahendran:


test/SpecRunner.js
```diff
@@ -39,6 +39,12 @@ require.config({
         "preact-test-utils"             : "thirdparty/preact-test-utils/preact-test-utils",
         "simulate-event"                : "thirdparty/simulate-event/simulate-event",
         "xtend"                         : "thirdparty/xtend"
+    },
+    map: {
+        "*": {
+            "thirdparty/preact":            "preact-compat",
```
saurabh95: nit. the colon should be in the same column as the lines above and below.

test/spec/FileTreeView-test.js
```diff
@@ -29,9 +29,9 @@ define(function (require, exports, module) {
 
     var FileTreeView      = require("project/FileTreeView"),
         FileTreeViewModel = require("project/FileTreeViewModel"),
-        Preact            = require("preact-compat"),
+        Preact            = require("thirdparty/preact"),
         Immutable         = require("thirdparty/immutable"),
-        PTU               = require("preact-test-utils"),
+        PTU               = require("thirdparty/preact-test-utils"),
```
saurabh95: I would recommend having a more readable name instead of `PTU`, `PreactTestUtils` may be.

saurabh95 (changes requested):

swmitra (approved):

Issue #13656, opened 2017-09-04
Unable to view content of files when folder on desktop is opened with Brackets in Ubuntu 64 Bit


ayushrathi15:
### Description
Unable to view content of files when folder on desktop is opened with Brackets in Ubuntu 64 Bit

### Steps to Reproduce

1. Open Brackets
2. Open a folder placed on desktop.
3. Click on any file inside the folder to view the contents.
4. Content of the file is not displayed.

**Expected behavior:** Content of the file should be displayed

**Actual behavior:** Content of the file is not displayed

**Work Around:** Switch to another folder and come back. Contents will be visible. 

### Versions

Linux Ubuntu 64 Bit
Brackets Release 1.11 build 1.11.0-17507

hwalinga:
Seems a bit like the error I reported a few weeks ago: https://github.com/adobe/brackets/issues/13619
Hope this will also resolve mine.

Zhanxin:
Also reproduce on Debian 32 bit OS

MathiasMaurilio:
Hi @ayushrathi15 it's been a while since the last update on this issue.
Could you tell me if in the new version this is still happening?

kwcooper:
same for me on ubuntu 18.04

PR #13667, opened 2017-09-06
Adding null check for encoding preference

swmitra:
This PR is a fix for #13656.
For some strange reason, in the reported scenario encoding preference fetched is undefined which causes the code to break while checking for encoding cache with the given file path. This check is good to have even without the reported issue. 

@saurabh95 @nethip @petetnt  Can you please have a look at this PR?


nethip:
@swmitra I am OK with this change. Can you debug as to why we are getting encoding as null in these scenarios?

saurabh95:
Ideally it should be created on project open as per https://github.com/adobe/brackets/blob/master/src/editor/EditorStatusBar.js#L533 .
Could you please provide some repro steps.

Issue #13708, opened 2017-09-18
Drag and drop a folder in brackets in Win 10 and Win 8.1 opens live view.


ayushrathi15:
### Description

Drag and drop a folder in brackets in Win 10 and Win 8.1 opens live view.

### Steps to Reproduce

1. Launch brackets.
2. Drag and drop a sites folder.
3. It shows live view.

![win 10 drag and drop - 17517](https://user-images.githubusercontent.com/25339865/30529507-c2837766-9c5b-11e7-9471-c984f68e3d94.PNG)

![win 8 drag and drop - 17517](https://user-images.githubusercontent.com/25339865/30529508-c53b7df0-9c5b-11e7-88bc-2d32e8d7bc83.PNG)

**Expected behavior:** It should show code editor

**Actual behavior:** It opens live view.

### Versions

Windows 10 and Windows 8.1
Brackets Release 1.11 build 1.11.0-17517

ayushrathi15:
This has been verified.

PR #13711, opened 2017-09-19
Prevent data loss in case of window url change 

swmitra:
This PR fixes potential data loss in scenarios reported as part of #13708.
@narayani28 @nethip @petetnt  @boopeshmahendran  Please have a look at this PR.

### DIFF
--- a/src/document/DocumentCommandHandlers.js
+++ b/src/document/DocumentCommandHandlers.js
@@ -1718,6 +1718,20 @@ define(function (require, exports, module) {
 
     /** Reload Without Extensions commnad handler **/
     var handleReloadWithoutExts = _.partial(handleReload, true);
+    
+    /** 
+     * Attach a beforeunload handler to notify user about unsaved changes. 
+     * Prevents data loss in scenario reported under #13708
+    **/
+    window.onbeforeunload = function(e) {
+        var openDocs = DocumentManager.getAllOpenDocuments();
+        openDocs = openDocs.filter(function(doc) {
+            return doc && doc.isDirty;
+        });
+        if (openDocs.length > 0 && !_windowGoingAway) {
+            return Strings.WINDOW_UNLOAD_WARNING;
+        }
+    };
 
     /** Do some initialization when the DOM is ready **/
     AppInit.htmlReady(function () {
--- a/src/document/DocumentCommandHandlers.js
+++ b/src/document/DocumentCommandHandlers.js
@@ -1720,19 +1720,27 @@ define(function (require, exports, module) {
     var handleReloadWithoutExts = _.partial(handleReload, true);
     
     /** 
-     * Attach a beforeunload handler to notify user about unsaved changes. 
+     * Attach a beforeunload handler to notify user about unsaved changes and URL redirection in CEF. 
      * Prevents data loss in scenario reported under #13708
     **/
     window.onbeforeunload = function(e) {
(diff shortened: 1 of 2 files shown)


boopeshmahendran (approved):
LGTM

src/nls/root/strings.js
```diff
@@ -138,6 +138,7 @@ define({
     "EXT_MODIFIED_WARNING"              : "<span class='dialog-filename'>{0}</span> has been modified on disk outside of {APP_NAME}.<br /><br />Do you want to save the file and overwrite those changes?",
     "EXT_MODIFIED_MESSAGE"              : "<span class='dialog-filename'>{0}</span> has been modified on disk outside of {APP_NAME}, but also has unsaved changes in {APP_NAME}.<br /><br />Which version do you want to keep?",
     "EXT_DELETED_MESSAGE"               : "<span class='dialog-filename'>{0}</span> has been deleted on disk outside of {APP_NAME}, but has unsaved changes in {APP_NAME}.<br /><br />Do you want to keep your changes?",
+    "WINDOW_UNLOAD_WARNING"             : "You have unsaved changes! Are you sure you want to quit Brackets and navigate to a different URL?",
```
narayani28: The message looks incorrect. Quit brackets and navigate? Should you just have are you sure you want to quit brackets? or your unsaved changes will be lost?

src/document/DocumentCommandHandlers.js
```diff
@@ -1718,6 +1718,20 @@ define(function (require, exports, module) {
 
     /** Reload Without Extensions commnad handler **/
     var handleReloadWithoutExts = _.partial(handleReload, true);
+    
+    /** 
+     * Attach a beforeunload handler to notify user about unsaved changes. 
+     * Prevents data loss in scenario reported under #13708
+    **/
+    window.onbeforeunload = function(e) {
```
narayani28: this will be called even in normal scenarios right? will the return value be a NOP in those cases?

src/nls/root/strings.js
swmitra: Brackets is just an app which was being run in shell. Think of this as having an unpaid cart and try to move away from there by typing an URL in the address bar.

src/document/DocumentCommandHandlers.js
swmitra: Yes. To safeguard that , there is an additional  flag check along with non zero dirty files. We will get in this handler even for a normal app close, but we will not do anything in that case.

narayani28: ok.

src/nls/root/strings.js
narayani28: Do you want to rephrase it as "Are you sure you want to navigate to a different URL and quit Brackets?"

narayani28 (approved):
Except for the rephrasing of the string, the changes look good to me.

src/nls/root/strings.js
swmitra: Makes sense. I will change the message.

swmitra:
Test cases 
- [x] Unchanged app quit workflow 
- [x] Unchanged app reload workflow with extensions
- [x] Unchanged app reload workflow without extensions 
- [x] User confirmation dialog with unsaved changes on programmatic URL redirection 
- [x] User confirmation dialog on programmatic URL redirection 
- [x] User confirmation dialog on default drop handler invocation after file path drag

PR #13736, opened 2017-09-26
Update README.md to reflect Linux updates

petetnt:
This PR removes the notes about Linux version being behind MacOS/Windows versions as it should no longer be the case with Brackets 1.11 👯


swmitra (approved):

swmitra:
@petetnt Thanks a lot for noticing this and update the relevant part.

PR #13751, opened 2017-10-02
Update index.html

mohmmad111:


petetnt (commented):
The weird looking tags are because of the Grunt plugin that manages those. Thanks for contributing anyway!

petetnt:
I think t

PR #13752, opened 2017-10-02
Set hidden attribute as a flag

abaracedo:
The `hidden` attribute should be displayed as a flag.

Now when is selected from the suggestion list to autocomplete is shown like this:

```
<p hidden=""></p>
```

With this change it is shown like this:

```
<p hidden></p>
```


petetnt:
Hi @abaracedo, could you sign our CLA? http://dev.brackets.io/brackets-contributor-license-agreement.html

abaracedo:
@petetnt Done! :)

petetnt (approved):
LGTM, thanks for this contribution @abaracedo, hopefully you'll contribute to Brackets in the future too!

PR #13759, opened 2017-10-08
-bugfix: Fixed bug which caused SearchResultsView to crash when double clicking buttons

facchettos:
-added : added an event listener to the pagination-col class with a event.stopPropagation to prevent the dblclick event to propagate to the toolbar, which caused the SearchResultsView to crash when clicking rapidly the next/previous button. First this will avoid the view to crash, second it will prevent users to accidentally close the panel.

-added: new event listener on mouse 3 when clicking the toolbar, which allows a better internal consistency (middle click to close is used to close the files in the working set), and a better external consistency (middle clicks are often used to close tabs on web browsers for example)

-added: tests to ensure that the event listeners I've had are working as expected


facchettos:
The error causing The Travis CI build to fail is this:

data: {"message":"API rate limit exceeded for 52.54.31.11. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://developer.github.com/v3/#rate-limiting"}

I don't really know what it is but I dont thing that it is coming from the code i wrote.

PR #13760, opened 2017-10-08
bugfix : double click to crash SearchResultsView

facchettos:
-added : added an event listener to the pagination-col class with a event.stopPropagation to prevent the dblclick event to propagate to the toolbar, which caused the SearchResultsView to crash when clicking rapidly the next/previous button. First this will avoid the view to crash, second it will prevent users to accidentally close the panel.

-added: new event listener on mouse 3 when clicking the toolbar, which allows a better internal consistency (middle click to close is used to close the files in the working set), and a better external consistency (middle clicks are often used to close tabs on web browsers for example)

-added: tests to ensure that the event listeners I've added are working as expected

This pull request fix the #13768 issue

### DIFF
--- a/src/search/SearchResultsView.js
+++ b/src/search/SearchResultsView.js
@@ -181,7 +181,7 @@ define(function (require, exports, module) {
                 HealthLogger.searchDone(HealthLogger.SEARCH_PREV_PAGE);
             })
             // The link to go to the next page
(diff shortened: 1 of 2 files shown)


facchettos:
here are the screenshots:
screenshot from 2017-10-08 20-21-07
and after rapid clicks on the next buttons:
screenshot from 2017-10-08 20-20-56

PR #13830, opened 2017-11-09
Add wrapper functions to call the fs-extra ones.

ficristo:
fs-extra.copy doesn't seem to be called correctly.
Apparently NodeDomain doesn't work on functions which have optional
arguments like fs-extra.copy

Now LowLevelFileIO unit tests run again.

### DIFF
--- a/test/node/TestingDomain.js
+++ b/test/node/TestingDomain.js
@@ -28,6 +28,18 @@
 
 var fs = require("fs-extra");
 
+function remove(path, cb) {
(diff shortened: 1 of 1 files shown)


petetnt (approved):
LGTM (as in tests run again), however the tests are failing for me (on MacOS). Is this expected?

ficristo:
Thanks @petetnt !
Yes, I think currently there are quite a few failures across the testsuite.

PR #13847, opened 2017-11-15
 Edit typographical errors

JeongGoEun:
appled -> applied
descandent-> descendant


Gruntfile.js
```diff
@@ -335,7 +335,7 @@ module.exports = function (grunt) {
                     // The need for this should go away once the version of phantomjs gets upgraded to 2.0
                     'test/polyfills.js',
 
-                    'src/thirdparty/jquery-2.1.3.min.js',
+                    'src/thirdparty/jquery-3.2.1.min.js',
```
petetnt: Hmm upgrade to 3.2.1 seems dangerous here, can we do it in a separate PR?

src/index.html
```diff
@@ -1,4 +1,4 @@
-<!doctype html>
+<!DOCTYPE html>
```
petetnt: `doctype` is fine

src/index.html
```diff
@@ -35,19 +35,19 @@
     <script type="application/javascript" src="xorigin.js"></script>
 
     <link rel="stylesheet" type="text/css" href="thirdparty/CodeMirror/lib/codemirror.css">
-    <!--(if target dev)><!-->
+    <!--(if target dev)-->
```
petetnt: I wonder if these should be like they were in the original? The changes make sense though. @ficristo ?

petetnt (changes requested):
Hi,

left you some comments

Gruntfile.js
JeongGoEun: @petetnt Ok. I'll try it
Thank you!

src/index.html
ficristo: Those comments are used by https://github.com/changer/grunt-targethtml
The changes here should be reverted.

petetnt: Yep, that's what I though, thanks @ficristo !

Gruntfile.js
JeongGoEun: #13854
@petetnt There is a separate PR :)

PR #13824, opened 2017-11-08
Add api for adding context submenus

boopeshmahendran:
This PR will add api  for creating and removing context submenus.

**API:**

1. `addSubMenu`

**Example Code:-**
```js
// Put this code in src/command/DefaultMenus.js
var subMenu = editor_cmenu.addSubMenu("submenu", "test");
subMenu.addMenuItem(Commands.EDIT_CUT);
subMenu.addMenuItem(Commands.EDIT_PASTE);
subMenu.addMenuItem(Commands.EDIT_SELECT_ALL);
subMenu.addMenuItem(Commands.TOGGLE_QUICK_EDIT);
subMenu.addMenuItem(Commands.EDIT_COPY);
```

2. `removeSubMenu`

**Example Code:-**
```js
editor_cmenu.removeSubMenu(subMenu.id);
```

@swmitra @petetnt @nethip Please review

PR #13873, opened 2017-11-17
Adding interface for refactoring.

navch:
This PR is just to add interface for Code refactoring. This have same architecture as CodeHintManager. To provide some refactoring features (i.e. Rename, Extract etc) we can register a provider(as extension) directly. This interface will take care of adding commands and menus for all available handlers. This also responsible for calling appropriate handler based on priority.

What's next - 
----------------
- [ ] This interface should also take care of some common UI pieces for all languages

Upcoming PR's which are going to use this interface
- [ ] Rename reference in scope for JavaScript
- [ ] Extract some expression to a variable in scope for JavaScript

Note :- This PR Uses submenu API @boopeshmahendran raise a [PR](https://github.com/adobe/brackets/pull/13824) to provide addSubmenu API.

Please Review @swmitra @petetnt @nethip @boopeshmahendran

### DIFF
--- a/src/editor/CodeRefactoringManager.js
+++ b/src/editor/CodeRefactoringManager.js
@@ -1,0 +1,231 @@
+/*
+ * Copyright (c) 2012 - present Adobe Systems Incorporated. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+/*
+ * __CodeRefactoringManager Overview:__
+ *
+ * The CodeRefactoringManager mediates the interaction between the editor and a
+ * collection of refactoring providers. If user opens context menus 
+ * then the providers registered for the current language are queried
+ * for their ability to provide refactoring handlers in order of descending priority by
+ * way their hasHanlders methods.
+ * Hanlders means all availble methods of refactoring in that provider, For example
+ * Rename, Extract, Change Signature etc.
+ *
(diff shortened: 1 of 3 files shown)


petetnt:
Cool stuff @navch! I'll try to review this soon-ish.

src/editor/CodeRefactoringManager.js
```diff
@@ -0,0 +1,231 @@
...
+ */
+
+/*
+ * __CodeRefactoringManager Overview:__
+ *
+ * The CodeRefactoringManager mediates the interaction between the editor and a
+ * collection of refactoring providers. If user opens context menus 
+ * then the providers registered for the current language are queried
+ * for their ability to provide refactoring handlers in order of descending priority by
+ * way their hasHanlders methods.
```
nethip: nit: correct the spelling mistake.

nethip: I think there are quite a few spelling mistakes in the description. Could you correct all of these?

src/editor/CodeRefactoringManager.js
```diff
@@ -0,0 +1,231 @@
...
+     */
+    function _providerSort(a, b) {
+        return b.priority - a.priority;
+    }
+
+    /**
+     * The method by which a CodeRefactoringProvider registers its willingness to
+     * help in code refactoring for editors in a given language.
+     *
+     * @param {!providerInfo} provider
```
nethip: Can use mention the structure of `providerInfo`?

src/editor/CodeRefactoringManager.js
```diff
@@ -0,0 +1,231 @@
...
+
+        var editor   = EditorManager.getActiveEditor();
+
+        // Don't look for refactoring handlers if we have a multiple selection.
+        if (editor.getSelections().length > 1) {
+            return;
+        }
+
+        // Find a suitable provider, if any
+        var language = editor.getLanguageForSelection(),
```
nethip: How expensive is this operation `getLanguageForSelection`?

src/editor/CodeRefactoringManager.js
```diff
@@ -0,0 +1,231 @@
...
+                editorCmenu.removeSubMenu(Commands.CODE_REFACTOR);
+            }
+        }
+    }
+    AppInit.htmlReady(function () {
+
+        editorCmenu = Menus.getContextMenu(Menus.ContextMenuIds.EDITOR_MENU);
+
+        // Check for handlers on right click
+        editorCmenu.on("beforeContextMenuOpen", function (e) {
```
nethip: Check `editorCmenu` for null?

src/editor/CodeRefactoringManager.js
```diff
@@ -0,0 +1,231 @@
...
+                submenu.addMenuItem(element.id);
+            });
+        }
+    }
+
+    /**
+     * Find a suitable provider for a language id and
+     * Add a context menu if there is any provider available
+     */
+    function getHandlersForLanguageId() {
```
nethip: You might want to change the name of this function as it more than a `getter`. Also all private function need to have `_` prepended.

nethip:
@navch Thanks for this PR! Here is my feedback

* My view is that the framework seems to be too generic. You might want to club common operations as common commands and let the extension handle the language specific part of Rename. You can think of it as three common commands exposed, and the later ask the extensions to add in for more. Otherwise from the UX point of view, it is a distraction for me to know the order and understand the menus when we switch between languages like HTML/JS/CSS.

* I am not sure having the refactor part only in sub-menus is a good option. If some extension overwrites the context menu then there is no way refactor can ever be used. How about having those options in the top menu as well? I know that we don't yet support sub menus but it is a good to give it a thought in this perspective.

* Have an option to associate a shortcut with the rename and refactoring menus is a good idea.

* Apart from the back-end part, there is no help w.r.t UI paradigm for extension developers. It might be a good thing, in terms of freedom but also bad thing in terms of different providers implementing rename and refactoring differently for different languages. We might want to still stick with this framework but come up with a Wiki page as to document the most common practices.


@petetnt Did you get a chance to review this PR?

navch:
Closing this PR for now.  Later on, if we plan to ship refactoring feature for other languages then we can open this again after addressing these review comments.

PR #13879, opened 2017-11-18
Edit the wrong spacing

kimohyeong:
builtin -> built in


petetnt:
Hi @kimohyeong,

there's some new conflicts that would need to be solved before we can merge this. If you need any help with that, do reply in this PR 👍

PR #13887, opened 2017-11-20
Fix Grammar adverb location

JeongGoEun:
The adverb 'always' is usually put after the verb 'be'.
So, always be -> be always


petetnt (commented):
Both are grammatically correct, I think `always be` might cover the intention better in this case 🤔 No strong feeling about this though.

PR #13888, opened 2017-11-20
Correct typo (it run -> it runs)

Jeonghyunsu:
it run -> it runs (41 line)
(/src/utils/Async.js)


petetnt:
I think the correct tense here is pluperfect, as in `had run` (would have run...).

Thanks for contributing!

PR #13892, opened 2017-11-21
Fix typo (initalization -> initialization)

JungGaBin:
initalization -> initialization


petetnt:
This is a typo in a third party dep and should be filed to jquery/jquery-ui

thanks!

PR #13903, opened 2017-11-22
Correct typo (ommitted -> omitted)

Jeonghyunsu:
(/src/extensions/default/JavaScriptQuickEdit/unittest-files/jquery-ui/external/jshint.js) (282line)

ommitted -> omitted


petetnt:
Hi @Jeonghyunsu,

this typo is in a 3rd party dependency so it should be fixed in jquery/jquery-ui.

Thanks for contributing!

Issue #13800, opened 2017-10-28
Impossible to read text too small on the side bar and system messages on a 3840 x 2160 screen


gustavocast:
Using Lenovo Yoga 910 with 3840 x 2160 screen resolution
OS: Ubuntu 17.04 
Brackets: 1.11 Release 1.11.0 17524
Problem 1: It is impossible to read text on the side bar and on the pop up system messages. The text it is too small. I had to use a 4x magnifier to be able to read data to write this report. 
Problem 2: Preview can't launch either Chrome or Firefox. Connecting to browser. In order for live preview to connect, Chrome needs to relaunch with remote debugging. The button relaunch Chrome is not responding. No changes if I restart chrome manually.

nethip:
Could you try download the installer from http://brackets.io  and try it out?

PR #13933, opened 2017-11-27
Fixed Issue #13800 - Side-bar now readable on bigger screens

JoseFreixo:
Fixes Issue #13800

capturar


up201504988:
Can someone give us an inclination as to what we might be doing wrong and is causing the pull request to be rejected?

ficristo:
The error "API rate limit exceeded for..." can happen from time to time. When the github API are used too many times they could start failing with that message.
It has nothing to do with your PR.

ficristo:
I'm sorry I cannot review your PR, but I have a question: if you resize the window the font size will change?

up201504988:
@ficristo Yes, it is related to the height of the window.  This is what makes the font bigger on bigger screens. But we believe that most people will keep the window height the same size as the screen and only change the width.

Issue #11800, opened 2015-10-10
Core: The shortcuts key (Ctrl++, Ctrl+-) would not work if input a large number as the font size on Themes Settings dialog


Zhanxin:
Steps:
1. launch the Brackets
2. Click View > Themes...
3. Input a large number as the font size, such as "100px, 1000px..."
4. Click Done
5. Try to Press Ctrl++ and Ctrl+- to increase or decrease the font size

Result: The shortcuts key doesn't work
Expect: The shortcuts key could work normally

julietafrade97:
The problem was that when you try to increase or decrease the font size, brackets doesn't let you if the new font size is more than the maxium font size (72px) or less than the minium font size (1px). Therefore, if the the font size in themes settings is 100px and you try to drecrease it pressing Ctrl+- nothing happens because the new font size will be 99px which is still more than 72px. 
Because I think it makes sense to be able to decrease the font size even though it is still more than the allowed I tried to solve the problem. So now if the font size is more than the maxium and you try to decrease it, it will decrese to the maxium font size, and if the font size is less than the minium and you increase it, it will increase to the minium font size.
I'm going to make a pull request now and wait for the update in the master repository.

PR #13934, opened 2017-11-27
fixed issue #11800

julietafrade97:
The problem was that when you try to increase or decrease the font size, brackets doesn't let you if the new font size is more than the maxium font size (72px) or less than the minium font size (1px). Therefore, if the the font size in themes settings is 100px and you try to drecrease it pressing Ctrl+- nothing happens because the new font size will be 99px which is still more than 72px. 
Because I think it makes sense to be able to decrease the font size even though it is still more than the allowed I tried to solve the problem. So now if the font size is more than the maxium and you try to decrease it, it will decrese to the maxium font size, and if the font size is less than the minium and you increase it, it will increase to the minium font size.
I'm going to make a pull request now and wait for the update in the master repository.


julietafrade97:
I don't understand why this wasn't aprooved. Can someone explain?

petetnt:
Hi @julietafrade97,

sorry about that, I just haven't had the time to properly review / test it yet.

Issue #13890, opened 2017-11-20
Search goes back to previous search string


Badestrand:
### Description

When I want to search for the next occurence of my last search it sometimes falls back to searching the previous input.

### Steps to Reproduce

1. Open search with Cmd+F (or probably Ctrl+F for Windows)
2. Type "text1" and press Enter a few times to circle through the occurences
3. Exit search (ESC)
3. Double click on some other text, e.g. "text2"
4. Open search (contains "text2" now) and again press enter a few times to go to different occurences of "text2"
5. Exit search (ESC)
6. Open search -> "text1" is in search although my last search was for "text2" and the editor jumps to the next occurence of "text1" which might be wherever in the file

**Expected behavior:** Search input shows "text2" and editor should not jump anywhere

**Actual behavior:** Search input shows "text1" and the editor immediately jumps to the next occurence of "text1"

The editor obviously treats search differently depending on whether I filled the input by typing or by selecting text - I find that highly confusing.

### Versions

Release 1.11 Build 1.11.0-17524 (release 86b5e2d4e)

macOS Sierra 10.12.3

veryprofessionaldodo:
This seems to not be an issue on my build...

nethip:
@Badestrand Do you have extension installed? If yes can you try reloading Brackets without extensions using `Debug->Reload without Extensions` and check if you are seeing the issue.

Badestrand:
@nethip Yes, I had tested it without extensions (and just to be sure tested it again right now).

Badestrand:
#13955 seems to address the same issue.

I also updated the description -- to be super precise, there is an additional step between 6 and 7

TrocaTudo95:
It is really an issue, I tried it myself in MacOs, Windows and Linux and was definitely an issue and i solve it, because it was only populating the searchBar only with typed strings

TrocaTudo95:
@nethip can you check my pull request please?

PR #13955, opened 2017-12-04
bug on find solved

TrocaTudo95:
I solved the issue #13890 . The problem was that when the find bar is opened the function that gets the inital query only was getting the last typed text or a keyword selected and never a word that was searched without typing, now the problem is solved.


TrocaTudo95:
any feedback please?

swmitra (approved):
Good work 👍 @TrocaTudo95!

swmitra:
@petetnt @nethip Can you please have a look at this change?

nethip:
Thanks @TrocaTudo95. Is it OK to squash and merge this change?

nethip:
Thanks @TrocaTudo95

PR #13966, opened 2017-12-09
Possible Fix Issue #7855

pedrojlazevedo:
Added alternative key binding for "Save All" only for Linux.
Ctrl-Alt-S is a default key binding for Linux, so I created an alternative only for Linux, Shift-Alt-S.

### DIFF
--- a/src/base-config/keyboard.json
+++ b/src/base-config/keyboard.json
@@ -18,7 +18,13 @@
         "Ctrl-S"
     ],
     "file.saveAll":  [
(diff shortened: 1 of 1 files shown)


sorryifikillyou:
I verified that one of the Travis checks have failed and it appeared this message:

> "data: {"message":"API rate limit exceeded for 52.45.185.117. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)"
It's your problem?

PR #13980, opened 2017-12-15
Add Extract to Variable and Extract to Function features in js

boopeshmahendran:
This PR will add **Extract to Variable** and **Extract to Function** features

## Extract to Variable
1. Select **any expression** in js
2. Right-click and click **Extract to Variable**
3. If the expression is not valid, display error.
4. Else, it will remove leading and trailing spaces and semicolons and create a new variable in that scope and with a **unique name** and **replace all occurrences of that expression in that scope.**
5. Multi-select variable names for **renaming**

You can also click **Extract to Variable** without any selection in which case it will display a inline menu with **all surrounding expressions**(sorted by their length) in that line and just select the expression. Supports **sub expressions and ES6 expressions** also.

## **Extract to Function**
1. Select **any expression or a set of statement** in js
2. Right-click and click **Extract to function**
3. If not a valid set of statements or expression, displays error.
4. Displays a menu with all the surrounding scopes in it.
5. **Select the scope** where you want to extract.
6. Creates a new function in that scope.
7. Finds the **parameters which needs to be passed** based on 
               - Identifiers used in the selection
               - Identifiers available in the extracted scope
8. Finds the **return parameters which needs to be returned** from the function based on
               - Values of identifiers required in the rest of the scope after the selection.
               - Identifiers whose values are changed in the selection
9. Creates a function with **a unique name**
10. Multi-select function names for  **renaming**.

This feature **preserves variable declarations** and also handles **block scoped variables, this pointer, return statement, class scopes.**

**Note:-** The **InlineMenu widget** is inspired and copied from `editor/CodeHintList.js` and will be changed and/or refined.

**Please try these features and provide feedback.**

@nethip @navch @swmitra @petetnt @redmunds @zaggino @ficristo @MarcelGerber @ingorichter Please review.


swmitra:
@boopeshmahendran Amazing work! 💯 🥇  The implemented features worked flawlessly in my limited testing. In terms of code I really don't have any comments ( I don't want to comment solely on some coding styles ). A bold **LGTM** from my side. Do resolve the conflicts so that I can merge this PR.

swmitra (approved):

nethip:
A big thank you to @boopeshmahendran for this PR. Awesome! 👍

PR #13965, opened 2017-12-07
JavaScript Refactoring - Rename Identifier, Wrap selection, Convert to arrow function, Create Getters & Setters

navch:
This PR implements "rename" functionality in JS mode to enable intelligent rename feature by using Tern's scope analysis and inference functions. Renaming across multiple files has been intentionally blocked for the time being as we are not analyzing all the JS files in current project root right now.

The rename function works by selecting a variable def/ref or placing cursor in and using Ctrl+R.

This code is written by @swmitra and already reviewed by @petetnt ([PR 13047](https://github.com/adobe/brackets/pull/13047))

I just restructured this code and created new extension called JavascriptCodeRefactoring.

**Known issues:-**

In a particular case Tern inference engine is picking up wrong hit. For example :-
```javascript
function fn1(error) {
    error.a = "abc";
    var a = "test";
    console.log(a);
}
```
If we try to rename a in line 3 or 4 , Tern picks up fake ref from line 2 as well.

#Issue reported in tern [950](https://github.com/ternjs/tern/issues/950), [951](https://github.com/ternjs/tern/issues/951)
[951](https://github.com/ternjs/tern/issues/951) is already fixed.

**Wrap Selection in Try Catch/Condition** 
1) Select code/Just put cursor anywhere
2) Edit -> wrap in try catch/condition Or find this in submenu
3) If you just placed cursor it will find surrounding statements around cursor position
 else if you selected code then it will check weather these are statements or not
5) If statements -> It will wrap that code in TryCatch/Condition
 else error (Please select valid statements)

`AppInit.appReady(function () {
            brackets.test.doneLoading = true;
        });`

Select whole code or just place cursor on 3rd line it will wrap whole code in try catch block/Condition block

`try {
            AppInit.appReady(function () {
                brackets.test.doneLoading = true;
            });
        } catch (e) {
            console.log(e.message);
        }`
Put selection on 2nd line and do wrap in try catch/condition it will wrap that line only.

Know Issue:- Some time indention goes wrong in case of wrap in condition

**Convert to arrow function**
1) Put cursor anywhere in anonymous function and click convert to arrow function
If one param structure will be `param => {statements}`
If zero or more than one param `(param1, param2) => {statements}` 
In case of one statement, it will remove "{, }" and ";"as well

**Create Setters and Getters for given property**
1) Put cursor on property
2) Create getter/setter
`get getTest() {
 return this.test;
},
set setTest(val) {
this.test = val;
}`

[...]

PR #13994, opened 2017-12-21
Don't allow rename in inline widget if that identifier have reference…

navch:
This PR is in continuation of https://github.com/adobe/brackets/pull/13965.
This PR will help in deciding weather rename can be done or not inside inline widget.

Please review @nethip @swmitra @petetnt @boopeshmahendran

### DIFF
--- a/src/extensions/default/JavaScriptRefactoring/RenameIdentifier.js
+++ b/src/extensions/default/JavaScriptRefactoring/RenameIdentifier.js
@@ -111,12 +111,10 @@ define(function (require, exports, module) {
 
             //In case of inline widget if some references are outside widget's text range then don't allow for rename
             if (inlineWidget) {
-                var isInTextRange = true;
-                refs.forEach(function(item) {
-                    if (item.start.line < inlineWidget._startLine || item.end.line > inlineWidget._endLine) {
-                        isInTextRange = false;
-                    }
+                var isInTextRange  = !refs.find(function(item) {
+                    return (item.start.line < inlineWidget._startLine || item.end.line > inlineWidget._endLine);
                 });
+                
                 if (!isInTextRange) {
                     editor.displayErrorMessageAtCursor(Strings.ERROR_RENAME_QUICKEDIT);
                     return;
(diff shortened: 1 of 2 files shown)


src/extensions/default/JavaScriptRefactoring/RenameIdentifier.js
```diff
@@ -100,11 +100,27 @@ define(function (require, exports, module) {
          * If yes then select all references
          */
         function handleFindRefs (refsResp) {
+            var inlineWidget = EditorManager.getFocusedInlineWidget(),
+                editor = EditorManager.getActiveEditor();
+            //In case of inline widget if some references are outside widget's text range then don't allow for rename
+            if (inlineWidget) {
+                var isInTextRange = true;
+                refsResp.references.refs.forEach(function(item) {
```
petetnt: `refsResp.references.refs` might throw. In #L119 it's handled by checking for `if (resResp && ...)`.

I think easiest thing would be null checking it in the start of the function and either returning immediately if it's null and/or saving it to a variable if it is.

```js
function handleFindRefs (refsResp) {
  if (refsResp && refsResp.references && refsResp.references.refs) {
     return;
  }

  var refs = refsResp.references.refs;
  var type = refsResp.references.type;

  ...
}```

**Review by petetnt — CHANGES_REQUESTED** [review]

**Comment by petetnt** [comment]
Cool stuff! Left a small comment.

**Comment by navch** [comment]
Thanks @petetnt  for reviewing this. Sorry for late response. I addressed review comment. Please have a look.

**Review by petetnt — APPROVED** [review]
Codewise LGTM

File: src/extensions/default/JavaScriptRefactoring/RenameIdentifier.js [hunk 2]
```diff
@@ -100,15 +100,36 @@ define(function (require, exports, module) {
...
+            }
+
+            var inlineWidget = EditorManager.getFocusedInlineWidget(),
+                editor = EditorManager.getActiveEditor(),
+                refs = refsResp.references.refs,
+                type = refsResp.references.type;
+
+            //In case of inline widget if some references are outside widget's text range then don't allow for rename
+            if (inlineWidget) {
+                var isInTextRange = true;
```
petetnt: Can we use `Array.prototype.find` on all platforms? Would simplify this to
```js
var isInTextRange  = !refs.find(function(item) {
  return (item.start.line < inlineWidget._startLine || item.end.line > inlineWidget._endLine);
});
```

navch: 👍 Done :)

PR #14005, opened 2017-12-29
Always show full file path #13987

Matt-Rouse:
Fix for #13987 please see comments on issue.


src/editor/EditorOptionHandlers.js
```diff
@@ -96,6 +98,7 @@ define(function (require, exports, module) {
     CommandManager.register(Strings.CMD_TOGGLE_LINE_NUMBERS, Commands.TOGGLE_LINE_NUMBERS, _getToggler(SHOW_LINE_NUMBERS));
     CommandManager.register(Strings.CMD_TOGGLE_ACTIVE_LINE, Commands.TOGGLE_ACTIVE_LINE, _getToggler(STYLE_ACTIVE_LINE));
     CommandManager.register(Strings.CMD_TOGGLE_WORD_WRAP, Commands.TOGGLE_WORD_WRAP, _getToggler(WORD_WRAP));
+    CommandManager.register("Always show full file path", Commands.TOGGLE_ALWAYS_SHOW_FULL_FILE_PATH, _getToggler(ALWAYS_SHOW_FULL_FILE_PATH));
```
petetnt: Should be `Strings.CMD_ALWAYS_SHOW_FULL_FILE_PATH`

src/editor/Editor.js
```diff
@@ -2704,6 +2709,27 @@ define(function (require, exports, module) {
         return PreferencesManager.get(WORD_WRAP, _buildPreferencesContext(fullPath));
     };
 
+    /**
+     * Sets alwaqys show full file path option.
```
petetnt: Nit: `always`

petetnt (commented):
Nice work!

Few comments:

Haven't tested the code myself (yet), but if I understand correctly this can be now set to all the files separately? What would be the scenarios for toggling the setting on/off for individual files, or would one setting that rules them all be enough?

Few unit tests would be nice too 👍

src/editor/Editor.js
steamos: Good catch, fixed! :)

src/editor/EditorOptionHandlers.js
steamos: I was going to ask about that, as I'm not sure about all the translations, shall I just add a placeholder in the ones I'm not sure about?

steamos:
Thanks for your feedback, I'll created some unit tests! 👍

src/editor/EditorOptionHandlers.js
petetnt: Translations fall back to english if they are not specified so it's okay to leave them alone 👍

steamos: Okay brilliant, thanks. 

I'm just finishing off the unit testing so should be done very soon

steamos:
Hi @petetnt 
I'm struggling a little bit with the unit tests, at the moment I have something alone the lines of this:

```
runs(function () {
    var doc = DocumentManager.getCurrentDocument();
    var _fullPath = doc.file.fullPath;
                    
    // turn on always show full path
    PreferencesManager.set("AlwaysShowFullFilePath", true);
                    
    expect(DocumentCommandHandlers._shortTitleForDocument(doc)).toEqual(_fullPath);
});
```

But it appears that `PreferencesManager.set("AlwaysShowFullFilePath", true);` doesn't appear to be doing anything, could you please assist? 

Thanks,
Matt

petetnt:
Not completely sure what's up with that, in the PreferencesManager tests the PreferencesManager is taken from the test window instance, does that help?

https://github.com/adobe/brackets/blob/33be8022e60f1cb50c66244ce71ca20be6a3e487/test/spec/PreferencesManager-test.js#L39-L47

PR #14019, opened 2018-01-08
Fixed some bugs in JS Refactoring

navch:
Fixed some bug in JS Refactoring i.e.

1) Not allowing to rename some keywords
2) If someone tries to convert a named function to arrow function, earlier we used to convert surrounded anonymous function, now if cursor is inside named function then we will show error.
3) Handling default params in case of convert to arrow function

  Please Review @petetnt @swmitra @nethip @boopeshmahendran @saurabh95

PR #14033, opened 2018-01-12
Adding strings related to JS Refactoring Feature

navch:


navch:
Merging these strings first and waiting for review (https://github.com/adobe/brackets/pull/14019) complete. So String translation can be done.

PR #14057, opened 2018-01-22
Code to send Analytics Data to server

NehaModi:


NehaModi:
ping @swmitra  @nethip

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -179,6 +234,7 @@ define(function (require, exports, module) {
 
         HealthLogger.setHealthLogsEnabled(isHDTracking);
         window.clearTimeout(timeoutVar);
+         sendAnalyticsDataToServer("pingData", "", "", "");
```
swmitra: This call is not protected in try catch. If this fails neither health data nor analytics data would be sent periodically.

src/extensions/default/HealthData/HealthDataPreview.js
```diff
@@ -45,7 +45,11 @@ define(function (require, exports, module) {
         var result = new $.Deferred();
 
         HealthDataManager.getHealthData().done(function (healthDataObject) {
-            var content = JSON.stringify(healthDataObject, null, 4);
+            var combinedHealthAnalyticsData = HealthDataManager.getAnalyticsData("pingData", "", "", ""),
+                content;
+            combinedHealthAnalyticsData = [healthDataObject, combinedHealthAnalyticsData ];
+            combinedHealthAnalyticsData.join("\n\n");
```
swmitra: Sorry if I missed anything, but array join returns the joined elements as a string. Where it's received? I doubt `content` would ever be what it's intended to. It will be a serialized array representation and not the joined elements as a string!

src/extensions/default/HealthData/HealthDataPreview.js
```diff
@@ -45,7 +45,11 @@ define(function (require, exports, module) {
...
+            content = JSON.stringify(combinedHealthAnalyticsData, null, 4);
```
swmitra: By joining the elements in the previous statement we could have produced a string already. Why are we serializing it again?

swmitra (changes requested):

swmitra:
Please sign brackets cla, build has failed while doing cla check pull.

src/extensions/default/HealthData/HealthDataManager.js
petetnt: Brackets also ships with the modern `fetch` api now, so we could use that:

```js
var healthData = getAnalyticsData(category, subcategory, type, subtype);

return fetch(brackets.config.analyticsDataServerURL, {
  method: 'POST', // or 'PUT'
  body: JSON.stringify({events: [healthData]}), 
  headers: new Headers({
    'Content-Type': 'application/json',
    'x-api-key': brackets.config.projectName
  })
});

...

sendAnalyticsDataToServer("pingData", "", "", "").catch(console.err);
```

src/config.json
```diff
@@ -20,7 +20,9 @@
         "extension_url": "https://s3.amazonaws.com/extend.brackets/{0}/{0}-{1}.zip",
         "linting.enabled_by_default": true,
         "build_timestamp": "",
-        "healthDataServerURL": "https://healthdev.brackets.io/healthDataLog"
+        "healthDataServerURL": "https://health.brackets.io/healthDataLog",
+        "analyticsDataServerURL" : "https://cc-api-data-stage.adobe.io/ingest/",
```
nethip: Looks like this is still referring to stage server. I think you need to create a grunt task to point to the actual server. Also don't change `healthDataServerURL` from stage to prod as the grunt job takes care of it and the `analyticsDataServerURL` should be doing the same thing.

src/extensions/default/HealthData/HealthDataPreview.js
NehaModi: This is addressed in current commit

src/extensions/default/HealthData/HealthDataManager.js
NehaModi: @swmitra  This is not required  in the code,this is added just for testing

src/extensions/default/HealthData/HealthDataPreview.js
NehaModi: @swmitra  updated the code with the changes

src/config.json
NehaModi: I have added the stage and prod URL and is available in current commit

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -121,16 +119,61 @@ define(function (require, exports, module) {
...
+    //Get Analytics data
+    
+    function getAnalyticsData(category, subcategory, type, subtype) {
+        var userUuid = PreferencesManager.getViewState("UUID"),
+            olderUuid = PreferencesManager.getViewState("OlderUUID"),
+            isoDate = new Date().toISOString(),
+            userAgent = window.navigator.userAgent,
+            isNeedToSetViewState = false;
+        if (!userUuid) {
+            userUuid = uuid.v4();
```
swmitra: We will end up setting wrong UUID as we are not using the macHash and this function is getting executed before sendHealthData call.

swmitra (dismissed):

src/extensions/default/HealthData/HealthDataManager.js
NehaModi: Now I am calling it once we have send health data server

NehaModi: Changed with current commit

swmitra: the following piece of code is not required in that case - 
``` javascript
if (isNeedToSetViewState) {
            PreferencesManager.setViewState("UUID", userUuid);
            PreferencesManager.setViewState("OlderUUID", olderUuid);
        }
```

swmitra:
@nethip Refactored the code to some extent. Will execute some tests in the morning.

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -121,16 +120,41 @@ define(function (require, exports, module) {
...
     }
+    
+    // Get Analytics data
+    function getAnalyticsData() {
+        var userUuid = PreferencesManager.getViewState("UUID"),
+            olderUuid = PreferencesManager.getViewState("OlderUUID");
+
+        return {
+            project: brackets.config.serviceKey,
+            environment: "production",
```
nethip: We might have to externalize this as when posting to stage this is going to be `environment: "stage"`

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -191,12 +241,11 @@ define(function (require, exports, module) {
...
 
-            if (currentTime >= nextTimeToSend) {
-                // Bump up nextHealthDataSendTime now to avoid any chance of sending data again before 24 hours, e.g. if the server request fails
-                // or the code below crashes
+            if (currentTime >= nextTimeToSend || forceSend) {
+                // Bump up nextHealthDataSendTime at the begining of chaining to avoid any chance of sending data again before 24 hours, // e.g. if the server request fails or the code below crashes
                 PreferencesManager.setViewState("nextHealthDataSendTime", currentTime + ONE_DAY);
-
-                sendHealthDataToServer()
+                sendAnalyticsDataToServer().always(function() {
```
nethip: You might want to put this under `done` of `sendHealthDataToServer`

PR #14140, opened 2018-03-01
Add 'onHighlight' optional event callback mechanism for better extensibility features in hint providers

swmitra:
This PR is targeted to add an optional `onHighlight` hook handling for Code hint providers. This will help extension developers to add interactive workflows when a user crawls over hint items in hint popup. 

@nethip @petetnt All yours now, please have a look at the change set.

### DIFF
--- a/src/editor/CodeHintList.js
+++ b/src/editor/CodeHintList.js
@@ -150,6 +150,10 @@ define(function (require, exports, module) {
 
             $item.find("a").addClass("highlight");
             ViewUtils.scrollElementIntoView($view, $item, false);
+
+            if (this.handleHighlight) {
+                this.handleHighlight($item.find("a"));
+            }
         }
     };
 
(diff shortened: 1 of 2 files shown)


petetnt (commented):
Not sure if this should be a Event Emitter instead?

Documentation would be nice to the userguide.

swmitra:
Thanks a lot @petetnt for quick review :+1.

This would be a provider specific implementation. IMHO, scoping the event emitter to a provider will make the listener code ambiguous. Also, keeping this as a hook keeps the reach of the handling very shallow and declaration of the intent is upfront. But that's my thought, want to hear the advantages of event emitter in this case as well.

nethip:
@swmitra LGTM. But +1 on @petetnt comment about documentation.

petetnt:
FWIW I am okay with this _not_ being an event emitter too as code hint list providers provide functions anyway now that I looked at this more 👍

swmitra:
@nethip @petetnt Added this to our wiki for code hint API's. 
https://github.com/adobe/brackets/wiki/New-Code-Hinting-API-Proposal

nethip:
Thanks @swmitra

swmitra:
Thanks @nethip 😃

PR #14143, opened 2018-03-03
resolves #14126, add preference option to show search panel at  bottom

### LINKED ISSUE #14126: Move search panel at bottom
We need a extension that is capable to move Search Pannel at bottom .
Same like the extension that move git pannel at bottom .
Somebody can make this ?

ad1992:


navch:
@ad1992 Can you sign the CLA located at http://dev.brackets.io/brackets-contributor-license-agreement.html?

ad1992:
done @navch . The tests are passing now.

navch:
@ad1992 I think there is one problem. Please try opening Quick Open (Ctrl + Shift + O). Or click on drop down icon in Search box. This might not be the good experience.

image

ad1992:
@navch I have fixed the above issue. Can you please check now.

navch:
@ad1992 Cool Stuff. Working like a charm 👍

package.json
```diff
@@ -73,4 +73,4 @@
             "url": "https://github.com/adobe/brackets/blob/master/LICENSE"
         }
     ]
-}
\ No newline at end of file
+}
```
navch: May be we can revert this file.

swmitra:
@ad1992 Great work 👍 

In Brackets we show all document commands UI in a top anchored modal. That's how most of our users are familiar in seeing such functionalities. Although, with a preference this looks acceptable, but then for users, who wants to see such functionalities in a bottom panel should see every document command UI as bottom panels. So in order to make this a reality, I guess the whole UI paradigm has to be configurable via preference instead of a chosen few. 

If you are ready to make such widespread changes with supporting tests, it can be part of the core. But if it's targeted to one or two of such functionalities, I guess it would server better as an extension. Brackets DOM is extendible enough to be customized from an extension by just using css/less. 

Please let us know if you need any such pointers.

ad1992:
Thank you @navch @swmitra  for looking into this. I had actually discussed with @petent about the approach before creating the PR. I agree that for better user experience all document command UI should be shown as bottom panels. Since this would take some time so I can build an extension for solving the issue referenced in the PR.  If this feature helps then I would love to work on it as a core and making changes across all document command UI as well. But may be should we wait for such request from the users and then start working on core ? what do you suggest ?

PR #14153, opened 2018-03-10
Enable opening of remote(http|https) files in Brackets

swmitra:
This PR adds File protocol adoption model to support various file protocols which mainly differs in read and write mechanism. With this PR we are adding a reference implementation for this protocol adapter model by enabling `HTTP|HTTPS` protocols. We are introducing a new protocol adapter registration hook in `FileSystem` module as `function registerProtocolAdapter(protocol, adapter)` where `protocol` can be any valid file protocol string like 

> HTTP: | HTTPS: | FTP: | SFTP: 

 for which we want to add support. `adapter` is the the actual protocol adapter object which should expose `canRead(filePath)` function handle and `fileImpl` which is the `File` interface implementation prototype handle, this will be be used by `FileSystem` to instantiate this new type of `File` if the protocol adapter is selected based on the `filePath`. 

This change list also adds a plugin in `QuickOpen`, enabling user to input remote file path (`HTTP | HTTPS `) in QuickOpen file search dialog to open a file. 

TODO
------

- [x] Unit tests
- [x] Ux workflow
- [x] List remote files in Quick open file search list

### DIFF
--- a/src/extensions/default/RemoteFileAdapter/RemoteFile.js
+++ b/src/extensions/default/RemoteFileAdapter/RemoteFile.js
@@ -56,7 +56,7 @@ define(function (require, exports, module) {
      * @param {!string} fullPath The full path for this File.
      * @param {!FileSystem} fileSystem The file system associated with this File.
      */
-    function RemoteFile(fullPath, fileSystem) {
+    function RemoteFile(protocol, fullPath, fileSystem) {
         this._isFile = true;
         this._isDirectory = false;
         this._path = fullPath;
@@ -65,6 +65,8 @@ define(function (require, exports, module) {
         this._name = fullPath.split('/').pop();
         this._fileSystem = fileSystem;
         this.donotWatch = true;
+        this.protocol = protocol;
+        this.encodedPath = fullPath;
     }
 
     // Add "fullPath", "name", "parent", "id", "isFile" and "isDirectory" getters
@@ -103,7 +105,21 @@ define(function (require, exports, module) {
      * Helpful toString for debugging and equality check purposes
      */
     RemoteFile.prototype.toString = function () {
-        return "[" + (this._isDirectory ? "Directory " : "RemoteFile ") + this._path + "]";
+        return "[RemoteFile " + this._path + "]";
+    };
+
+    /**
+     * Returns the stats for the remote entry.
+     *
+     * @param {function (?string, FileSystemStats=)} callback Callback with a
+     *      FileSystemError string or FileSystemStats object.
+     */
+    RemoteFile.prototype.stat = function (callback) {
+        if (this._stat) {
+            callback(null, this._stat);
+        } else {
+            callback(FileSystemError.NOT_FOUND);
+        }
     };
 
     RemoteFile.prototype.constructor = RemoteFile;
--- a/src/extensions/default/RemoteFileAdapter/RemoteFile.js
+++ b/src/extensions/default/RemoteFileAdapter/RemoteFile.js
@@ -99,15 +99,11 @@ define(function (require, exports, module) {
         }
     });
 
-    RemoteFile.prototype.exists = function (callback) {
-        callback(null, true);
-    };
-
     /**
      * Helpful toString for debugging and equality check purposes
      */
     RemoteFile.prototype.toString = function () {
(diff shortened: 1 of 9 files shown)


vickramdhawal:
How are the users supposed to trigger opening of a remote file ? Is there any UI associated with it ?
Currently, users can open files only from the mounted file system.

swmitra:
@vickramdhawal It can be done by file open command.
```javascript
CommandManager.execute(Commands.FILE_OPEN, {
    fullPath: "<remote-file-uri here>"
});
```

abose:
"protocol can be any valid file protocol string like HTTP: | HTTPS: | FTP: | SFTP:"
Is FTP and SFTP supported in this change?

abose:
Also would be good if this could be hooked into some UI. maybe the quick open cmd+o hotkey.
This feature would help to *adapt* code from the internets.

swmitra:
@abose This PR provides implementation only for `http/https`, but a plugin framework is also being added in this PR, where other extensions can hook onto and provide custom `read/write` handles, or add support for a new protocol.

Let me see if this can be hooked to quickopen where user can type/paste a remote path to open  a file.

abose:
Now that quick open works on http/s URL , normal file[file:///c/a.txt or c:\a.txt etc] paths cannot be opened with the same. Would be good if this behaved consistently for file path open as well.

boopeshmahendran (approved):

src/extensions/default/RemoteFileAdapter/main.js
```diff
@@ -0,0 +1,70 @@
...
+        PathUtils       = brackets.getModule("thirdparty/path-utils/path-utils"),
+        CommandManager  = brackets.getModule("command/CommandManager"),
+        Commands        = brackets.getModule("command/Commands"),
+        RemoteFile      = require("RemoteFile");
+
+    var HTTP_PROTOCOL = "http:",
+        HTTPS_PROTOCOL = "https:";
+
+    AppInit.htmlReady(function () {
+        var protocolAdapter = {
```
vickramdhawal: Since this is an extension shouldn't the definition of ProtocolAdapter be provided by Brackets-Core for consistency.

swmitra: @vickramdhawal Sorry for the confusion. `protocolAdapter` is just a variable to hold the custom imprint of the adapter definition to be passed as protocol adapters for `HTTP:` and `HTTPS:`. Registration happens by invoking this part -
 ```javascript
// Register the custom object as HTTP and HTTPS protocol adapter
FileSystem.registerProtocolAdapter(HTTP_PROTOCOL, protocolAdapter);
FileSystem.registerProtocolAdapter(HTTPS_PROTOCOL, protocolAdapter);
```
I can't really define an interface (which if was possible, would have been part of core) and `Duck Typing` would be a gimmick as type checking is not something what we are interested about here.

File: src/extensions/default/RemoteFileAdapter/RemoteFile.js [hunk 2]
```diff
@@ -0,0 +1,209 @@
...
+
+define(function (require, exports, module) {
+    "use strict";
+
+    var FileSystemError = brackets.getModule("filesystem/FileSystemError"),
+        FileSystemStats = brackets.getModule("filesystem/FileSystemStats");
+
+    var SESSION_START_TIME = new Date();
+
+    /*
```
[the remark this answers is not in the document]
swmitra: Addressed in latest commit 👍

abose:
LGTM.

src/extensions/default/RemoteFileAdapter/RemoteFile.js
```diff
@@ -99,15 +99,11 @@ define(function (require, exports, module) {
...
-    RemoteFile.prototype.exists = function (callback) {
-        callback(null, true);
-    };
-
     /**
      * Helpful toString for debugging and equality check purposes
      */
     RemoteFile.prototype.toString = function () {
-        return "[" + (this._isDirectory ? "Directory " : "File ") + this._path + "]";
+        return "[" + (this._isDirectory ? "Directory " : "RemoteFile ") + this._path + "]";
```
vickramdhawal: Can a remote URI be described as a directory ? Looks weird to me.

navch (approved):

PR #14168, opened 2018-03-22
Bringing in ImmutableJS 3.8.2 

vickramdhawal:
Bringing in ImmutableJS 3.8.2 which is release under MIT license. Source of immutableJS https://github.com/facebook/immutable-js/releases/tag/v3.8.2.
Also updating the gitignore file to exclude node_modules inside src/JSUtils

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@ package-lock.json
 
 # ignore node_modules inside src
 /src/node_modules
+/src/JSUtils/node_modules
+/src/JSUtils/node/node_modules
 
 # ignore files copied from node_modules to src/thirdparty
 /src/thirdparty/CodeMirror
--- a/src/thirdparty/immutable.js
+++ b/src/thirdparty/immutable.js
@@ -1,16 +1,14 @@
 /**
- *  Copyright (c) 2014-2015, Facebook, Inc.
- *  All rights reserved.
+ * Copyright (c) 2014-present, Facebook, Inc.
  *
- *  This source code is licensed under the BSD-style license found in the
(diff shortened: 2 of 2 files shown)


nethip:
Thanks for the PR @vickramdhawal :+1: 

Did you get a chance to run unit tests?

nethip:
Also looks like CLA has not been signed that is why the travis tests are failing. Can you signup using this link? Please sign the CLA here http://dev.brackets.io/brackets-contributor-license-agreement.html.

vickramdhawal:
I've signed up now. I did run the unit tests, but have to bear with multiple dialogs to get it to completion. Will update once its done.

nethip:
Thanks @vickramdhawal I have restarted the build.

nethip:
You can disable the dialog popups by returning from the below function. We will have to figure out other means of handling the dialog popups.
https://github.com/adobe/brackets/blob/master/src/document/DocumentCommandHandlers.js#L1727

vickramdhawal:
immutableJS is used only by the FileTreeView & the FileTreeViewModel and the tests are passing for both. Unit tests are also passing except for the ones already failing on master.

swmitra:
Good work @vickramdhawal.

swmitra (approved):

.gitignore
```diff
@@ -12,6 +12,8 @@ package-lock.json
 
 # ignore node_modules inside src
 /src/node_modules
+/src/JSUtils/node_modules
+/src/JSUtils/node/node_modules
```
ficristo: IMO this fixes the effects and not the cause of the problem.

PR #13964, opened 2017-12-07
Don't prompt before unload while in SpecRunner

petetnt:
Fixes #13963

Tested manually by:

- Opening test runner and navigating between the tabs and confirming that the prompts do not pop up anymore.

- Opening dev tools and trying `window.location.pathname = 'https://google.com'` to see that the prompt still pops up

Signed-off-by: petetnt <pete.a.nykanen@gmail.com>

PR #14169, opened 2018-03-22
Fix to avoid window location change confirmation dialog while running tests - onbeforeunload conditional registration

swmitra:
This PR adds `testEnvironment` search param validation to skip `onbeforeunload` registration if the current brackets window is a test window.

@nethip @petetnt Please review.

@vickramdhawal You can use this patch to run tests 😄

### DIFF
--- a/src/document/DocumentCommandHandlers.js
+++ b/src/document/DocumentCommandHandlers.js
@@ -1728,10 +1728,6 @@ define(function (require, exports, module) {
     var isTestWindow = (new window.URLSearchParams(window.location.search || "")).get("testEnvironment");
     if (!isTestWindow) {
         window.onbeforeunload = function(e) {
-            if (window.location.pathname.indexOf('SpecRunner') > -1) {
-                return;
-            }
-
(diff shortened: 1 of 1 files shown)


petetnt (approved):
LGTM, though I thought that https://github.com/adobe/brackets/pull/13964 would have handled this 🤔

swmitra:
@petetnt Thanks a lot for the quick review 👍 
Main problem was integration tests where the actual Brackets window navigations were giving the same problem. In order to fix all those issues we needed to differentiate between actual Brackets instances and instances launched for test.

nethip:
Thanks for the PR @swmitra. This solves the spec runner case! Is there a possibility to handle the reload case as well?

swmitra:
Thanks for reviewing the PR @nethip 👍 
Let me try to fix the Brackets reload scenario.

swmitra:
@nethip While trying to fix this problem in case of Brackets reload, noticed that, it can't be reproduced 😄

swmitra:
Thank you @nethip 😃

Issue #14163, opened 2018-03-17
Replace function with regex remove text on big file


zouglou:
I had a file with multiple email addresses (like 38 000), one by line. I wanted to replace every line break by a comma and a space, « \n » by « ,  ».
So I've tried with Find and replace function, it works doing one by one, but using "Replace everything" it replaces everything by a lot of comma and spaces, addresses removed.

No plugins,  v1.12.0-17621, Ubuntu 64bits.

navch:
Hi @zouglou 

Thanks For reporting this issue. I am able to repro this issue. This is happening only in case of the new line. We are looking into this.

PR #14175, opened 2018-03-27
Fixing issue 14163, F&R replaces all content of file when replacing new line

navch:
Edge case for "\n". In this case, the value of ch was -1 because of match.index value and value returned by lastIndexOf is same.

Issue https://github.com/adobe/brackets/issues/14163

Please review @swmitra @nethip


sobisht (commented):
LGTM

sobisht (approved):

swmitra (approved):

PR #14177, opened 2018-03-27
AutoUpdate Framework

mbhavi:
This PR includes changes for Auto Update of Brackets.
The changes are with respect to:
1) Windows implementation
2) Mac implementation
3) A new UI

ping @swmitra @nethip for review

The Auto Update workflow along with the screenshots is described below : 

- **Notification of Update** : through current UpdateNotification dialog, which is triggered either from Help->CheckForUpdates or from the side bar button notification

image

- **Download of Update** : Once user clicks on GetItNow button, a download of the latest installer is triggered in the background, with a progress bar shown on the status bar.

image

- **Download retrial in case of failure** : If the download fails, it is retried for a maximum of 5 attempts, and the same is reflected on the status bar UI.

image

- **Validation of the downloaded update** : Once download is successful, the SHA256 checksum is validated, and the info is displayed on status bar.

image

- **Download Failure** : In case download fails, error is shown to the user towards bottom of the editor.

image

- **Download and Validation success** : User is prompted with the update bar, with buttons for "Restart" and "Later" options.  Clicking Restart button would start the update procedure by quitting the app, and launching the installer. Clicking on Later button allows user to continue with the normal flow of Brackets.

image

- **Validation Failure** : If the validation of the downloaded update fails, a validation error is displayed to the user in update bar.

image

- **Update in Progress** : Once the user clicks Restart button, the app is quit and installer gets launched.

image

- **Update Success** : If update is successful, new version of Brackets is launched automatically, with the success message on the update bar.

image

- **Update Failure** : If the update fails, the older version of Brackets is relaunched automatically, and the error message is displayed to the user in update bar, with a link to the brackets website for a manual update operation.

image

Any **warnings** or errors encountered during the worflow are reflected in the update bar.
Screenshots for one of the warnings is attached here. It corresponds to the scenario where user clicked on Restart button, but had dirty files to which he/she chose Cancel in File Save dialog. In this case, the update will happen on the next quit of Brackets, and the same is displayed to the user as a warning on update bar.

image

On MAC, instead of the installer, the update procedure is carried out by an update shell script. Rest everything is same for MAC, as above.


abose:
Awesome! @mbhavi Can you add some screenshots of the new flow?

abose:
Is there a preference to disable auto updates?

abose:
Sometimes, users disable auto-updates to prevent breaking changes to their setup[extension compatibilities Et.Al.]. It would be good if it is configurable by a preference without having to remove the extension. 
Also is there an opt-in at first launch? or is the update process kicked off without confirmation from the user?
@petetnt @swmitra

mbhavi:
@abose, I have attached the screenshots and the step-by-step workflow description.
The update is kicked off only after the confirmation from the user as an input of GetItNow button click in UpdateNotification dialog.

src/document/DocumentCommandHandlers.js
```diff
@@ -142,6 +142,17 @@ define(function (require, exports, module) {
...
     var handleFileSaveAs;
 
+    /**
+    Event triggerred when Auto update is in progress, and
+    user clicks Cancel in File Save prompt, after clicking UpdateNow
+    */
+    var DIRTY_FILESAVE_CANCELLED = "dirtyFileSaveCancelled";
+
+    /**
+    Event triggered when error is returned when tried to check if Auto Update is progress
```
nethip: nit: `trying` instead of tried.

src/document/DocumentCommandHandlers.js
```diff
@@ -848,6 +859,19 @@ define(function (require, exports, module) {
         return result.promise();
     }
 
+    /**
+     * Checks and handles if auto update is currently in progress
+     */
+    function checkIfAutoUpdateInProgress() {
+        brackets.app.isAutoUpdateInProgress(function (err, isAutoUpdateInProgress) {
```
nethip: Add `if (brackets.app.isAutoUpdateInProgress) {` to check if this function is registered or not.

src/document/DocumentCommandHandlers.js
```diff
@@ -848,6 +859,19 @@ define(function (require, exports, module) {
...
+            if (err === 0 && isAutoUpdateInProgress) {
```
nethip: No need for an explicit check, general rule we follow in `Brackets` is to check for error condition first. So your statements should be

```

if (err) {
 
} else (isAutoUpdateInProgress) {

}

```

src/document/DocumentCommandHandlers.js
```diff
@@ -848,6 +859,19 @@ define(function (require, exports, module) {
...
+                CommandManager.trigger(exports.DIRTY_FILESAVE_CANCELLED);
```
nethip: Use `execute` over `trigger`.

nethip: Help me understand. The name of the method is `checkIfAutoUpdateInProgress` but you are executing `exports.DIRTY_FILESAVE_CANCELLED` ? Shouldn't this be something like whoever is calling this method, should be returned a promise and the client can then act upon the promise state.

src/document/DocumentCommandHandlers.js
```diff
@@ -996,6 +1020,7 @@ define(function (require, exports, module) {
                     if (selectedPath) {
                         _doSaveAfterSaveDialog(selectedPath);
                     } else {
+                        checkIfAutoUpdateInProgress();
```
nethip: Could you explain (workflow) why this call is required here?

src/extensions/default/AutoUpdate/StateHandler.js
```diff
@@ -0,0 +1,205 @@
...
+     * @returns {$.Deferred} - a jquery deferred promise,
+     *                       that is resolved with existence or non-existence
+     *                       of json file.
+     */
+    StateHandler.prototype.exists = function () {
+        var result = $.Deferred(),
+            _file = FileSystem.getFileForPath(this.path);
+
+        _file.exists(function (err, exists) {
+            if (exists) {
```
nethip: Check for `err`.

src/document/DocumentCommandHandlers.js
mbhavi: This is needed for the workflow where user clicked Restart button for the update to happen, but had some dirty files, because of which the app could not be quit, and instead, he got the Unsaved Files dialog, This dialog has 3 buttons : Save/Don't Save/Cancel. If Don't Save is pressed, we quit the app and run the update. If Save is pressed and user saves the file successfully, then also we quit the app and update. But when the user chooses to either press Cancel button in the above dialog, or presses Save followed by Cancel in the Save As dialog, we show him a message saying that the update will now be reflected on relaunch, which inherently means that the update will now be carried out the next time the app quits.
So, for this, we need to know whether or not the save file was cancelled by either of the two scenarios mentioned here, so that we can prompt him with the required message.

src/extensions/default/AutoUpdate/StateHandler.js
```diff
@@ -0,0 +1,205 @@
...
+    /**
+     * Performs the write of JSON object to a file.
+     * @param   {string} filepath - path to JSON file
+     * @param   {object} json  - JSON object to write
+     * @returns {$.Deferred} - a jquery deferred promise,
+     *                       that is resolved with the write success or failure
+     */
+    function _write(filePath, json) {
+        var result = $.Deferred(),
+            _file = FileSystem.getFileForPath(filePath);
```
nethip: check for validity of `_file`

src/extensions/default/AutoUpdate/UpdateInfoBar.js
```diff
@@ -0,0 +1,167 @@
...
+        //Event handlers on the Update Bar
+
+        if($restart.length > 0) {
+            $restart.click(function () {
+                cleanUpdateBar();
+                MainViewManager.trigger(exports.RESTART_BTN_CLICKED);
+            });
+
+            $restart.keyup(function (e) {
+                if (e.which === 32) { //32 is the keycode for space key
```
nethip: use a named variable for better readability.

src/extensions/default/AutoUpdate/main.js
```diff
@@ -0,0 +1,608 @@
...
+    var _modulePath = FileUtils.getNativeModuleDirectoryPath(module),
+        _nodePath = "node/AutoUpdateDomain",
+        _domainPath = [_modulePath, _nodePath].join("/"),
+        updateDomain = new NodeDomain("AutoUpdate", _domainPath);
+
+    var appSupportDirectory = brackets.app.getApplicationSupportDirectory(),
+        updateDir = appSupportDirectory + '/updateTemp',
+        updateJsonPath = updateDir + '/' + 'updateHelper.json';
+
+    var updateJsonHandler = new StateHandler.StateHandler(updateJsonPath);
```
nethip: StateHandler.StateHandler?

src/extensions/default/AutoUpdate/main.js
```diff
@@ -0,0 +1,608 @@
...
+        functionMap["brackets.showErrorMessage"]             = showErrorMessage;
+        functionMap["brackets.notifyDownloadFailure"]        = handleDownloadFailure;
+        functionMap["brackets.notifySafeToDownload"]         = handleSafeToDownload;
+        functionMap["brackets.notifyvalidationStatus"]       = handleValidationStatus;
+    }
+
+    functionMap["brackets.registerBracketsFunctions"] = registerBracketsFunctions;
+
+    UpdateNotification.on(UpdateNotification.GET_AUTOUPDATE_INSTALLER, handleGetNow);
+    CommandManager.on(DocumentCommandHandlers.DIRTY_FILESAVE_CANCELLED, dirtyFileSaveCancelled);
```
nethip: Refer to the comment that I put above in `DocumentCommandHandler` about generic command handlers.

src/extensions/default/AutoUpdate/main.js
```diff
@@ -0,0 +1,608 @@
...
+    CommandManager.on(DocumentCommandHandlers.AUTOUPDATE_ERROR, handleAutoUpdateError);
+
+    updateDomain.exec('initNode', {
+        messageIds: MessageIds,
+        updateDir: updateDir
+    });
+    updateDomain.on('data', receiveMessageFromNode);
+
+    AppInit.appReady(function () {
```
nethip: Check for platform and only then initialize AppInit.appReady as auto update is targeted for only Win and Mac.

src/extensions/default/AutoUpdate/main.js
```diff
@@ -0,0 +1,608 @@
...
+        FileUtils               = brackets.getModule("file/FileUtils"),
+        Strings                 = brackets.getModule("strings"),
+        StateHandler            = require("StateHandler"),
+        DocumentCommandHandlers = brackets.getModule("document/DocumentCommandHandlers"),
+        MessageIds              = require("MessageIds"),
+        UpdateStatus            = require("UpdateStatus"),
+        UpdateInfoBar           = require("UpdateInfoBar");
+
+
+    var _modulePath = FileUtils.getNativeModuleDirectoryPath(module),
```
nethip: Execute the steps only for `Win` and `Mac`

src/extensions/default/AutoUpdate/node/AutoUpdateDomain.js
```diff
@@ -0,0 +1,380 @@
...
+                {
+                    name: "msgObj",
+                    type: "object",
+                    description: "json object containing message info"
+                }
+            ],
+            []
+        );
+
+        domainManager.registerEvent(
```
nethip: Empty function?

src/htmlContent/update-dialog.html
```diff
@@ -1,17 +1,17 @@
 <div class="update-dialog modal">
     <div class="modal-header">
+        <img class="update-icon" src="styles/images/update_large_icon.svg">
```
nethip: Check the validity of the changed HTML on `linux`.

src/utils/UpdateNotification.js
```diff
@@ -31,6 +31,7 @@ define(function (require, exports, module) {
     var Dialogs              = require("widgets/Dialogs"),
```
nethip: Can all the changes made here be moved to the extension?

nethip:
I have reviewed this PR . Please look at the comments and incorporated the changes wherever required.

Also make sure that for `linux` no new code, added by this PR, is getting called. Make sure the new update workflow is for `win` and `mac` only and make changes appropriately, either in `HTML` or in the extension code.

Also it is a good idea to have a key to disable this feature. Please add a preference to disable auto update.

src/document/DocumentCommandHandlers.js
nethip: Thanks for explanation @mbhavi

src/document/DocumentCommandHandlers.js
mbhavi: isAutoUpdateInProgress() function has been added to the core. It is not a registered function. Hence, we need not have this check.

src/htmlContent/update-dialog.html
mbhavi: @ayushrathi15 , please validate this in the linux builds.

src/utils/UpdateNotification.js
mbhavi: The changes made here for AutoUpdate feature heavily use the update information like version info, URLs etc., which is maintained in this file. Hence, the changes should not be moved to the extension.

src/extensions/default/AutoUpdate/main.js
mbhavi: Addressed this in 537430b

src/document/DocumentCommandHandlers.js
mbhavi: Addressed this in 537430b

src/document/DocumentCommandHandlers.js
mbhavi: Addressed this in 537430b

src/document/DocumentCommandHandlers.js
mbhavi: I have now placed all the events related to Auto Update, needed in core, in UpdateNotification in order to ensure consistency.
Also, I now chose a better name for checkIfAutoUpdateProgress(). It is actually a handler function. It iinternally calls a function isAutoUpdateInProgress(), which returns a bool, and the wrapper function handles this bool. So, I have renamed the wrapper function to handleIfAutoUpdateInProgress(). 
Please refer to 537430b

src/document/DocumentCommandHandlers.js
```diff
@@ -1217,6 +1242,7 @@ define(function (require, exports, module) {
             )
                 .done(function (id) {
                     if (id === Dialogs.DIALOG_BTN_CANCEL) {
+                        checkIfAutoUpdateInProgress();
```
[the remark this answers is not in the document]
mbhavi: I am indeed sending the notification of a Cancel click on dirty file save operation. Please refer to my comments above.

src/extensions/default/AutoUpdate/StateHandler.js
mbhavi: Addressed in 537430b

src/extensions/default/AutoUpdate/UpdateInfoBar.js
mbhavi: Addressed in  537430b

src/extensions/default/AutoUpdate/main.js
mbhavi: Addressed in  537430b

src/extensions/default/AutoUpdate/main.js
```diff
@@ -0,0 +1,608 @@
...
+     * Enables/disables the state of "Check For Updates" menu entry under Help Menu
+     */
+    function enableCheckForUpdateEntry(enable) {
+        var cfuCommand = CommandManager.get(Commands.HELP_CHECK_FOR_UPDATE);
+        cfuCommand.setEnabled(enable);
+    }
+
+    /**
+     * Checks if it is the first iteration of download
+     * @returns {boolean} - true if first iteration, false if is a retrial of download
```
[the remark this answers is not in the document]
mbhavi: Addressed in  537430b

src/extensions/default/AutoUpdate/main.js
```diff
@@ -0,0 +1,608 @@
...
+
+    /**
+     * Sets the update state in updateHelper.json in Appdata
+     * @param {string} key   - key to be set
+     * @param {string} value  - value to be set
+     * @param {function} callback - the callback function
+     *                            to be called in case of
+     *                            successful setting of update state
+     */
+    function setUpdateStateInJSON(key, value, callback) {
```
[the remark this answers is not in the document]
mbhavi: This is indeed a promise based function. The same promise handling code snippet was getting used at multiple places, and so, in order to avoid the repetition of code, I have templatized the code snippet.

src/extensions/default/AutoUpdate/main.js
```diff
@@ -0,0 +1,608 @@
...
+    }
+
+    /**
+     * Handles a safe download of the latest installer,
+     * safety is ensured by cleaning up any pre-existing installers
+     * from update directory before beginning a fresh download
+     */
+    function handleSafeToDownload() {
+
+        var downloadCB = function() {
```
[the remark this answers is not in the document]
mbhavi: Please refer to my comment above.

src/extensions/default/AutoUpdate/main.js
mbhavi: Please refer to my comments in DocumentCommandHandlers and UpdateNotification, where I have explained the use cases and requirements.

src/extensions/default/AutoUpdate/main.js
mbhavi: Addressed in  537430b

src/extensions/default/AutoUpdate/node/AutoUpdateDomain.js
mbhavi: This is actually an event registration, for which no naming is required.

mbhavi:
@nethip , I have addressed and committed all the required changes. Also, I have replied to all the comments.
The PR is ready for review again.

nethip:
@mbhavi I think the preference to disable is still pending.

nethip:
@mbhavi I don't think the changes which I had proposed with `DocumentCommandHandler.js` are incorporated. 

There should be no reference of `AutoUpdate` anywhere inside `DocumentCommandHandler.js`. You could still borrow the mechanism but no reference to auto update. I still see methods like `handleIfAutoUpdateInProgress` or  in the PR. I am not sure if there is any any strong reason for hard wiring auto update into `DocumentCommandHandler.js`, as you could create new events and emit those events from here. The `AutoUpdate` extension can then register for those events and react accordingly..

src/extensions/default/AutoUpdate/StateHandler.js
nethip: I think there is not harm in checking for validity of `_file` given the fact that you are going to `stringify` the JSON irrespective of whether `_file` is valid or not.

src/extensions/default/AutoUpdate/main.js
```diff
@@ -0,0 +1,630 @@
...
+        functionMap["brackets.notifyvalidationStatus"]       = handleValidationStatus;
+    }
+
+    functionMap["brackets.registerBracketsFunctions"] = registerBracketsFunctions;
+
+    UpdateNotification.on(UpdateNotification.GET_AUTOUPDATE_INSTALLER, handleGetNow);
+    UpdateNotification.on(UpdateNotification.DIRTY_FILESAVE_CANCELLED, dirtyFileSaveCancelled);
+    UpdateNotification.on(UpdateNotification.AUTOUPDATE_ERROR, handleAutoUpdateError);
+
+    if(brackets.platform !== "linux") {
```
nethip: How about checking for the shell functions existence like `setUpdateParamsAndQuit`, and if not present, just bail out?

src/extensions/default/AutoUpdate/node/AutoUpdateDomain.js
nethip: Got it! Thanks

src/document/DocumentCommandHandlers.js
mbhavi: I strongly believe suppressing the error would turn out to be harmful in this case, not showing the error to the user would be misleading to him/her believing that the feature can work with just the brackets changes. 
Moreover, I don't see that with any of the other functions' invocations. I believe doing that will be unnecessary bloating of the code, as this is most likely to happen only in the dev machines, which I think should be okay.

src/utils/UpdateNotification.js
```diff
@@ -333,6 +447,12 @@ define(function (require, exports, module) {
         var result = new $.Deferred();
         var versionInfoUrl;
 
+        //AUTOUPDATE_UNITTESTING
+        var updateJSONLocalhostURL = PreferencesManager.get("updateJSONLocalhostURL");
```
nethip: This needs to be removed.

src/utils/UpdateNotification.js
nethip: @mbhavi One of the main reasons, we move most of the new code to an extension is to leverage pluggability model. So imagine, `AutoUpdate` extension is removed from the list of default extensions, the core should function fine and preferably with other functionality, which in this case is just open the `brackets.io` page. You need to follow something like what @swmitra has done with https://github.com/adobe/brackets/pull/14140. Have a default implementation, which an extension can override and provide implementation on how to get the latest update.

src/utils/UpdateNotification.js
```diff
@@ -389,6 +509,16 @@ define(function (require, exports, module) {
 
                         // Update prefs with the last notified build number
                         lastNotifiedBuildNumber = allUpdates[0].buildNumber;
+
+                        //If no checksum field is present then we're setting it to 0, just as a safety check,
+                        // although ideally this situation should never occur in releases post its introduction.
+                        var platformInfo = getPlatformInfo(),
```
nethip: If you want to retain this , you might want to move all of this to `notifyGetUpdateInstaller`

mbhavi:
@nethip , the review comments have been addressed now. I have added the preference to enable/disable auto update, made AutoUpdate no-op on Linux, replaced the icons, and moved most of the core code onto extension side. Please find the changes in b50f8de
The changes are ready to be reviewed again.

src/utils/UpdateNotification.js
```diff
@@ -61,6 +71,10 @@ define(function (require, exports, module) {
     // Data about available updates in the registry
     PreferencesManager.stateManager.definePreference("extensionUpdateInfo", "Array", []);
 
+    //Preference for hosting latest installer on a local server
+    //AUTOUPDATE_UNITTESTING
+    PreferencesManager.stateManager.definePreference("autoUpdate.testUrl", "string", "");
```
nethip: Do you want to still retain this preference?

src/extensions/default/AutoUpdate/StateHandler.js
mbhavi: @nethip , the requirement here is to do a force write. If the file exists, the content is overwritten. If it doesn't exist, the file is created and written. I am passing the boolean to be "true" in writeText, which ensures this.
Hence, the validity is not needed here.

src/utils/UpdateNotification.js
mbhavi: @nethip , Yes we need to retain this preference, as this will be very much needed for unit testing purposes.

nethip: Please have this preference in a separate branch. We can't be shipping with test URLs preference.

src/extensions/default/AutoUpdate/StateHandler.js
nethip: @mbhavi Let us not have discussion of having a null check or not. It is a simple `null` check. If you still feel that is not required, post merging this PR, one of us will add it.

src/extensions/default/AutoUpdate/main.js
```diff
@@ -0,0 +1,833 @@
...
+    }
+    
+    /**
+     * Overriding the appReady for Auto update
+     */
+
+    AppInit.appReady(function () {
+        
+        // Auto Update is supported on Win and Mac, as of now
+        if (brackets.platform !== "linux") {
```
nethip: I think one of the review comments got missed here. You need to check if `brackets.app.setUpdateParams` is registered or not. Only then run through the flow.

nethip:
Thanks for addressing the review comments @mbhavi . I have added more comments. Please address those comments and I think we are good to merge from the dev perspective.

src/extensions/default/AutoUpdate/StateHandler.js
mbhavi: @nethip , addressed in 4fb6582

mbhavi:
@nethip, I have addressed the new comments in 4fb6582. Please review.

src/utils/UpdateNotification.js
mbhavi: @nethip, I have removed the preference from this branch. I will probably include the changes in a separate branch then, for unit testing purposes.

nethip (approved):

nethip:
@mbhavi One last check. What if the node domain fails? Can we figure that out and bail out in your extension?

mbhavi:
@nethip , I have addressed the failure scenario where the node domain fails in b272289. Please have a quick review for the same.

ayushrathi15:
LGTM.

@nethip @mbhavi I am done with the private build testing. We are good to go ahead with the merge.

src/extensions/default/AutoUpdate/main.js
```diff
@@ -0,0 +1,840 @@
...
+        updateDomain.promise()
+             .done(function() {
+                 setupAutoUpdatePreference();
+                 if(_isAutoUpdateEnabled()) {
+                     setupAutoUpdate();
+                     UpdateNotification.registerUpdateHandler(_updateProcessHandler);
+                 }
+             })
+             .fail(function() {
+                 return;
```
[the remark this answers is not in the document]
mbhavi: @nethip, addressed this in 9b89697.

dmittleman:
Just downloaded 1.13 and am reading discussion from early April.  Here is a **use case** as to why **you need a preference to disable auto updates**.

I teach at DePaul using Brackets as the editor for my in class demonstrations.  Brackets is installed on all student machines in our teaching labs.  We are a large college (maybe 4,000 or so students).  We currently have about 30 computer labs with Brackets installed on about 800 machines.  Our IT staff annually creates 20 or so different "images" for these computers (based on lab purpose and computer OS/capabilities.   The IT staff applies specific images to specific computers, and re-images the computers every week or so via an automated process to ensure the machines are in a known and usable state.

So, if an individual student is at a machine when the update message comes up and opts to update, then that machine is in a different state than expected for the rest of the week.  Some students sitting in a classroom lab may be on a different version of Brackets than the instructor.

Our IT staff certainly wants control over changes of state overriding what an individual student user might opt to do.   I don't think we are unique in this situation.

Please tell me this use case has been considered.

PR #14189, opened 2018-03-30
Partial fix for "Cleanup unit tests that wait for a promise #4638"

spwg:
This is my first time contributing here (or anywhere in open source actually). Sorry about any basic mistakes. I looked at starter bugs and chose #4638. I read the tests in Async-test.js, looked at some of the docs for Jasmine, and made the change. My PR refactors one function, not all older async unit tests across the entire branch. I worked on CSS-Utils-test.js for a bit also, but it seems like that whole file needs to be refactored.


spwg:
Not sure why continuous-integration/travis-ci/pr failed. At line 1793 of the 'Details' button above, it says "API rate limit exceeded." Is this the cause of the failure? How do I fix it?

PR #14196, opened 2018-04-03
Remove unused acorn node module

boopeshmahendran:


shubhsnov:
LGTM 👍

ficristo:
I think you should be able to move those deps to the top level package.json and then remove this one.

ficristo:
I think you should remove this file anyway and move `tern` dependecy to the top level package.json.
package.json files should be present:
- at the top level (for node deps)
- inside src folder (for browser deps)
- inside extensions folder
- for tests
all the others should be removed

PR #14220, opened 2018-04-13
Auto update data logging

niteskum:


abose:
You might want to raise this PR tracking https://github.com/adobe/brackets/pull/14177 branch instead of adobe:master .  Looks like the auto-update code is also pulled into this CR.

Issue #12996, opened 2016-12-21
Touch Bar Support for Brackets


petercampanelli:
It would be really cool if Brackets could support the new Touch Bar on the late 2016 MacBook Pros. There are plenty of exciting functions you could add like a color picker, live view button, autocomplete, etc. As more and more Brackets users start using Macs with the Touch Bar, it is a good idea to be on the front end of this feature.

PR #14221, opened 2018-04-16
Base Level TouchBar Support

jordanwolff:
An extension for the Brackets app. The main.js file implements base level classes that can be utilized to implement TouchBar for Apple MacBook Pro computers. The code has been written with the goal of being easily extensible by other developers. I created this in response to Issue #12996. The Brackets app needs features like this one in order to keep pace with other text editing tools. I hope its well received and would appreciate any feedback and ideas.

Attached (below) is the goal I had for what the end implementation may look like. Thank you all!

touchbar-proto 001


bellarosedc:
Consistent and clear naming convention and structure throughout the code. Comments were very helpful, especially the RMEs for each function. I didn't notice any bugs. All questions I had in regards to code design were easily justified.

swmitra:
@jordanwolff The change set suggests that this is just a feature mock and not the actual feature implementation. Closing this PR now and will look forward for the actual feature implementation from you, probably once we have a CEF shell that provides support for TouchBar 👍

narcarsiss:
Do we have any progress on this?
would like to see a feature set where in split view, you can long hold the touch bar "tab" to move tabs to either side.
[ tab     |     tab ].

Colour Picker; with ability to set Design defaults, or read SASS colour variables.
Default commands, like shortcuts, (undo, redo, save, live-load), support for brackets-git, emmet.

Regards.

PR #14227, opened 2018-04-18
[WIP] Related Files - Extension Module

shraddhasarda:
- Extension that provides related files (CSS and JS) for given HTML document.
- Unit test case to validate related files provided by the extension.

### DIFF
--- a/src/extensions/default/RelatedFiles/main.js
+++ b/src/extensions/default/RelatedFiles/main.js
@@ -1,0 +1,202 @@
+/*
+ * Copyright (c) 2018 - present Adobe Systems Incorporated. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE 
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+define(function (require, exports, module) {
+    'use strict';
+    
+    // Load dependent modules.
+    var AppInit             = brackets.getModule("utils/AppInit"),
+        DocumentManager     = brackets.getModule("document/DocumentManager"),
+     	Filesystem          = brackets.getModule("filesystem/FileSystem"),
+        CommandManager      = brackets.getModule("command/CommandManager"),
+        Commands            = brackets.getModule("command/Commands"),
+        StringMatch         = brackets.getModule("utils/StringMatch"),
+        RelatedFilesManager = brackets.getModule("search/RelatedFilesManager"),
+        PathUtils           = brackets.getModule("thirdparty/path-utils/path-utils");
+
+    /**
+    * @private
+    * 
+    * Used to keep track of list of Related Files for a given document.
+    *
+    **/
+    var relatedFiles;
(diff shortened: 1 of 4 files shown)


swmitra:
@shraddhasarda There are lint errors in the change list for this PR. 
Please refer to the following errors from the build log - 
image

src/extensions/default/RelatedFiles/main.js
```diff
@@ -0,0 +1,202 @@
...
+			var docText = currentDocument.getText();
+            var document = new DOMParser().parseFromString(docText, "text/html");
+        
+            var linkTags = $(document).find("link[rel='stylesheet']");
+            var scriptTags = $(document).find("script[src!=null]");
+            
+        	var i, fileURI, searchResult;
+			
+            // Search for href in link tag to get related file.
+            for (i = 0; i < linkTags.length; i++) {
```
swmitra: Extract this as a separate function and use that function for both styles and scripts.

src/extensions/default/RelatedFiles/main.js
```diff
@@ -0,0 +1,202 @@
+/*
```
swmitra: Please fix indentation issues. Whole file is having indentation issues scattered across.

src/extensions/default/RelatedFiles/main.js
```diff
@@ -0,0 +1,202 @@
...
+        return searchResult;
+    }
+    
+    /**
+    * Create a SearchResult object which has path information about the remote file url.
+    * @params {string} tagurl The url from which file path information is to be extracted.
+    * @returns {Object} SearchResult object which has remote link path information.
+    **/
+    function getLinkFromRemoteURL (tagurl) {
+        var regex = /:\/\/(.[^/]+)/;
```
swmitra: Instead of using regex here, you can simply use `FileUtils` to get the file name which can be used as `label`.

test/SpecRunner.js
```diff
@@ -76,6 +76,7 @@ define(function (require, exports, module) {
     require("JSUtils/Session");
     require("JSUtils/ScopeManager");
     require("widgets/InlineMenu");
+    require("search/RelatedFilesManager");
```
swmitra: Should this change be part of the other PR for `RelatedFilesManager`?

src/extensions/default/RelatedFiles/main.js
```diff
@@ -0,0 +1,202 @@
...
+    
+    function RelatedFiles() {
+        
+    }
+    
+    /**
+    * Gets list of realted files.
+    * @returns {List} List of realted files.
+    **/
+    RelatedFiles.prototype.getRelatedFiles = function() {
```
swmitra: From the code it seems like we are using assumed dependency, which is the `currentDocument`. This limits the scope of extensibility. We could use injected dependency by simply accepting a `document` in this function signature and then pass it on to `parseHTML`. If a valid document is passed, parse that document text ( if it's 'html') else assume the document to be parsed as `currentDocument`.

swmitra: Lets keep this signature uniform which can be leveraged later by other extensions as well. IMO, this should be a `promise` returning function.

src/extensions/default/RelatedFiles/main.js
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+    * Create abosulte path for a file from relative path.
+    * @params {string} base Path of the current document.
+    * @params {string} relative Relative path of a file which is to be converted to absolute path.
+    * @returns {string} Absolute path of the given relative path with calculated wrt base.
+    **/
+    function getAbsoultePath(base, relative) {
+        var baseUrl = base.split("/"),
```
swmitra: It seems if there are 20 related files, this code will do this base path extraction by 20 times, where as for all of them it remains same and can be extracted only once in the caller by using the containing html document path. You can modify the logic and accept a base path rather than a base file path.

src/extensions/default/RelatedFiles/main.js
```diff
@@ -0,0 +1,202 @@
...
+    
+    /**
+    * Create a SearchResult object which has path information about the local file.
+    * @params {string} tagurl The url from which file path information is to be extracted.
+    * @returns {Object} SearchResult object which has file path information.
+    **/
+    function getLinkFromLocalURL(tagurl) {
+        var searchResult = new StringMatch.SearchResult (tagurl);
+        var file = Filesystem.getFileForPath(tagurl);
+        searchResult.label = file._name;
```
swmitra: `searchResult` entry creation can be extracted as a parameterised function and called from `getLinkFromLocalURL` and `getLinkFromRemoteURL`.

src/extensions/default/RelatedFiles/main.js
```diff
@@ -0,0 +1,202 @@
...
+    * Identify whether a link is for local file or is a remote link.
+    * Store all related files link in relatedFiles variable.
+    **/
+    function parseHTML()
+    {
+        var currentDocument = DocumentManager.getCurrentDocument();
+        if (window.DOMParser && currentDocument)
+        {
+			var docText = currentDocument.getText();
+            var document = new DOMParser().parseFromString(docText, "text/html");
```
swmitra: We need not create `DOMParser` instances in every `parseHTML` call. It can be created in the `RelatedFiles` constructor. `parseFromString` is a pure function and reentrant.

swmitra (changes requested):

vickramdhawal:
Closing this PR as this is not in progress any more.

PR #14228, opened 2018-04-19
move xorigin.js after css

phapdinh:
This improves page load performance as it moves the script tag for xorigin.js after css

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -31,9 +31,6 @@
 
     <!-- NOTE: All scripts must be external for Chrome App support: http://developer.chrome.com/apps/app_csp.html -->
 
-    <!-- Warn about failed cross origin requests in Chrome -->
-    <script type="application/javascript" src="xorigin.js"></script>
-
     <link rel="stylesheet" type="text/css" href="thirdparty/CodeMirror/lib/codemirror.css">
     <!--(if target dev)><!-->
     <link rel="stylesheet/less" type="text/css" href="styles/brackets.less">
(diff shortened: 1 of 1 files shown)


vickramdhawal:
Hi Phapdinh.
Thanks for the contribution. 

Can you provide some data on the performance improvement gained with this change ?
Or is there is a perceived performance improvement in UI Rendering with this change ? Does brackets seem to load faster ? Can you establish this ?

phapdinh:
The change will cause a very slight performance improvement that is hard to measure. Moving the file down also helps with keeping the code organized and inline with the comments in the html.

src/index.html
```diff
@@ -45,6 +42,8 @@
     <!-- JavaScript -->
 
     <!-- Pre-load third party scripts that cannot be async loaded. -->
+     <!-- Warn about failed cross origin requests in Chrome -->
```
saulotoledo: Please fix indent (extra space).

src/index.html
```diff
@@ -57,14 +56,14 @@
...
+
     <!-- All other scripts are loaded through require. -->
     <!-- build:js main.js -->
     <script src="thirdparty/requirejs/require.js" data-main="main"></script>
     <!-- endbuild -->
 
     <!-- Verify that all dependencies are loaded. -->
     <script src="dependencies.js"></script>
-
+
```
saulotoledo: Unnecessary change. Please remove it.

saulotoledo (changes requested):
Minor comments to prevent us to start ignoring minor details in the project.

phapdinh:
@saulotoledo I have squashed the five commits into 1

saulotoledo (approved):
@phapdinh It is ok for me. I will keep the rest of the discussion with @vickramdhawal. :)

saulotoledo:
@navch What do you think about this PR?

navch:
I am fine with this change. Just make sure this is not causing any problem. To confirm that we should run all unit tests and make sure there should not be any new error in the console because of this change.

PR #14230, opened 2018-04-19
AutoUpdate: Fix for Update Failed message coming up on app relaunch,…

mbhavi:
… even when the update is successful.

ping @nethip  @swmitra for review.


navch:
LGTM 👍

PR #14014, opened 2018-01-05
Update CSS property hints

valtlai:
Reordered properties and their values alphabetically.

Added CSS-wide keywords – `inherit`, `initial`, `revert`, `unset` – to the end of arrays.

<details>
<summary>Added properties</summary>

* `backdrop-filter`
* `background-position-x`
* `background-position-y`
* `block-size`
* `border-block-end-color`
* `border-block-end-style`
* `border-block-end-width`
* `border-block-end`
* `border-block-start-color`
* `border-block-start-style`
* `border-block-start-width`
* `border-block-start`
* `border-inline-end-color`
* `border-inline-end-style`
* `border-inline-end-width`
* `border-inline-end`
* `border-inline-start-color`
* `border-inline-start-style`
* `border-inline-start-width`
* `border-inline-start`
* `clip-path`
* `clip-rule`
* `contain`
* `font-variation-settings`
* `image-rendering`
* `inline-size`
* `justify-items`
* `justify-self`
* `margin-block-end`
* `margin-block-start`
* `margin-inline-end`
* `margin-inline-start`
* `max-block-size`
* `max-inline-size`
* `min-block-size`
* `min-inline-size`
* `offset-distance`
* `offset-path`
* `offset-rotate`
* `offset`
* `overflow-anchor`
* `overflow-wrap`
* `overscroll-behavior-x`
* `overscroll-behavior-y`
* `overscroll-behavior`
* `padding-block-end`
* `padding-block-start`
* `padding-inline-end`
* `padding-inline-start`
* `place-content`
* `place-items`
* `place-self`
* `scroll-margin-block-end`
* `scroll-margin-block-start`
* `scroll-margin-block`
* `scroll-margin-bottom`
* `scroll-margin-inline-end`
* `scroll-margin-inline-start`
* `scroll-margin-inline`
* `scroll-margin-left`
* `scroll-margin-right`
* `scroll-margin-top`
* `scroll-margin`
* `scroll-padding-block-end`
* `scroll-padding-block-start`
* `scroll-padding-block`
* `scroll-padding-bottom`
* `scroll-padding-inline-end`
* `scroll-padding-inline-start`
* `scroll-padding-inline`
* `scroll-padding-left`
* `scroll-padding-right`
* `scroll-padding-top`
* `scroll-padding`
* `scroll-snap-align`
* `scroll-snap-stop`
* `scroll-snap-type`
* `speak`
* `text-combine-upright`
* `text-justify`
* `text-orientation`
* `text-size-adjust`
[...]

PR #14238, opened 2018-04-21
Append global values dynamically to CSS property hints

valtlai:
* Move the global CSS values to a separate file.
* Append them dynamically to the hint list.
* Alphabetically sort CSS properties and their values on CSSProperties.json.


valtlai:
Buggy. Closing for now.

src/extensions/default/CSSCodeHints/CSSGlobalValues.json
```diff
@@ -0,0 +1,7 @@
+[
+    "inherit",
+    "initial",
+    "revert",
+    "unset",
+    "var()"
```
navch: In this case, on inserting hint cursor will be after ")". Which is not much useful. Because user need to place cursor manually.

src/extensions/default/CSSCodeHints/main.js
```diff
@@ -38,7 +38,9 @@ define(function (require, exports, module) {
         ColorUtils          = brackets.getModule("utils/ColorUtils"),
         Strings             = brackets.getModule("strings"),
         CSSProperties       = require("text!CSSProperties.json"),
-        properties          = JSON.parse(CSSProperties);
+        CSSGlobalValues     = require("text!CSSGlobalValues.json"),
+        properties          = JSON.parse(CSSProperties),
```
navch: nit: Please write these entries separately.

navch (commented):
Thanks @valtlai for this PR.

I have some other questions as well.
As we are sorting hints dynamically then why we need sorted CSS properties?
Please make sure hints are sorted after appending global hints.

src/extensions/default/CSSCodeHints/main.js
navch: i.e.
```
         CSSProperties       = require("text!CSSProperties.json");

var CSSGlobalValues     = require("text!CSSGlobalValues.json"),
        properties          = JSON.parse(CSSProperties);
```

navch:
Adding reference https://github.com/adobe/brackets/pull/14014

valtlai:
I’m no longer working on this, so closing.

PR #14251, opened 2018-04-25
Support for Auto-update in Prereleases

mbhavi:
Support for Auto-update in Prereleases.

ping @swmitra @navch @shubhsnov for review.

### DIFF
--- a/src/extensions/default/AutoUpdate/main.js
+++ b/src/extensions/default/AutoUpdate/main.js
@@ -338,9 +338,9 @@ define(function (require, exports, module) {
                     downloadURL;
 
                 //AUTOUPDATE_PRERELEASE BEGIN
-                // This code change is needed for update support in prereleases.
-                // And should be removed once the feature goes to production.
-                // Overwriting the tag and downloadURL for prerelease
+                // This code change is needed for Auto-update support in prereleases
+                // And will be removed eventually for stable releases
+                // Overwriting the tag and downloadURL for prereleases
                 {
                     if(buildName && buildName.indexOf("Pre Release") !== -1) {
 
--- a/src/extensions/default/AutoUpdate/main.js
+++ b/src/extensions/default/AutoUpdate/main.js
@@ -337,6 +337,19 @@ define(function (require, exports, module) {
                     installerName = "Brackets." + buildName.split(" ").join(".") + ext,
                     downloadURL;
 
+                //AUTOUPDATE_PRERELEASE BEGIN
+                // This code change is needed for update support in prereleases.
+                // And should be removed once the feature goes to production.
+                // Overwriting the tag and downloadURL for prerelease
+                {
+                    if(buildName && buildName.indexOf("Pre Release") !== -1) {
+
+                        tag = "release-" + buildName.split(" ").pop() + "-prerelease";
+                        installerName = "Brackets." + buildName.split(" ").join("-") + ext;
+                    }
+                }
+                //AUTOUPDATE_PRERELEASE END
+
                 downloadURL = brackets.config.update_download_url + tag + "/" + installerName;
 
                 downloadInfo = {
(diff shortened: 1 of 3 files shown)


src/utils/UpdateNotification.js
```diff
@@ -115,6 +115,28 @@ define(function (require, exports, module) {
...
+            var jsonLocale = locale;
+            if (locale && locale.toLowerCase().startsWith("de")) {
+                jsonLocale = "de";
+            } else if (locale && locale.toLowerCase().startsWith("es")) {
+                jsonLocale = "es";
+            } else if (locale && locale.toLowerCase().startsWith("fr")) {
+                jsonLocale = "fr";
+            } else if (locale && locale.toLowerCase().startsWith("ja")) {
+                jsonLocale = "ja";
+            } else if (!locale || !(locale === "de" || locale === "en" || locale === "es" || locale === "fr" || locale === "ja")) {
```
swmitra: could be the default `else` statement. Why are we even doing a check here. The logic is - 
* check for known locales
* if found do specific handling
* if not then fallback to `en`

src/utils/UpdateNotification.js
```diff
@@ -115,6 +115,28 @@ define(function (require, exports, module) {
             locale = locale.substring(0, 2);
         }
 
+         //AUTOUPDATE_PRERELEASE_BEGIN
+        // The following code is needed for supporting Auto Update in prerelease,
+        //and will be removed eventually for stable releases
+        {
+            var jsonLocale = locale;
+            if (locale && locale.toLowerCase().startsWith("de")) {
```
swmitra: This entire set of branches could be more readable with `switch` cases.

src/utils/UpdateNotification.js
```diff
@@ -115,6 +115,28 @@ define(function (require, exports, module) {
             locale = locale.substring(0, 2);
         }
 
+         //AUTOUPDATE_PRERELEASE_BEGIN
+        // The following code is needed for supporting Auto Update in prerelease,
+        //and will be removed eventually for stable releases
+        {
+            var jsonLocale = locale;
```
swmitra: Why another variable? No mutation would happen on the original string anyway...

src/extensions/default/AutoUpdate/main.js
```diff
@@ -337,6 +337,19 @@ define(function (require, exports, module) {
                     installerName = "Brackets." + buildName.split(" ").join(".") + ext,
                     downloadURL;
 
+                //AUTOUPDATE_PRERELEASE BEGIN
+                // This code change is needed for Auto-update support in prereleases
+                // And will be removed eventually for stable releases
+                // Overwriting the tag and downloadURL for prereleases
+                {
+                    if(buildName && buildName.indexOf("Pre Release") !== -1) {
```
swmitra: Ideally we shouldn't do this kind of index of check. This are prone to failures because of mixed cases. We can introduce a prerelease flag in the release entry and just put a check for that. `buildName` is being constructed with `buildNumber` and `versionString` I guess. Don't put spaces in the value.

swmitra:
@mbhavi Please update the `config.json` as part of this PR. We would revert this PR alone when going for stable release.

swmitra (changes requested):

swmitra:
While reviewing this PR, I thought we could improve on the overall workflow by introducing an asset container config in the updates json itself. That will give us liberty over putting the installer assets in any container without changing even a single line of code! That would conform to the common best practices by not computing a resource URI on the client side, rather it would depend on the config URL it gets from `updates.json`.
@mbhavi @vickramdhawal @nethip Your thoughts?

mbhavi:
@swmitra , I have moved the resource Uris to the server, as suggested. And raised a PR for the same as [[14257]](https://github.com/adobe/brackets/pull/14257)

I will upload only the prerelease changes here, once the above PR is merged, in order to ensure that we can revert this PR alone when going for stable release.

swmitra (approved):

PR #14256, opened 2018-04-25
AutoUpdate Event UpdateAvailableDialogBoxRendered SubType String not correct Fixed

niteskum:
changes the UpdateAvailableDialogBoxRendered Event Sub Type String based on the trigger point
"userAction" -> when User manually click "Check For Updates" in Help Command
"auto" -> when AutomaticUpdates are launched.


niteskum:
ping @mbhavi @navch @swmitra

navch (approved):

PR #14282, opened 2018-04-27
Change to send Analytics Data to HealthData Server also

niteskum:
This PR includes changes for sending Analytics Data to HealthData Server.

### DIFF
--- a/src/extensions/default/HealthData/HealthDataManager.js
+++ b/src/extensions/default/HealthData/HealthDataManager.js
@@ -213,20 +213,27 @@ define(function (require, exports, module) {
 
     // Send Analytics data to Server
     function sendAnalyticsDataToServer(eventParams, serverUrl) {
-        var result = new $.Deferred();
+        var result = new $.Deferred(),
+            url = serverUrl || brackets.config.analyticsDataServerURL;
 
-        serverUrl = serverUrl || brackets.config.analyticsDataServerURL;
+        var analyticsData = getAnalyticsData(eventParams),
+            ajaxParams = {
+                url: url,
+                type: "POST",
+                data: JSON.stringify({events: [analyticsData]})
+            };
 
-        var analyticsData = getAnalyticsData(eventParams);
-        $.ajax({
-            url: serverUrl,
-            type: "POST",
-            data: JSON.stringify({events: [analyticsData]}),
-            headers: {
+        if(serverUrl) {
+            ajaxParams.dataType = "text";
+            ajaxParams.contentType = "text/plain";
+        } else {
+            ajaxParams.headers =  {
                 "Content-Type": "application/json",
                 "x-api-key": brackets.config.serviceKey
-            }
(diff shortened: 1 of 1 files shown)


niteskum:
ping @swmitra @navch @vickramdhawal  for review

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -306,25 +308,41 @@ define(function (require, exports, module) {
...
+            url = brackets.config.healthDataServerURL,
+            sendFailed = false;
 
         var options = {
             location: {
                 scope: "default"
             }
         };
 
+        var handleFail = function handleFail() {
```
navch: Probably you wrote this syntax by mistake. This is useful when we are running in IE or when we want to assign a small function name to use inside the function.

navch: It's not a good idea to write nested function if you function executed multiple time.

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -306,25 +308,41 @@ define(function (require, exports, module) {
...
+            if(sendFailed) {
+                PreferencesManager.setViewState(Eventparams.eventName, 0, options);
+            } else {
+                sendFailed = true;
```
navch: What is the use of this sendFailed? Looks like it will always execute else part.

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -306,25 +308,41 @@ define(function (require, exports, module) {
...
-                    .done(function () {
-                        PreferencesManager.setViewState(Eventparams.eventName, 1, options);
-                        result.resolve();
-                    }).fail(function () {
-                        PreferencesManager.setViewState(Eventparams.eventName, 0, options);
-                        result.reject();
+                    .always(function () {
+                        sendAnalyticsDataToServer(Eventparams, url)
+                           .done(function () {
+                               result.resolve();
```
navch: Just make sure view state is correct in every case like not able to send health data server but successfully sent on analytics server.

navch (changes requested):

src/extensions/default/HealthData/HealthDataManager.js
niteskum: View State is unset only if it fails to send Data to both the server. If it fails for any of one server then it will execute else part and sendFailed is assigned to true. when it will fail for second server also It will go in  if part and will unset the view state.

navch (approved):

vickramdhawal:
@niteskum I'm not sure i understand the purpose of this PR. If the user is offline even posting to health data url will fail. Should we not have simply passed on the data to the HealthLogger which would have guaranteed eventual persistence with the health data.

abose:
What is the intent of this change?

PR #14287, opened 2018-04-30
Store drag and drop data as a global variable

boopeshmahendran:
In the newer CEF versions(such as 3.2785 the one which Brackets on Linux currently uses), the drag and drop data cannot be accessed in the dragOver event, so store it as a global variable to access it in the dragOver event.


swmitra (approved):
LGTM.

boopeshmahendran:
Thank you @swmitra

PR #14312, opened 2018-05-07
Fix issue 14298

### LINKED ISSUE #14298: Dilato: Core: The selected file isn't highlighted after move it two times
Steps:
1. Open Brackets
2. Go to file→open folder
3. Select some folder which has many subfolders and files with different types
4. Select a file and drag and drop a file to another folder
5. Select the same file and drag and drop a file to another folder

Result: The selected file isn't highlighted after move it two times

Expected: The selected file still keep highlight after move it two times

issue

boopeshmahendran:
Fixes #14298 
If the selected file is renamed(or moved), keep track of it in the `_currentPath` variable

### DIFF
--- a/src/project/ProjectModel.js
+++ b/src/project/ProjectModel.js
@@ -991,6 +991,7 @@ define(function (require, exports, module) {
             viewModel.renameItem(oldProjectPath, self.makeProjectRelativeIfPossible(newPath));
             if (self._selections.selected && self._selections.selected.indexOf(oldPath) === 0) {
                 self._selections.selected = newPath + self._selections.selected.slice(oldPath.length);
+                self.setCurrentFile(newPath);
             }
         }
 


src/project/ProjectModel.js
```diff
@@ -991,6 +991,7 @@ define(function (require, exports, module) {
             viewModel.renameItem(oldProjectPath, self.makeProjectRelativeIfPossible(newPath));
             if (self._selections.selected && self._selections.selected.indexOf(oldPath) === 0) {
                 self._selections.selected = newPath + self._selections.selected.slice(oldPath.length);
+                self.setCurrentFile(newPath);
```
navch: Whats the reason behind _selections.selected being null in second drag?

boopeshmahendran: @navch This [line](https://github.com/adobe/brackets/blob/master/src/project/ProjectModel.js#L627) sets selection to null

navch: 👍 Sounds reasonable.

navch (approved):

Issue #14216, opened 2018-04-13
[Brackets auto-update Windows/Mac] If a user has multiple brackets window open then update bar comes on all the instances.


ayushrathi15:
### Description
[Brackets auto-update Windows/Mac]  If a user has multiple brackets window open then update bar comes on all the instances.

### Steps to Reproduce
1. Launch brackets 1.13.
2. Click on Update Notification Button.
3. Click on Get it Now.
4. Click on later in update bar.
5. Open another brackets window.
6. In the new window, click on update notification icon.
7. The update bar can be seen on both the instances.

<img width="2048" alt="screen shot 2018-04-12 at 1 33 19 am" src="https://user-images.githubusercontent.com/25339865/38725009-86d18418-3f23-11e8-833f-e296480bef05.png">

**Expected behavior:** Update bar should be displayed only on one instance.

**Actual behavior:** [Brackets auto-update Windows/Mac] If a user has multiple brackets window open then update bar comes on all the instances.

### Versions
Windows 10 64 Bit
Mac 10.13
Release 1.13 build 1.13.0-17665

ayushrathi15:
Working fine now.
Verified in latest build - Release 1.13 prerelease build 1.13.0-17689.

PR #14320, opened 2018-05-09
Fixing multiple windows issues in AutoUpdate

mbhavi:
Fixing multiple windows issues in AutoUpdate.
Issues fixed : https://github.com/adobe/brackets/issues/14215, https://github.com/adobe/brackets/issues/14216

ping @swmitra , @shubhsnov , @navch  for review


src/extensions/default/AutoUpdate/main.js
```diff
@@ -104,9 +129,12 @@ define(function (require, exports, module) {
      * @param {object}   msgObj - json containing - {
      *                          fn - function to execute on Brackets side
      *                          args - arguments to the above function
+     *                          requester - ID of the current requester domain
      */
     function receiveMessageFromNode(event, msgObj) {
-        functionMap[msgObj.fn].apply(null, msgObj.args);
+        if (domainID === msgObj.requester) {
```
navch: seems like this will always be true for a particular session?

src/extensions/default/AutoUpdate/main.js
```diff
@@ -275,11 +304,15 @@ define(function (require, exports, module) {
      * Sets up the Auto Update environment
      */
     function setupAutoUpdate() {
+        if (!checkIfAnotherSessionInProgress()) {
+            setAutoUpdateSessionInProgress(false);
```
navch: Session in progress is already false. Why are we setting this again?

src/extensions/default/AutoUpdate/main.js
```diff
@@ -394,47 +428,59 @@ define(function (require, exports, module) {
...
 
-         var updateParams = {
-            downloadURL: downloadURL,
-            installerName: installerName,
-            latestBuildNumber: latestUpdate.buildNumber,
-            checksum: checksum
-        };
+            } else {
+                // Update not present for current platform
+                return;
```
navch: Do we need to reset value of "AutoUpdate.inProgress" in these failures?

src/extensions/default/AutoUpdate/main.js
```diff
@@ -507,6 +555,7 @@ define(function (require, exports, module) {
              .done(function () {
                 setupAutoUpdatePreference();
                 if (_isAutoUpdateEnabled()) {
+                    setupAutoUpdateInProgressState();
```
navch: we can initialize domainID here only.

navch (commented):
Good Job 👍 Just leaving few comments
1) How can we make sure we are resetting preference in every failure or whenerer required.
2) Indentation is wrong in some cases ?

src/extensions/default/AutoUpdate/main.js
mbhavi: Addressed in 036e4a8

src/extensions/default/AutoUpdate/main.js
mbhavi: Addressed in 036e4a8

src/extensions/default/AutoUpdate/main.js
mbhavi: Addressed in 036e4a8

src/extensions/default/AutoUpdate/main.js
mbhavi: This will always be true for a particular window session, but different for different windows.

src/extensions/default/AutoUpdate/main.js
```diff
@@ -148,7 +179,8 @@ define(function (require, exports, module) {
     function postMessageToNode(messageId) {
         var msg = {
             fn: messageId,
-            args: getFunctionArgs(arguments)
+            args: getFunctionArgs(arguments),
+            requester: domainID
```
shubhsnov: This looks fine, but can you see if we can somehow notify the user of Update failure if the requesting window is closed. That should at least provide some info to the user rather than leave him hanging.

src/extensions/default/AutoUpdate/main.js
```diff
@@ -279,7 +311,8 @@ define(function (require, exports, module) {
 
         updateDomain.exec('initNode', {
             messageIds: MessageIds,
-            updateDir: updateDir
+            updateDir: updateDir,
+            requester: domainID
```
swmitra: Looking at the command name it seems `initNode` is for initialising the node domain. Why do we need `domainID` in that case?

swmitra (changes requested):
As a whole the changed files contains mixed use of tabs and spaces. Please correct them and make the indentation with 4 spaces.

src/extensions/default/AutoUpdate/node/AutoUpdateDomain.js
```diff
@@ -87,12 +91,19 @@
      * @param {string} messageId - Message to be passed
      */
     function postMessageToBrackets(messageId) {
+        if(!requesters[currentRequester]) {
+            for (var key in requesters) {
+                currentRequester = key;
```
vickramdhawal: why do we need the requesters list if we always send message only to the currentRequester ?

vickramdhawal (changes requested):

src/extensions/default/AutoUpdate/main.js
```diff
@@ -90,6 +92,35 @@ define(function (require, exports, module) {
...
+     */
+    function checkIfAnotherSessionInProgress() {
+        var result = $.Deferred();
+        if(updateJsonHandler) {
+            var state = updateJsonHandler.state;
+
+            updateJsonHandler.read(updateProgressKey)
+                .done(function(val) {
+                    updateJsonHandler.state = state;
+                    setUpdateStateInJSON(updateProgressKey, val)
```
vickramdhawal: why are we setting the value we just read ?

src/extensions/default/AutoUpdate/main.js
```diff
@@ -354,23 +397,48 @@ define(function (require, exports, module) {
...
+                                description: Strings.AUTOUPDATE_IN_PROGRESS
+                            });
+                        } else {
+                             setUpdateStateInJSON("autoUpdateInProgress", true)
+                                .done(setUpdateInProgress);
+
+                        }
+                    })
+                    .fail(function() {
+                        setUpdateStateInJSON("autoUpdateInProgress", true)
```
vickramdhawal: If we fail to read the status of the session, should we be trying to start a session ? I would rather bail auto-update all together ?

vickramdhawal (changes requested):

src/extensions/default/AutoUpdate/main.js
niteskum: This is addressed , If requesting window is closed , in Node anyone of available windows is made as requesting window

src/extensions/default/AutoUpdate/main.js
niteskum: Corrected.

src/extensions/default/AutoUpdate/main.js
niteskum: Mainly It will fail to read status of the session when File is not available or Key is not present in File.
in such cases we should try starting a session.

vickramdhawal (approved):

PR #14321, opened 2018-05-11
Appending Remote word in working set

### LINKED ISSUE #14309: [Remote File] If user save the file name having same name as url then nothing is displayed in front of file name.
### Description
If user save the file name having same name as url then nothing is displayed in front of file name.

### Steps to Reproduce
1. Launch brackets 1.13.
2. Press ctrl+shift+O for win and cmd+shift+O for mac.
3. Quick open modal bar will open.
4. Enter a network url that supports http or https. Example: http://www.muse-php.com
5. The code for the page is displayed in brackets and file is visible in working files in file tree.
6. Go to file -> Save As
7. Save the file with same name i.e. muse-php.com

two file

**Expected behavior:** Their should be a text so that user can differentiate between two files.

**Actual behavior:** If user save the file name having same name as url then nothing is displayed in front of file name.

### Versions
Win 10 64 Bit
Mac 10.12
Release 1.13 build 1.13.0-17668

navch:
Fixes #14309

### DIFF
--- a/src/utils/ViewUtils.js
+++ b/src/utils/ViewUtils.js
@@ -25,8 +25,7 @@ define(function (require, exports, module) {
     "use strict";
 
     var _                   = require("thirdparty/lodash"),
-        LanguageManager     = require("language/LanguageManager"),
-        Strings             = require("strings");
+        LanguageManager     = require("language/LanguageManager");
 
     var SCROLL_SHADOW_HEIGHT = 5;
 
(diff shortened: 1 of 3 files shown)


src/utils/ViewUtils.js
```diff
@@ -427,6 +428,15 @@ define(function (require, exports, module) {
             var fp = file.fullPath.split("/");
             fp.pop(); // Remove the filename itself
             displayPaths[index] = fp.pop();
+
```
shubhsnov: See if we can use useIconProviders and useClassProviders from Working Set View for this as well

shubhsnov: ping @swmitra  for guidance. I believe the "Remote" word can be added to the :after of a class and then that class can be registered with working set view.

swmitra:
We should use the file tree decoration model by providing icon and class provider. It should be a pure CSS solution in decorating the remote file entries.

swmitra (changes requested):
Let's use just the protocol name as marker. It's more friendly for a developer instead of calling Remote. This can be done using a pseudo selector with content set to protocol name.

navch:
@swmitra Review comment addressed. Please have a look now.

shubhsnov:
@navch  I think we should add prototype.name in all File, InMemoryFile and RemoteFile and then check against that directly

boopeshmahendran:
Could we check `constructor.name` like [this](https://github.com/adobe/brackets/blob/master/src/extensions/default/NavigationAndHistory/main.js#L604)?

swmitra:
Closing this PR as #14330 adds protocol badge to all remote file entries in working set and recent file list.

Issue #14310, opened 2018-05-07
[Remote File] User should be given an option to save as the edited file when he closes a remote file from working set in file tree.


ayushrathi15:
### Description
User should be given an option to save as the edited file when he closes a remote file from working set in file tree.

### Steps to Reproduce
1. Launch brackets 1.13.
2. Press ctrl+shift+O for win and cmd+shift+O for mac.
3. Quick open modal bar will open.
4. Enter a network url that supports http or https. Example: https://www.google.com
5. The code for the page is displayed in brackets and file is visible in working files in file tree.
6. Edit this code.
7. Now close this dirty file from working set.
8. User gets a dialogue to save or dont save the file.
9. Click on save file.
10. User gets "error saving file" message.

![save](https://user-images.githubusercontent.com/25339865/39697141-8ea84008-520d-11e8-9234-5752ad311fe6.PNG)

![save file](https://user-images.githubusercontent.com/25339865/39697144-911833ca-520d-11e8-8b69-8ad3b836a7cc.PNG)

**Expected behavior:** User should be given an option to save as the edited file when he closes a remote file from working set in file tree.

**Actual behavior:** No option is given to user and he gets an error saving file message.

### Versions
Win 10 64 Bit
Mac 10.12
Release 1.13 build 1.13.0-17668

PR #14322, opened 2018-05-11
Disabling save button on closing dirty remote file

navch:
Fixing issue https://github.com/adobe/brackets/issues/14310

### DIFF
--- a/src/document/DocumentCommandHandlers.js
+++ b/src/document/DocumentCommandHandlers.js
@@ -1205,7 +1205,12 @@ define(function (require, exports, module) {
         if (doc && doc.isDirty && !_forceClose && (MainViewManager.isExclusiveToPane(doc.file, paneId) || _spawnedRequest)) {
             // Document is dirty: prompt to save changes before closing if only the document is exclusively
             // listed in the requested pane or this is part of a list close request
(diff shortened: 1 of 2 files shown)


src/document/DocumentCommandHandlers.js
```diff
@@ -1205,7 +1205,12 @@ define(function (require, exports, module) {
         if (doc && doc.isDirty && !_forceClose && (MainViewManager.isExclusiveToPane(doc.file, paneId) || _spawnedRequest)) {
             // Document is dirty: prompt to save changes before closing if only the document is exclusively
             // listed in the requested pane or this is part of a list close request
-            var filename = FileUtils.getBaseName(doc.file.fullPath);
+            var filename = FileUtils.getBaseName(doc.file.fullPath),
+                disableSaveBtn = "";
+
+            if (file.toString().indexOf("[RemoteFile") === 0) {
```
swmitra: We should use a better way for RemoteFile instance detection.

swmitra (changes requested):

src/document/DocumentCommandHandlers.js
navch: I tried instanceof that's not working for me. Is there anything which we can use?

PR #14326, opened 2018-05-12
Disabling certain working set context menu items for remote files

navch:
Issue Fixed : - Disabling four commands Save, Show in Explorer, Show in the file tree, Rename.


swmitra (changes requested):
It checks for currently viewed file only. But context menu event can be triggered on any file listed in working set. So, if the currently viewed file is not a RemoteFile, but user has right clicked on a RemoteFile, we would still end up showing all these commands.

navch:
When we open the context menu on any file then first it switches active file viewing then it opens the context menu. As you can check this [here](https://github.com/adobe/brackets/blob/master/src/command/DefaultMenus.js#L53)

shubhsnov:
👍 LGTM

swmitra (approved):

PR #14360, opened 2018-05-25
Adding a pre release build step

vickramdhawal:
Introducing a new brackets.config.prerelease.json which points to the pre release auto update url. Related json files are also updated. a new grunt task is added to support pre release builds.

### DIFF
--- a/src/brackets.config.prerelease.json
+++ b/src/brackets.config.prerelease.json
@@ -3,5 +3,5 @@
     "analyticsDataServerURL"  : "https://cc-api-data.adobe.io/ingest",
     "serviceKey"              : "brackets-service",
     "environment"             : "production",
-    "update_info_url"         : "https://getupdates.brackets.io/getupdates/"
+    "update_info_url"         : "https://s3.amazonaws.com/files.brackets.io/updates/prerelease/<locale>.json"
 }
--- a/src/brackets.config.prerelease.json
+++ b/src/brackets.config.prerelease.json
@@ -2,6 +2,6 @@
     "healthDataServerURL"     : "https://health.brackets.io/healthDataLog",
     "analyticsDataServerURL"  : "https://cc-api-data.adobe.io/ingest",
     "serviceKey"              : "brackets-service",
-    "environment"             : "production",
+    "environment"             : "prerelease",
     "update_info_url"         : "https://s3.amazonaws.com/files.brackets.io/updates/prerelease/<locale>.json"
 }
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -410,7 +410,7 @@ module.exports = function (grunt) {
         'cleanempty',
         'usemin',
         'build-config'
-    ])
+    ]);
 
     // task: build
     grunt.registerTask('build', [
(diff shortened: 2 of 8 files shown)


src/brackets.config.prerelease.json
```diff
@@ -0,0 +1,7 @@
+{
+    "healthDataServerURL"     : "https://health.brackets.io/healthDataLog",
+    "analyticsDataServerURL"  : "https://cc-api-data.adobe.io/ingest",
+    "serviceKey"              : "brackets-service",
+    "environment"             : "production",
```
swmitra: Can it be marked with a different environment `beta` or `experimental` or `prerelease`?

vickramdhawal: I'm not sure what the impact will be in the client code if it is being used. I've just tried to retain it the way it was earlier.
If there are pointers on where all i'll have to look if i change the environment I can take a shot at changing this as well.

boopeshmahendran:
I think, going forward, we can remove the `src/config.json` as it is a machine-generated file. And having it in git just causes confusion whenever we change the config.

swmitra:
in nls/root/Strings.js
```javascript
    // Strings for main-view.html
    "EXPERIMENTAL_BUILD"             : "experimental build",
    "RELEASE_BUILD"                        : "build",
    "PRERELEASE_BUILD"                 : "prerelease build",
    "DEVELOPMENT_BUILD"             : "development build",
```

in src/Strings.js
```javascript
var strings     = require("i18n!nls/strings"),
        urls        = require("i18n!nls/urls"),
        stringsApp  = require("i18n!nls/strings-app"),
        StringUtils = require("utils/StringUtils"),
        BuildConfig = require("text!config.json"),
        ConfigObj   = JSON.parse(BuildConfig);
```

```javascript
var isDevBuild = !StringUtils.endsWith(decodeURI(window.location.pathname), "/www/index.html");
    if (isDevBuild) {
        additionalGlobals.BUILD_TYPE = strings.DEVELOPMENT_BUILD;
    } else {
        if (ConfigObj.environment === 'production') {
            additionalGlobals.BUILD_TYPE = strings.RELEASE_BUILD;
        } else {
            additionalGlobals.BUILD_TYPE = strings.PRERELEASE_BUILD;
        }
        
    }
```
If we can do this small set of changes, prerelease information would be visible to the end user as well 😄

swmitra (approved):

src/brackets.config.prerelease.json
swmitra: The only check I know is in HealthDataManager to avoid logging from dev setups. 
```javascript
// Expose a command to test data sending capability, but limit it to dev environment only
    CommandManager.register("Sends health data and Analytics data for testing purpose", "sendHealthData", function() {
        if (brackets.config.environment === "stage") {
            return checkHealthDataSend(true);
        } else {
            return $.Deferred().reject().promise();
        }
    });
```

vickramdhawal: Yes. I saw this as well. And thought in case of pre-release it should go to production. what do you think ?

vickramdhawal:
@swmitra completed the requested changes. validating the impact, give me some time to update.

src/brackets.config.prerelease.json
swmitra: I gave wrong explanation. I had exposed this command to explicitly send analytics data only when we are in dev environment. This block of code has no functionality when in production.

vickramdhawal: yes. i saw that. have made the change for consistency anyways.

swmitra (approved):

Issue #14243, opened 2018-04-23
[Brackest auto-update] The update logs should be retained irrespective of whether the update is successful or failed


DivyaBC:
Description
[Brackest auto-update] The update logs should be retained irrespective of whether the update is successful or failed

Steps to Reproduce
Install Brackets 1.13
Click on Check for updates.
Click on Get it now.
Once download is complete, Click on Restart in the pop up
Go to  C:\Users\<username>\AppData\Roaming\Brackets\updateTemp and check that there is a the log file generated
Once Brackets is auto launched go to Help->About Brackets
Verify that updated build number shown
Go to  C:\Users\<username>\AppData\Roaming\Brackets\

Expected behavior: The log file should be retained irrespective of whether the update is successful or failed
Also the logs should be moved to C:\Users\<username>\AppData\Local\Temp

Actual behavior: The log file is deleted once update is successful
Versions: Win 10
Release 1.13 build 1.13.0-17656

mbhavi:
@DivyaBC , the removal of update logs on successful update is as per design. The intention is to not bloat up user's storage with heavy logs unless needed. Only in case of update failure is when we really need the logs to be retained, for further information about failure reasons. And so, if the update is successfully installed , keeping the logs is not needed.
Hence, not a bug.

DivyaBC:
@mbhavi , this ask was logged because we once saw an issue where the update was successful but on launching brackets, it was still the older version and because the logs were also cleared we could not figure out what went wrong. In such cases logs can be helpful if retained. hence, the ask to retain the logs irrespective of whether the update is successful or failure

niteskum:
we will keep the installer log within Brackets folder in Appdata.
On every subsequent auto update, we can cleanup previous installation logs.

ayushrathi15:
Verified on 1.13 Prerelease Build 17689

Issue #14225, opened 2018-04-17
Auto update: Dilato: I18N: FR: The percent format of the process number for the auto update build downloading is incorrect.


Zhanxin:
1. Launch Brackets in FR and set the URL for the update server
2. Click Help -> Check for update
3. Click Get it now button

Result: There is no space between number and % in the download progress.
Expected: There should be a space between number and % in the download progress.
![numnerunloc](https://user-images.githubusercontent.com/2245179/38877600-9fe058dc-4291-11e8-8fa4-9f6b786d5828.png)

niteskum:
in Brackets I don't see number is localized anywhere , can we confirm if this is really need to be fixed?

Zhanxin:
This number need to be localized because for some languages, there should be a space between the number and %. For example, in French, the number would display as 9 %.

Zhanxin:
Haven't fix on build 1.13.17689, waiting new builds.

Zhanxin:
Now the build on update server is unavailable for testing, will regress on Monday.

ZuoBaogang:
Haven't fix on build 1.13.17693-1056, waiting new builds.

PR #14364, opened 2018-05-28
Fix for Localization Bug and Retaining the Installer Log

niteskum:
This PR include Fix for Issue #14225 #14243 

ping @navch @boopeshmahendran @shubhsnov for review

### DIFF
--- a/src/extensions/default/AutoUpdate/main.js
+++ b/src/extensions/default/AutoUpdate/main.js
@@ -181,7 +181,7 @@ define(function (require, exports, module) {
      * Checks and handles the update success and failure scenarios
      */
     function checkUpdateStatus() {
-        var filesToCache = null,
+        var filesToCache = ['.logs'],
             downloadCompleted = updateJsonHandler.get("downloadCompleted"),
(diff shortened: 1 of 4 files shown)


boopeshmahendran (approved):

niteskum:
ping @vickramdhawal @swmitra  for review

src/extensions/default/AutoUpdate/main.js
```diff
@@ -204,7 +204,6 @@ define(function (require, exports, module) {
             } else {
                 // We get here if the update started but failed
                 checkInstallationStatus();
-                filesToCache = ['.logs']; //AUTOUPDATE_PRERELEASE
```
vickramdhawal: in line 216 & 218 should we pushing in to the filesToCache array instead of replacing it with a new one ?

niteskum: Replacing with New one is fine. as in that condition Update was not initiated only so there won't be any installer logs.

vickramdhawal (approved):

vickramdhawal:
LGTM 👍

PR #14374, opened 2018-05-30
creating a new function that checks for uuid/olderUuid, if they are null they are populated and set in PreferencesManager

SnchitGrover:
@vickramdhawal

### DIFF
--- a/src/extensions/default/HealthData/HealthDataManager.js
+++ b/src/extensions/default/HealthData/HealthDataManager.js
@@ -200,7 +200,7 @@ define(function (require, exports, module) {
                 }
             };
 
-            return result.resolve(ingestData);
+            result.resolve(ingestData);
         });
         return result.promise();
     }
--- a/src/extensions/default/HealthData/HealthDataManager.js
+++ b/src/extensions/default/HealthData/HealthDataManager.js
@@ -69,63 +69,83 @@ define(function (require, exports, module) {
                         oneTimeHealthData.bracketsTheme = bracketsTheme;
                     })
                     .always(function () {
-                        var userUuid  = PreferencesManager.getViewState("UUID");
-                        var olderUuid = PreferencesManager.getViewState("OlderUUID");
+                        getSetUserID().done(function (userGuids) {
+                            oneTimeHealthData.uuid  = userGuids.userUuid;
+                            oneTimeHealthData.olderuuid = userGuids.olderUuid;
 
-                        if (userUuid && olderUuid) {
-                            oneTimeHealthData.uuid      = userUuid;
-                            oneTimeHealthData.olderuuid = olderUuid;
                             return result.resolve(oneTimeHealthData);
-                        } else {
-
-                            // So we are going to get the Machine hash in either of the cases.
-                            if (appshell.app.getMachineHash) {
-                                appshell.app.getMachineHash(function (err, macHash) {
-
-                                    var generatedUuid;
-                                    if (err) {
-                                        generatedUuid = uuid.v4();
-                                    } else {
-                                        generatedUuid = macHash;
-                                    }
-
-                                    if (!userUuid) {
-                                        // Could be a new user. In this case
-                                        // both will remain the same.
-                                        userUuid = olderUuid = generatedUuid;
-                                    } else {
-                                        // For existing user, we will still cache
-                                        // the older uuid, so that we can improve
-                                        // our reporting in terms of figuring out
-                                        // the new users accurately.
-                                        olderUuid = userUuid;
-                                        userUuid  = generatedUuid;
-                                    }
-
-                                    PreferencesManager.setViewState("UUID", userUuid);
-                                    PreferencesManager.setViewState("OlderUUID", olderUuid);
-
-                                    oneTimeHealthData.uuid      = userUuid;
-                                    oneTimeHealthData.olderuuid = olderUuid;
-                                    return result.resolve(oneTimeHealthData);
-                                });
-                            } else {
-                                // Probably running on older shell, in which case we will
-                                // assign the same uuid to olderuuid.
-                                if (!userUuid) {
-                                    oneTimeHealthData.uuid = oneTimeHealthData.olderuuid = uuid.v4();
-                                } else {
-                                    oneTimeHealthData.olderuuid = userUuid;
-                                }
-
-                                PreferencesManager.setViewState("UUID",      oneTimeHealthData.uuid);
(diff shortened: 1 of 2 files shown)


vickramdhawal:
@niteskum Can you take a look at these changes ?

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -69,113 +69,141 @@ define(function (require, exports, module) {
...
+            userUuid = PreferencesManager.getViewState("UUID"),
+            olderUuid = PreferencesManager.getViewState("OlderUUID"),
+            userGuids = {};
+
+
+        if (userUuid && olderUuid) {
+            userGuids.userUuid      = userUuid;
+            userGuids.olderUuid = olderUuid;
+
+            return result.resolve(userGuids);
```
vickramdhawal: I think you should be returning  result.resolve(userGuids).promise().

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -69,113 +69,141 @@ define(function (require, exports, module) {
...
+                // assign the same uuid to olderuuid.
+                if (!userUuid) {
+                    userGuids.uuid = userGuids.olderuuid = uuid.v4();
+                } else {
+                    userGuids.olderuuid = userUuid;
+                }
+
+                PreferencesManager.setViewState("UUID",      userGuids.uuid);
+                PreferencesManager.setViewState("OlderUUID", userGuids.olderuuid);
+                return result.resolve(userGuids);
```
vickramdhawal: I think you should be returning result.resolve(userGuids).promise().

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -69,113 +69,141 @@ define(function (require, exports, module) {
...
-                        var userUuid  = PreferencesManager.getViewState("UUID");
-                        var olderUuid = PreferencesManager.getViewState("OlderUUID");
+                        getSetUserID().done(function (userGuids) {
+                            oneTimeHealthData.uuid  = userGuids.userUuid;
+                            oneTimeHealthData.olderuuid = userGuids.olderUuid;
 
-                        if (userUuid && olderUuid) {
-                            oneTimeHealthData.uuid      = userUuid;
-                            oneTimeHealthData.olderuuid = olderUuid;
                             return result.resolve(oneTimeHealthData);
```
vickramdhawal: you can remove the return here and just have result.resolve()

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -69,113 +69,141 @@ define(function (require, exports, module) {
...
+                        olderUuid = userUuid;
+                        userUuid  = generatedUuid;
+                    }
+
+                    PreferencesManager.setViewState("UUID", userUuid);
+                    PreferencesManager.setViewState("OlderUUID", olderUuid);
+
+                    userGuids.uuid      = userUuid;
+                    userGuids.olderuuid = olderUuid;
+                    return result.resolve(userGuids);
```
vickramdhawal: you can remove the return here and just have result.resolve()

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -69,113 +69,141 @@ define(function (require, exports, module) {
...
+                    "event.subtype": eventParams.eventSubType,
+                    "event.user_agent": window.navigator.userAgent || "",
+                    "event.language": brackets.app.language,
+                    "source.name": brackets.metadata.version,
+                    "source.platform": brackets.platform,
+                    "source.version": brackets.metadata.version
+                }
+            };
+
+            return result.resolve(ingestData);
```
vickramdhawal: you can remove the return here and just have result.resolve()

src/extensions/default/HealthData/HealthDataPreview.js
```diff
@@ -42,36 +42,36 @@ define(function (require, exports, module) {
...
+
+                    if (id === "save") {
+                        var newHDPref = $template.find("[data-target]:checkbox").is(":checked");
+                        if (hdPref !== newHDPref) {
+                            prefs.set("healthDataTracking", newHDPref);
+                        }
                     }
-                }
+                });
+                return result.resolve();
```
vickramdhawal: what is this dialog for ? Should the result be resolved once the dialog is shown ?

vickramdhawal (changes requested):

src/extensions/default/HealthData/HealthDataPreview.js
SnchitGrover: dialog is to display the healthData json and adobe analytics json.
result should be resolved when dialog is displayed.

vickramdhawal: then should we not move it inside the function ?

vickramdhawal: or maybe we can add an always() and resolve it there.

src/extensions/default/HealthData/HealthDataManager.js
vickramdhawal: I dont see a need for a return here.

niteskum (approved):

abose:
@SnchitGrover there are quite a lot of redundant commits. can you squash some of them appropriately?

src/extensions/default/HealthData/HealthDataManager.js
```diff
@@ -69,113 +69,140 @@ define(function (require, exports, module) {
...
+        if (userUuid && olderUuid) {
+            userGuids.userUuid      = userUuid;
+            userGuids.olderUuid = olderUuid;
+
+            result.resolve(userGuids);
+        } else {
+
+            // So we are going to get the Machine hash in either of the cases.
+            if (appshell.app.getMachineHash) {
+                appshell.app.getMachineHash(function (err, macHash) {
```
abose: This looks like machine fingerprinting in health reports.  Why is this change made? This looks like it might affect the 'anonymous' part of our health data metrics. UUIDs were deliberately made random to prevent traceability.

SnchitGrover: @abose This functionality was already baked into brackets. I just moved it to a function so that can be called from multiple places.

abose: ah! got it. It just occurred to me that this is machine fingerprinting then. That is not a recommended practice for anonymous logging services. But the legality of that in desktop software might be different to that of hosted services. I think we should consider just using the randomly generated UUID here instead of hardwareID hashes that can uniquely identify a machine.

PR #14390, opened 2018-06-04
Auto Update Unit Test Module

niteskum:
@swmitra @vickramdhawal  @navch @nethip  for review


nethip:
@niteskum Can you come up with a different approach instead of externalizing test related data structures inside code?

niteskum:
@nethip  Changed the Code.Now Test Cases are made pass and fail based on the State in Json File . for this a new function getUpdateStateInJSON is added in main.js , and this function is externalized so that It can be called from unittests Module.
For UnitTesting we are not downloading Installer File from server , we just copy it from local folder  so for this There is one change in "downloadInstaller" function of AutoUpdateDomain.js. 
Added a check If downloadURL is not Http or Https Url , It will just create a ReadStream of the local file and simulate Download Scenario.

src/extensions/default/AutoUpdate/unittests.js
```diff
@@ -0,0 +1,491 @@
...
+
+        _modulePath = FileUtils.getNativeModuleDirectoryPath(module),
+        _unittestFilesFolderName = "unittest-files",
+        _unittestFilesFolder = [_modulePath, _unittestFilesFolderName].join("/"),
+
+        validInstallerFilePaths = {
+            "WIN": [_unittestFilesFolder, "Brackets.test.msi"].join("/"),
+            "OSX": [_unittestFilesFolder, "Brackets.test.dmg"].join("/")
+        },
+        invalidInstallerFilePaths = {
```
nethip: @niteskum Instead of having these test images in the repo, can we think of hosting these in an S3 bucket. That will keep the footprint of the repo low.

src/extensions/default/AutoUpdate/unittests.js
```diff
@@ -0,0 +1,491 @@
...
+                btn;
+
+            // Test if the update bar button has been displayed.
+            expect(doc.find(UPDATE_BAR).length).toBe(1);
+            btn = doc.find(btnId);
+            expect(btn.length).toBe(1);
+
+            if (btn && click) {
+                // We will wait for 3 seconds so that the Update Bar is visible to the tester.
+                waitForMilliSeconds(3000);
```
nethip: @niteskum  I am thinking instead of determining the show state based on a wait period, is it possible to listen to state changes and then act upon that. I am not sure if we have done something like this in the other test cases, but can you check and update if that pattern is followed in any test cases?

src/extensions/default/AutoUpdate/unittests.js
```diff
@@ -0,0 +1,491 @@
...
+        });
+
+
+        it("should attempt download, fail, retry and eventually prompt for failure", downloadFailure);
+
+        /**
+         * Tests the download failure scenario for Auto Update
+         * and validate download failure message in Update Bar prompt.
+         */
+        function downloadFailure() {
```
nethip: Is the intension to simulate a failure scenario or just check if their are any failures? The former one makes more sense and can be simulated by giving a wrong URL.

niteskum: intension is to simulate failure scenario by giving a wrong URL for download failure and passing invalid checksum value for validation failure.

src/extensions/default/AutoUpdate/unittests.js
niteskum: Test Image size is of around 800 kbs , so I think it should be fine to have in repo.

PR #14394, opened 2018-06-04
Updating the license text for immutableJS in Notices

vickramdhawal:


sobisht:
Should we update it in release as well ?

Issue #14411, opened 2018-06-11
[Brackets Auto-Update] Update failed message appears in update notification bar even after a successful update


ayushrathi15:
### Description
[Brackets Auto-Update] Update failed message appears in update notification bar even after a successful update

### Steps to Reproduce
1. Launch brackets 1.13.
2. Create a new file and type some text. Do not save this file.
3. Go to help > Check for updates.
4. Update dialogue appears. Click on Get it now.
5. Click on Update Now.
6. Brackets will close and update will begin.
7. Brackets will auto launch once update is done.
8. Update failed message appears in update notification bar even after a successful update.

**Expected behavior:** User should get a update successful message in update notification bar on successful update of brackets.

**Actual behavior:** Update failed message appears in update notification bar even after a successful update

Note : This issue has been encountered 3 times so far.

### Versions
Windows 10 64 Bit
Mac 10.13
Release 1.13 build 1.13.0-17624

shubhsnov:
Fixed in https://github.com/adobe/brackets/pull/14412

PR #14412, opened 2018-06-11
Auto Update Error Handlng Fix

niteskum:


niteskum:
ping @nethip @vickramdhawal  for review

nitheshgangadharsalian:
Bug: https://github.com/adobe/brackets/issues/14411

nethip:
@niteskum Could you provide description here as to how the bugs has been fixed?

src/extensions/default/AutoUpdate/main.js
```diff
@@ -321,15 +322,14 @@ define(function (require, exports, module) {
...
+        updateDomain.on('data', receiveMessageFromNode);
 
         updateDomain.exec('initNode', {
             messageIds: MessageIds,
             updateDir: updateDir,
             requester: domainID
         });
 
-        updateDomain.on('data', receiveMessageFromNode);
-        initState();
```
nethip: `initState()` not required anymore?

niteskum:
@nethip  Since initNode() and Brackets Side initState()  function were called one after another.  So there were no synchronization between setUpdateInProgressState() and CheckUpdateStatus()
InitNode Function notifies Brackets to reset inProgress Flag when Bracket is launched. and initState() function used to call checkUpdateFunction() which cleans up Update Temp Folder If Update was initiated in prev session. sometime UpdateTemp were being cleaned up before setting up the flags.  and Exception were thrown in setUpdateInProgress() function because there was no File to write.

Changes in this PR: 
Renamed the function "setAutoUpdateInProgressFlag" to "nodeDomainInitialized". and from initNode Function Node Notifies to Brackets to execute "nodeDomainInitialized" function and pass a parameter "reset" which is true If Brackets is launched. and false If New Brackets Window is opened.
and in nodeDomainInitialized function , initState() Function is called . and resetting of "InProgress"Flag and  checkUpdateStatus() are executed in sequence.
so this makes sure There is No RaceCondtion between resetting the flag and cleaning up the temp folder.

src/extensions/default/AutoUpdate/main.js
niteskum: Now initState() is called from nodeDomainInitialized() function. so it is not required  here.

src/extensions/default/AutoUpdate/main.js
```diff
@@ -594,11 +594,17 @@ define(function (require, exports, module) {
...
+    function nodeDomainInitialized(reset) {
+        if(reset) {
+            updateJsonHandler.parse()
+                .done(function() {
+                    setUpdateStateInJSON("autoUpdateInProgress", !reset)
+                        .always(initState);
+                 })
+                 .fail(initState);
+        } else {
+            initState();
```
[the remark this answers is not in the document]
niteskum: Restructured the Code. , Removed the CheckUpdateStatus() from initNode() function. Now initNode Function returns Promise . which is resolved when parsing is done. initNode is called and in done function checkUpdateStatus() function is called.

nethip (approved):

PR #14413, opened 2018-06-12
Allow undo/redo persistence between Brackets sessions - #10222 -

nyteksf:
_Topic: Issue #10222 - "Easy to hit Reload by accident, losing Undo history for files that don't have unsaved changes"_

About: This patch uses localStorage to preserve the CodeMirror history between sessions, and until file is renamed or deleted within Brackets. It was created in response to @peterflynn's second part of one of his replies, as stated: 

> **If you do reload Brackets, or quit & relaunch, etc., it would be cool if the Undo history were preserved**.

Happy to help out with this awesome tool--thanks again for it. If there are any suggestions or needed changes, please let me know, and I'll get right to them.


nyteksf:
As a PS, if it would be preferred, I can quickly change this feature to also delete localStorage for the file being handled on each save. Doing so would ensure only each documents unsaved changes are preserved in localStorage instead. As of now, this will preserve the entire undo/redo history between saves, as already stated.

nethip:
@nyteksf Awesome!

nyteksf:
Any feedback or problems with this, please let me know. That aside, maybe this could be two prefs? Like, as in, "persist undo/redo history for files not yet saved", and another, "persist undo/redo history between sessions"? I only ask as I can imagine people wanting either option at different times.

nyteksf:
@nethip, whoops! I missed your comment until now somehow. Thank you, by the way! That aside, I'll get right on adapting this PR to persist all unsaved changes instead.

Issue #14414, opened 2018-06-12
Find in files does not work if their is no file opened in working set.


ayushrathi15:
### Description
Find in files does not work if their is no file opened in working set.

### Steps to Reproduce
1. Launch brackets 1.13.
2. Close all the files from working files set in file tree.
3. Hit Ctrl+shift+F or Go to Find > Find in Files.
4. Find in files do not work.

**Expected behavior:** Find in files should work even if their is no file in working files in file tree.

**Actual behavior:** Find in files does not work if their is no file opened in working set.

### Versions
Windows 10 64 Bit
Mac 10.13
Release 1.13 build 1.13.0-17693

ayushrathi15:
Tested on Release Build 17696.

PR #14416, opened 2018-06-12
Fix for FindInFiles UI not opening when no file is open 

sobisht:
Issue https://github.com/adobe/brackets/issues/14414

### DIFF
--- a/src/search/FindBar.js
+++ b/src/search/FindBar.js
@@ -705,11 +705,8 @@ define(function (require, exports, module) {
      */
     FindBar.getInitialQuery = function (currentFindBar, editor) {
         var query,
-            selection = "",
+            selection = editor ? FindBar._getInitialQueryFromSelection(editor) : "",
             replaceText = "";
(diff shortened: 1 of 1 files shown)


src/search/FindBar.js
```diff
@@ -705,8 +705,11 @@ define(function (require, exports, module) {
      */
     FindBar.getInitialQuery = function (currentFindBar, editor) {
         var query,
-            selection = FindBar._getInitialQueryFromSelection(editor),
+            selection = "",
```
navch: nit clean: Please use ternary operator, 
`selection = editor ? FindBar._getInitialQueryFromSelection(editor) : "";`

navch (approved):
Good quick fix 👍

src/search/FindBar.js
sobisht: Thanks for pointing it out. Forgot to add the ternary operator. Corrected.

nethip (approved):

PR #14421, opened 2018-06-13
Fix for Title bar text and Save button issue for empty active pane

sobisht:
- _updateTitle() failed when there is no file opened in active pane. Causing the title bar retaining the text of last opened document as title. Added a null check for that.
- _setMenuItemsVisible() failed when there is no file opened in active pane. Causing the state of file menu options same as for previous file. Added a check to enable menu options if no file is present in active pane.

### DIFF
--- a/src/document/DocumentCommandHandlers.js
+++ b/src/document/DocumentCommandHandlers.js
@@ -158,8 +158,8 @@ define(function (require, exports, module) {
         var currentDoc          = DocumentManager.getCurrentDocument(),
             windowTitle         = brackets.config.app_title,
             currentlyViewedFile = MainViewManager.getCurrentlyViewedFile(MainViewManager.ACTIVE_PANE),
-            currentlyViewedPath = currentlyViewedFile.fullPath,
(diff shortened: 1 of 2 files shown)


boopeshmahendran (approved):

src/document/DocumentCommandHandlers.js
```diff
@@ -158,8 +158,8 @@ define(function (require, exports, module) {
         var currentDoc          = DocumentManager.getCurrentDocument(),
             windowTitle         = brackets.config.app_title,
             currentlyViewedFile = MainViewManager.getCurrentlyViewedFile(MainViewManager.ACTIVE_PANE),
-            currentlyViewedPath = currentlyViewedFile.fullPath,
```
navch: Good fix 👍  Actually one console error was there because of this line on project change. Can you please verify that as well?

sobisht: Yes, this will fix the console error also.

navch (approved):

Issue #10779, opened 2015-03-22
Mashing F5 crashes Brackets


petetnt:
Found this small issue that most won't run into when testing out for #10769 but anyway:

(Windows 8.1., Brackets 1.2.)
1. Open Brackets
2. Repeatedly reload the project with `F5`
3. Soon-ish Brackets hangs and can only be terminated via Task Manager: 

![image](https://cloud.githubusercontent.com/assets/7641760/6769358/adf4d2ce-d0a0-11e4-9d22-7fdbee6cc66a.png)

prksingh:
Am able to reproduce on Windows 7 as well. Thanks for reporting. Adding 'move to backlog' so we can assign some priority to this. Looks like a very rare scenario so may not be an urgent fix.

ryanstewart:
Marking low priority.

petetnt:
Just noticed when trying out #10906 that holding down F5 will crash Brackets within seconds

sagiegurari:
Why is this one low priority? When you install an extension it results in reload and can freeze brackets.
The only way out is to kill the brackets process.
After 6 months with this issue, I'm not sure that anyone is looking at this.

abose:
This has been fixed in 1.5. Please update to latest version :+1:

abose:
Feel free to reopen this issue if you find this in the latest version.

sagiegurari:
I saw the setTimeout workaround, but I still reproduced it with 1.5
Are you 100% sure this doesn't happen to others with 1.5?

petetnt:
I am almost sure (I haven't installed the newest build yet that ships with different `shell` I have on now...) that this particular bug is similar, but not fixed by #11505. Smashing F5 still crashes Brackets for good: easiest way to reproduce this is by holding F5 while Brackets is active.

petetnt:
Confirmed against latest master with the newest shell: The #11505 helps but does not completely fix this issue.

abose:
Does this happen when you press f5 and then while brackets is reloading, press f5 again- that is causing a crash in my machine. But simply pressing f5 and waiting for the reload to complete does not cause the crash.

abose:
Maybe a fix would be to hold off reloading till brackets app ready is done?

sagiegurari:
Yes exactly

petetnt:
@abose I think yes. Currently holding down `F5` constantly reloads the page (it's Chromium behaviour), so debouncing the reload event could help too.

PR #14429, opened 2018-06-16
Debounce handleReload(), Add delay on load to prevent premature reloading - #10779 -

nyteksf:
_Issue #10779 - Mashing F5 crashes Brackets_

This patch was made to handle two aspects of triggering a crash when F5 is mashed and pressed prematurely on reload.

I started by debouncing the handleReload() function itself. This change aims to prevent those who will mash F5 from being able to crash brackets as a consequence of triggering a premature reload.

Second to that--I found that I wasn't able to make sure the app had finished loading before allowing handleReload() to wind up being fired by the user prematurely. Even though I had used AppInit.appReady(...) to delay the loading of the two related handleReload() functions until the app was allegedly totally ready, Brackets still kept crashing on reload unless I'd waited a few seconds after that point. That tells me there's some kind of behind the scenes, unrelated process someone might know better than I about which loading can interrupt. The best workaround solution I found for this aspect of the problem was to add a delay on load of several seconds before the handleReload() functions were programmatically made available by AppInit.appReady(). This delay makes it way more unlikely that the user will crash brackets by forcing them to wait until the app has most likely finished loading before allowing a successful reload to take place as pressing F5 will not have any effect before then.


petetnt:
Very cool @nyteksf  👍

nyteksf:
Thanks @petetnt. That said, do you have any feedback or suggestions, if I may?

nyteksf:
@petetnt If I could ask for your help, what would need to be done for this PR to be merged?

petetnt:
Sorry @nyteksf, I haven't been deeply involved with Brackets in a while. Maybe some of the current contributors could answer here or at Slack 🤔

shubhsnov:
@nyteksf I apologize for viewing this so late. Do you still plan to keep this PR valid? I can take a look at it but the force push seems to have overwritten your changes.

PR #14462, opened 2018-07-04
Allow persistence for unsaved file changes in Brackets - #10222 -

nyteksf:
**Topic:**
Inspired By Issue #10222 -  _"Easy to hit Reload by accident, losing Undo history for files that don't have unsaved changes"_ 

**About:**
It was confirmed by @nethip in my prior PR ([#14413](https://github.com/adobe/brackets/pull/14413)) that my code should be modified to persist unsaved changes. This PR therefore aims to implement a preference based feature useful for retaining unsaved file changes similar to what is found in Sublime Text, Atom, or etc.

Here, in Brackets, this is accomplished through localStorage. Any and all changes to a given document will continue to sync to CodeMirror as usual--but a secondary sync to localStorage will also occur simultaneously. With "Hot Close" mode enabled, Brackets closes any file and exits without the usual "save/cancel/don't save" dialog appearing, and the user can simply continue from wherever they've left off whenever those files are reopened.

To manage space in localStorage, each unsaved file's undo/redo history, latest text, cursor position and scroll position are deleted upon save events, on renaming the file (through brackets), or deleting said file (similarly through brackets). The text that's stored is compressed first, and special characters are encoded. There is also a feature added to wipe localStorage through button click via dialog in the event that localStorage ever manages to fill up over time.

Any feedback, help, comments or suggestions would be welcomed. Thanks for your time!


nyteksf:
Can someone please advise? This doesn't seem to make sense to me for some reason. For Brackets to work in this PR, I have to require the compression utils as anything other than 'RawDeflate', and yet I still must use it in my code as 'RawDeflate.encode()', or 'RawDeflate.decode()', despite that difference.

The strangest thing to me is that requiring it as "RawDeflate" lets the Travis CI tests pass, but causes Brackets itself to break.

navch:
Hi @nyteksf  Thanks For this PR 👍 . I triggered the build again. But This Travis CL build doesn't runs all tests. You should run all tests locally. Anyways I will try this out soon.

src/document/DocumentCommandHandlers.js
```diff
@@ -49,15 +50,18 @@ define(function (require, exports, module) {
...
         Inspector           = require("LiveDevelopment/Inspector/Inspector"),
         Menus               = require("command/Menus"),
         UrlParams           = require("utils/UrlParams").UrlParams,
         StatusBar           = require("widgets/StatusBar"),
         WorkspaceManager    = require("view/WorkspaceManager"),
         LanguageManager     = require("language/LanguageManager"),
-        _                   = require("thirdparty/lodash");
+        _                   = require("thirdparty/lodash"),
+        CompressionUtils    = require("thirdparty/rawinflate"),
+        CompressionUtils    = require("thirdparty/rawdeflate"),
```
navch: Rename CompressionUtils with RawDeflate

src/editor/Editor.js
```diff
@@ -59,99 +59,120 @@
...
+        ViewUtils = require("utils/ViewUtils"),
+        MainViewManager = require("view/MainViewManager"),
+        DocumentManager = require("document/DocumentManager"),
+        FileUtils = require("file/FileUtils"),
+        StringUtils = require("utils/StringUtils"),
+        Dialogs = require("widgets/Dialogs"),
+        DefaultDialogs = require("widgets/DefaultDialogs"),
+        ProjectManager = require("project/ProjectManager"),
+        _ = require("thirdparty/lodash"),
+        CompressionUtils = require("thirdparty/rawdeflate"),
```
navch: Again Rename CompressionUtils with RawDeflate.

navch:
/home/travis/build/adobe/brackets/src/document/DocumentCommandHandlers.js
  591:67  error  'RawDeflate' is not defined  no-undef
  593:56  error  'RawDeflate' is not defined  no-undef
/home/travis/build/adobe/brackets/src/editor/Editor.js
    92:17  error  'RawDeflate' is not defined  no-undef
    92:29  error  'RawInflate' is not defined  no-undef
   960:30  error  'RawDeflate' is not defined  no-undef
   966:30  error  'RawDeflate' is not defined  no-undef
   989:45  error  'RawDeflate' is not defined  no-undef
   997:66  error  'RawDeflate' is not defined  no-undef
  1006:37  error  'RawDeflate' is not defined  no-undef
  1012:46  error  'RawDeflate' is not defined  no-undef
  1115:58  error  'RawDeflate' is not defined  no-undef
  1119:54  error  'RawDeflate' is not defined  no-undef
  1137:51  error  'RawDeflate' is not defined  no-undef

You need to fix these ESLint errors, by addressing review comments.

src/document/DocumentCommandHandlers.js
nyteksf: Hi @navch. Firstly, thank you for the quick reply! That aside, I am guessing you may not have seen my comment up above, but I'm having a unique problem when renaming to "RawDeflate" and trying to call RawDeflate.encode() or ...decode(). Hopefully you can help? I've never had this problem before. Require isn't working as expected.

Whenever I change variable name "CompressionUtils" to "RawDeflate", the Travis CI tests will pass, but Brackets stops working. When I run the code it says that "RawDeflate" is undefined in the console in dev tools when I call RawDeflate.encode() or .decode() as I was saying above. This doesn't make sense to me. Expected behavior is that if I require as RawDeflate, then RawDeflate.encode() or any other method should work.

However, when I change "RawDeflate" back to being literally any other variable name, Brackets will start to work again, but then the Travis CI tests fail because the variable names don't match. So this means even if I do ->
`CompressionUtils = require("thirdparty/rawinflate")`, I can't actually call CompressionUtils.inflate(). I still have to use "RawDeflate.inflate()", somehow, and somehow it works.

If you have a moment, can you please help?

nyteksf:
Also, @navch, in relation to you suggesting I do testing locally as well, I agree it's important. Although I seem to not be able to run those precommit tests on my machine for some reason. I've mentioned this before in another PR, but there's been no fix offered to date. The only workaround I have been given is to run "--no-verify" to skip the local tests as was suggested to me. Is there a way to manually run the tests you could help me with please? I would be happy to do this from now on in my future PRs.

navch:
Yes you can run unit test  by clicking "Debug -> Run test" (some tests will be failed for some memory or other reason but verify no extra test fails because of these changes)
This will open specrunner there you can run tests for everything using UI.

Pre-commit tests are nothing but ESLint checks. You should not do force pas to this. By reading your comment I am not able to understand exact problem. I will clone you branch will try it out. But right now I am quit busy, I will get back to you by Friday. Is that fine?

nyteksf:
@navch - Okay, I see now. Thanks for sharing that with me. I just now ran all tests on a version of Brackets compiled using Adobe's official repo, and then once again I tested using my own fork/branch used in this PR. Both versions yielded the same number of errors -- 38 in total. So it seems no new errors were introduced. 

As for ESLint checks crashing my prior attempts to push, I just now got ESLint to work by manually installing grunt first after inspecting package.json. And then I had to change "`grunt eslint`" to simply, "`grunt`" under my package.json -> `"scripts": {...` section. It should work when I push now. So all tests can be done.

As for not totally understanding my problem--I know it's a bit confusing. I'm confused here too. In my code, when I require the compression utils into var "RawDeflate", it breaks my code when I then go to call "RawDeflate.deflate()", for example. You would expect that it would work. But I for some reason get an error saying RawDeflate is undefined in that case. However, Travis CI says it passes all the tests. The second strange part of this is how no matter what I name the variable I require into, I still can only refer to its methods in my code as "RawDeflate.methodHere()".

Otherwise, yes, I understand completely about being busy. And I'm definitely looking forward to your help whenever you can. Thanks again.

PR #14504, opened 2018-08-03
Show full path on Brackets top bar (#13987)

shounak1:
Fix for #13987


shounak1:
Can anyone explain me what the error is ?

PR #14520, opened 2018-08-24
jsRefactor Rename Relative Path issue fixed

niteskum:
Rename doesn't work if TernNodeDomain return Refs with path relative to project Root.
So added check it will compare both Absolute path and Relative path so if either of these is true Refs will be considered in same file.

@boopeshmahendran @navch  please review

### DIFF
--- a/src/extensions/default/JavaScriptRefactoring/RenameIdentifier.js
+++ b/src/extensions/default/JavaScriptRefactoring/RenameIdentifier.js
@@ -29,7 +29,8 @@ define(function (require, exports, module) {
         Session              = brackets.getModule("JSUtils/Session"),
         MessageIds           = brackets.getModule("JSUtils/MessageIds"),
         TokenUtils           = brackets.getModule("utils/TokenUtils"),
-        Strings              = brackets.getModule("strings");
+        Strings              = brackets.getModule("strings"),
+        ProjectManager      = brackets.getModule("project/ProjectManager");
 
     var session             = null,  // object that encapsulates the current session state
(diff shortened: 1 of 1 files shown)


src/extensions/default/JavaScriptRefactoring/RenameIdentifier.js
```diff
@@ -98,7 +99,17 @@ define(function (require, exports, module) {
 
         function isInSameFile(obj, refsResp) {
             // In case of unsaved files, After renameing once Tern is returning filename without forward slash
-            return (obj && (obj.file === refsResp.file || obj.file === refsResp.file.slice(1, refsResp.file.length)));
+            var projectDir = ProjectManager.getProjectRoot().fullPath,
```
boopeshmahendran: Check if `ProjectManager.getProjectRoot()` returns null

src/extensions/default/JavaScriptRefactoring/RenameIdentifier.js
```diff
@@ -98,7 +99,17 @@ define(function (require, exports, module) {
...
+                fileName = "";
+
+            // get the relative path of File as Tern can also return
+            // references with file name as a relative path wrt projectRoot
+            // so refernce file will be compared with both relative and Absolute path ti check if it is same file
```
boopeshmahendran: nit: fix typos

boopeshmahendran (changes requested):

boopeshmahendran (approved):

Issue #14533, opened 2018-09-14
When closing a files and window on left vertical split right side disappears as well


Niklas-Lundborg:
### Prerequisites

* [x] Can you reproduce the problem with `Debug -> Reload Without Extensions`?
* [ ] Did you perform a cursory search to see if your bug or enhancement is already reported?
* [x] Did you read the [Troubleshooting guide](https://github.com/adobe/brackets/wiki/Troubleshooting)?

For more information on how to write a good bug report read [here](https://github.com/adobe/brackets/wiki/How-to-Report-an-Issue)
For more information on how to contribute read [here](https://github.com/adobe/brackets/blob/master/CONTRIBUTING.md)

### Description

When having a vertical split open I sometimes want to close the split and just focus on the split on the right side. When closing both the actual file and window on the left side instead of showing just the right side both windows are closed.

### Steps to Reproduce

1. [View -> Split vertical]
2. [Open files for both windows]
3. [Close left window on the "x" on the top right corner]
4. [Do the same thing again so that the left spilt window closes]

**Expected behavior:** 
Show only right window

**Actual behavior:**
Close both windows but file from right window is still left in working files.

Brackets Version 1.14
Mac OS High Sierra 10.13.6

Niklas-Lundborg:
Could this be a good contribution to start with?
If so, could anyone point me in the right direction?  :)

Niklas-Lundborg:
Uploaded a pull request with my fix :)
Hope it's ok!

PR #14534, opened 2018-09-14
MainViewManager shows right currently viewed file after left split closes

Niklas-Lundborg:
Fixed MainViewManager so that if you close the left split (firstPane) and open files exist in the right split (secondPane), then the file shown in secondPane will be shown in firstPane.
Before it was only done one-sidedly.

My first pull request for Brackets, hope I did ok!

Issue: #14533


Niklas-Lundborg:
Hmm, is my npm not up to date or something?

PR #14546, opened 2018-09-24
Add styles for stickyhighlight feature in livepreview

boopeshmahendran:


src/LiveDevelopment/Agents/RemoteFunctions.js
```diff
@@ -1096,11 +1115,31 @@ function RemoteFunctions(config, remoteWSPort) {
...
+        _hoverHighlight.add(element, false, true);
+    }
+
 
     function createWebSocket() {
         _ws = new WebSocket("ws://localhost:" + remoteWSPort);
         _ws.onopen = function () {
             window.document.addEventListener("click", onDocumentClick);
+            if (!experimental && config.stickyHighlight) {
+                window.document.addEventListener("mouseover", onDocumentHover);
```
vickramdhawal: I think you would need to listen in the capture phase.

vickramdhawal: Also, when updateConfig is called, the event listener should be attached even at that time and disengaged when the config changes again.

vickramdhawal (changes requested):

boopeshmahendran:
@vickramdhawal Thanks for the quick review. I have addressed the review comments

swmitra (commented):
Border would potentially hide/override any Broder used in the actual UI. The ideal way to highlight is either by using a outline or drawing an overlay div by constructing an overlay pane on top of the entire page. Then use element from point of mouse event and use the targets bounding client rect to draw the highlight overlay. Usage of the overlay pane is required to prevent click on an actual link or button.

boopeshmahendran:
Thanks for the review @swmitra The approach you have described is the exact one followed as you can see [here](https://github.com/adobe/brackets/blob/aff17281c33762a2070aea93a58249ec9f1ccb3f/src/LiveDevelopment/Agents/RemoteFunctions.js#L288) . The border is on the overlay div.

swmitra:
But still not in an overlay pane. However use outline and the actual elements bounding box for positioning the highlight overlay and styling.

vickramdhawal:
The current implementation follows the existing approach. Are you suggesting to introduce an overlay pane overall for both the highlight modes ?

vickramdhawal (approved):

PR #14550, opened 2018-09-30
Clean up unit tests that wait for a promise

HaraldNordgren:
Updates #4638.


HaraldNordgren:
@vickramdhawal @boopeshmahendran

Are these the kind of changes we want. If so, let's merge this one and I will move on to the rest.

PR #14555, opened 2018-10-05
Add css scroll-snap-type

abaracedo:
This adds the CSS `scroll-snap-type` property and its values for autocompletion.


narayani28:
LGTM

PR #14563, opened 2018-10-10
Update CSSProperties.json for text-align and text-justify

narayani28:
Adding css3 new values start, end, match-parent, justify-all to text-align property and adding the property text-justify with its values.
https://www.w3.org/TR/css-text-3/


src/extensions/default/CSSCodeHints/CSSProperties.json
```diff
@@ -209,6 +209,7 @@
     "text-emphasis-position":      {"values": ["above", "below", "left", "right"]},
     "text-emphasis-style":         {"values": ["circle", "dot", "double-circle", "filled", "none", "open", "sesame", "triangle"]},
     "text-indent":                 {"values": ["inherit"]},
+    "text-justify":                {"values": ["auto", "none", "inter-word", "inter-character"]},
```
vickramdhawal: we have "inherit" as an option for most of the properties. maybe we can add here as well since it is applicable.

vickramdhawal (approved):
LGTM 👍

swmitra (approved):

PR #14603, opened 2018-11-29
Disabling JS code insertion due JSRefactoring into PHP document or in any non-js scope

g-217:
JSRefactoring was inserting JS code into PHP document while using option wrap in condition/try-catch.

js-refactoring_in_php

Added check to make wrap in condition/try-catch in non-js context as no-op.

### DIFF
--- a/src/extensions/default/JavaScriptRefactoring/ExtractToFunction.js
+++ b/src/extensions/default/JavaScriptRefactoring/ExtractToFunction.js
@@ -271,7 +271,7 @@ define(function(require, exports, module) {
             result.resolve(Strings.ERROR_EXTRACTTO_FUNCTION_MULTICURSORS);
             return;
         }
-        if (!editor || editor.getModeForSelection() !== "javascript") {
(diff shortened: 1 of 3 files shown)


gautam0217:
closing now and will reopen later, so that Travis-CI build will be triggered again.
https://github.com/travis-ci/travis-ci/issues/576#issuecomment-65317915

navch:
Thanks @gautam0217 for this PR. But I guess disabling menu in all documents (except js) will be a better solution.

navch:
It looks like cla is not signed that's why build is failing. and you can also trigger build after logging in there.

gautam0217:
As we support mixed mode(e.g. HTML+JS etc.) in refactoring, making operations no-op is more aligned to current approach.

gautam0217:
Added generic error message for each case like following:

try-catch-err-msg

navch:
Is it possible to add a generic error like As of now Refactoring is available only for JS?

PR #14605, opened 2018-12-03
Executing default update process handler, in case auto-update fails.

nethip:
Added error checks to the auto update mechanism. So in case the auto update mechansim fails, we will now give chance to the default update process Handler to handle the update mechanism (Which is essentially taking the user to brackets.io).

@swmitra Could you please review this?


swmitra (approved):
LGTM.

PR #14665, opened 2019-03-11
 Add a new translation for khmer language

yeohsoonkeat:
Adding khmer(Cambodian language) translation


shubhsnov:
@swmitra @narayani28 Should we get this verified?

PR #14606, opened 2018-12-04
Language Server Protocol Support for Brackets

jhasubhash:
This covers the first cut of Language Server Protocol Support for Brackets.

**Basic core implementation to:**

- Initiate/Stop a Language Server
- Interact with Language Server
  - Completions
  - Signature Help
  - Diagnostics
  - Jump To Dec/Decl/Impl
  - Symbols
  - References
  - Synchronization Events (Text Document didOpen, didChange etc.)
  - Custom Request/Notification (so the developer can use commands not exposed in the initial implementation)
- Bidirectional Request/Notification system for communication between Brackets and Language Server
- Abstraction of LSP from Brackets (as far as possible)
- Creation of Language Client that works as the handle to Language Server from Brackets side
- Support for all communication types: IPC, Pipe, Socket and Standard I/O

**TODO:**
- [x] Move vscode-languageserver-protocol to Thirdparty
- [x] Tooling Manager
  - [x] CodeHints
  - [x] ParameterHints
  - [x] Diagnostics
  - [x] Jump to Def/Decl/Impl
  - [x] Priority
  - [x] Translate brackets events to LSP events
  - [x] Default handlers for all the above Tooling Interfaces
- [x] Project level event handling
  - [x] Project Switch
  - [x] Reload with Extensions
- [x] Grunt modifications for build generation
- [x] UI
- [x] Tests

### This feature can be experienced with the reference Language Client implementation for PHP here : https://github.com/adobe/brackets/pull/14671

PR #14671, opened 2019-03-20
Php Tooling Extensions Using LSP Framework

niteskum:
Language Server Client Extension using the https://github.com/felixfbecker/php-language-server

**Requirements**: 
- #14606 
- Php 7
- Composer 1.8.4

**Installation:** `npm install`

Set the following preferences in 

> brackets.json

 to configure the extension:

```
"php": {
    "enablePhpTooling": true,             // Enable/Disable PHP Tooling
    "executablePath": "php",              //Path of the php 7 executable
    "memoryLimit": "4095M",               //Memory limit for the Language Server Process
    "validateOnType": "false"             //Get validation results on Type or on Save
}
```


niteskum:
This PR is Using LSP Framework which are being reviewed in separate PR #14606 .

@shubhsnov @swmitra please review

src/extensions/default/PhpTooling/main.js
```diff
@@ -0,0 +1,192 @@
...
+        CodeInspection.register(["php"], {
+            name: "Diagnostics",
+            scanFile: lProvider.getInspectionResults.bind(lProvider)
+        });
+
+        languageClient.addOnDiagnostics(lProvider.setInspectionResults.bind(lProvider));
+    }
+
+    function startPhpServer() {
+        if(_client && phpConfig["enablePhpTooling"]) {
```
shubhsnov: Two suggestions:

- Decouple the validation logic from this function. You can do it on App Ready and then prompt the UI on opening a PHP file the first time. Save that to state.

- I know it seems like a nice optimization to start the server on first open of PHP file, but it doesn't seem like valid enforcement if the extension itself enforces it. Since this is a reference implementation, we'd want to provide the developer with a way where they can defer the activation of the server if they want from the brackets core itself. This is a nice optimization which other LSP extension can leverage as well.

shubhsnov:
Another thing to note: To test the validation logic against PHP5 and PHP7. Both should give different results, based on syntax.

src/extensions/default/PhpTooling/client.js
```diff
@@ -0,0 +1,130 @@
...
+/*global exports */
+/*global process */
+/*eslint-env es6, node*/
+/*eslint max-len: ["error", { "code": 200 }]*/
+"use strict";
+
+var LanguageClient = require(global.LanguageClientInfo.languageClientPath).LanguageClient,
+    net = require("net"),
+    cp = require("child_process"),
+    execa = require("execa"),
```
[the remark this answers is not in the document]
niteskum: yes execa and semver

src/extensions/default/PhpTooling/client.js
```diff
@@ -0,0 +1,130 @@
...
+                reject(["PHP_PROCESS_SPAWN_ERROR", err.code]);
+                console.error(err);
+            }
+            return;
+        });
+    });
+}
+
+var serverOptions = function (){
+        return new Promise(function (resolve, reject) {
```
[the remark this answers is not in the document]
niteskum: indentation is correct

src/extensions/default/PhpTooling/client.js
```diff
@@ -0,0 +1,130 @@
...
+    clientName = "PhpClient",
+    executablePath = "",
+    memoryLimit = "";
+
+function validatePhpExecutable(confParams) {
+    executablePath = confParams["executablePath"] ||
+        confParams["validate_executablePath"] ||
+        (process.platform === 'win32' ? 'php.exe' : 'php');
+
+    memoryLimit = confParams["memoryLimit"] || 'default';
```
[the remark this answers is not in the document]
niteskum: fixed

src/extensions/default/PhpTooling/client.js
```diff
@@ -0,0 +1,130 @@
...
+                });
+                server.close();
+                resolve({
+                    reader: socket,
+                    writer: socket
+                });
+            });
+            // Listen on random port
+            server.listen(0, '127.0.0.1', function () {
+                var pathToPHP = __dirname + "/vendor/bin/php-language-server.php";
```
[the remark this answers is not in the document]
niteskum: fixed

src/extensions/default/PhpTooling/main.js
```diff
@@ -0,0 +1,192 @@
...
+        Dialogs             = brackets.getModule("widgets/Dialogs"),
+        DefaultDialogs      = brackets.getModule("widgets/DefaultDialogs"),
+        Commands               = brackets.getModule("command/Commands"),
+        CommandManager         = brackets.getModule("command/CommandManager"),
+        StringUtils             = brackets.getModule("utils/StringUtils");
+
+    var clientFilePath = ExtensionUtils.getModulePath(module, "client.js"),
+        clientName = "PhpClient",
+        _client = null,
+        phpConfig = {},
```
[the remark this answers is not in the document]
niteskum: fixed

src/extensions/default/PhpTooling/main.js
```diff
@@ -0,0 +1,192 @@
...
+        memoryLimit: "default"
+    }, {
+        description: Strings.DESCRIPTION_PHP_TOOLING_CONFIGURATION
+    });
+
+    PreferencesManager.on("change", "php", function () {
+        var newPhpConfig = PreferencesManager.get("php");
+
+        if((newPhpConfig["executablePath"] !== phpConfig["executablePath"])
+            || (newPhpConfig["enablePhpTooling"] !== phpConfig["enablePhpTooling"]))  {
```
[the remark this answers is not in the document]
niteskum: removed validation_executablePath , as it is not required.

src/extensions/default/PhpTooling/main.js
niteskum: I think we should always validate Server config before starting. 
otherwise if server start fails for any reason (e.g user set wrong executable Path), we wont be able to notify user about failure.

shubhsnov:
@niteskum Please add support for PHP static code hinting as part of this on priority, for completeness.

shubhsnov:
Some other required features for reference impl:
Currently, there are two ways to handle the project root change, either restart or send folder change event. Tap into the server capability which is received on client start to figure out whether the server can handle the folder change event, and then decide dynamically which way to go.

Same for completion resolve.

niteskum:
@shubhsnov I can see all static code hints and Parameter hints comes from LSP Server except below few variables.
$GLOBALS
$_SERVER
$_GET
$_POST
$_FILES
$_REQUEST
$_SESSION
$_ENV
$_COOKIE
$php_errormsg
$HTTP_RAW_POST_DATA
$http_response_header
$argc
$argv
$this

are we sure we have to include this in code hints.??, There might be some reason lsp server is not giving these variables as hints.

src/extensions/default/PhpTooling/main.js
```diff
@@ -0,0 +1,192 @@
...
+
+    var clientFilePath = ExtensionUtils.getModulePath(module, "client.js"),
+        clientName = "PhpClient",
+        _client = null,
+        phpConfig = {},
+        DEBUG_OPEN_PREFERENCES_IN_SPLIT_VIEW  = "debug.openPrefsInSplitView";
+
+    PreferencesManager.definePreference("php", "object", {
+        enablePhpTooling: true,
+        showDiagnosisOnType: false,
```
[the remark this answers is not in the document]
shubhsnov: ?

src/extensions/default/PhpTooling/main.js
shubhsnov: It's up to you. I find the current way a bit clumsy.

shubhsnov:
> @shubhsnov I can see all static code hints and Parameter hints comes from LSP Server except below few variables.
> $GLOBALS
> $_SERVER
> $_GET
> $_POST
> $_FILES
> $_REQUEST
> $_SESSION
> $_ENV
> $_COOKIE
> $php_errormsg
> $HTTP_RAW_POST_DATA
> $http_response_header
> $argc
> $argv
> $this
> 
> are we sure we have to include this in code hints.??, There might be some reason lsp server is not giving these variables as hints.

Get this resolved. My take is that most of these appear to quite common globals. It would be good to have.

src/extensions/default/PhpTooling/main.js
```diff
@@ -0,0 +1,204 @@
...
+            startPhpServer();
+            return;
+        }
+        phpConfig = newPhpConfig;
+    });
+
+    var handleProjectOpen = function (event, directory) {
+        _client.stop()
+            .done(function () {
+                setTimeout(function () {
```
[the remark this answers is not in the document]
niteskum: It does not work without timeout function, Can you check from core side.
Better provide a restart function so Extension Developer need not to worry about these things

niteskum:
@shubhsnov  Regarding FolderChange Event my understanding is "workspace/didChangeWorkspaceFolders" event is for workspace folder structure change like (folder added or removed from workspace ), not for the workspace project root folder change.

correct me if my understanding is wrong.

shubhsnov:
@niteskum 
> @shubhsnov Regarding FolderChange Event my understanding is "workspace/didChangeWorkspaceFolders" event is for workspace folder structure change like (folder added or removed from workspace ), not for the workspace project root folder change.
> 
> correct me if my understanding is wrong.

Quoting from the actual specification here : 
https://microsoft.github.io/language-server-protocol/specification#workspace_workspaceFolders

> Many tools support more than one root folder per workspace. Examples for this are VS Code’s multi-root support, Atom’s project folder support or Sublime’s project support. If a client workspace consists of multiple roots then a server typically needs to know about this. The protocol up to now assumes one root folder which is announced to the server by the rootUri property of the InitializeParams. If the client supports workspace folders and announces them via the corresponding workspaceFolders client capability, the InitializeParams contain an additional property workspaceFolders with the configured workspace folders when the server starts.

In Brackets we only have one project root, so workspace and project are synonymous. This way we can piggyback on the client capability "workspaceFolders", to change project root. I've done so in the default capabilities. A change in the project root can thus be translated as an addition of a folder and removal of a folder from a workspace which only has one folder.

So quoting again:

> The workspace/didChangeWorkspaceFolders notification is sent from the client to the server to inform the server about workspace folder configuration changes. 

I believe the reason we have both "rootUri" and "workspaceFolders" in the specification is because of backward compatibility. Earlier the servers didn't support multiple workspace folders, and separate clients had to be created per workspace folder manually in the extension.

src/extensions/default/PhpTooling/CodeHintsProvider.js
```diff
@@ -0,0 +1,205 @@
...
+
+    CodeHintsProvider.prototype.hasHints = function (editor, implicitChar) {
+        if (!this.client) {
+            return false;
+        }
+
+        return true;
+    };
+
+    CodeHintsProvider.prototype.getHints = function (implicitChar) {
```
shubhsnov: I don't think we need this file. You can create override the getHInts function of the DefaultProvider object since that is the only place is where you need to change anything. See my comments below.

src/extensions/default/PhpTooling/main.js
```diff
@@ -0,0 +1,244 @@
...
+                        rootPath: directory.fullPath
+                    });
+                }, 1500);
+            }).fail(function () {
+                console.log("Error encountered while stoping Php Server.");
+            });
+    };
+
+    function registerToolingProviders() {
+        var chProvider = new CodeHintsProvider(_client),
```
shubhsnov: do this instead:
```
var chProvider = new DefaultProviders.CodeHintsProvider(_client),
chProvider.getHints = yourOwnGetHintsFunction;
```

niteskum:
> @niteskum
> 
> > @shubhsnov Regarding FolderChange Event my understanding is "workspace/didChangeWorkspaceFolders" event is for workspace folder structure change like (folder added or removed from workspace ), not for the workspace project root folder change.
> > correct me if my understanding is wrong.
> 
> Quoting from the actual specification here :
> https://microsoft.github.io/language-server-protocol/specification#workspace_workspaceFolders
> 
> > Many tools support more than one root folder per workspace. Examples for this are VS Code’s multi-root support, Atom’s project folder support or Sublime’s project support. If a client workspace consists of multiple roots then a server typically needs to know about this. The protocol up to now assumes one root folder which is announced to the server by the rootUri property of the InitializeParams. If the client supports workspace folders and announces them via the corresponding workspaceFolders client capability, the InitializeParams contain an additional property workspaceFolders with the configured workspace folders when the server starts.
> 
> In Brackets we only have one project root, so workspace and project are synonymous. This way we can piggyback on the client capability "workspaceFolders", to change project root. I've done so in the default capabilities. A change in the project root can thus be translated as an addition of a folder and removal of a folder from a workspace which only has one folder.
> 
> So quoting again:
> 
> > The workspace/didChangeWorkspaceFolders notification is sent from the client to the server to inform the server about workspace folder configuration changes.
> 
> I believe the reason we have both "rootUri" and "workspaceFolders" in the specification is because of backward compatibility. Earlier the servers didn't support multiple workspace folders, and separate clients had to be created per workspace folder manually in the extension.

Php server does not support WorkspaceFolders change Notification. so this can not be tested with Php client. so I am not using this in PHP client. we should not ship any feature without having tested.

instead I have used servercapabilities.definitionProvider in PHP client  this flag is true if Language Server support "go to definition". PHP Server support so i have used this PHP client.

shubhsnov:
@niteskum Don't worry, I have added tests for the same already, by creating a sample LSP server, which supports the event. What you need to do here is just check the existence of the server capability. If it's available then use the "didChangeWorkspaceFolder" event, otherwise do an explicit restart.

src/extensions/default/PhpTooling/CodeHintsProvider.js
```diff
@@ -0,0 +1,205 @@
...
+            var context = TokenUtils.getInitialContext(editor._codeMirror, pos),
+                hints = [];
+
+            self.query = context.token.string.slice(0, context.pos.ch - context.token.start);
+            if (msgObj) {
+                var res = msgObj.items || [];
+                // There is a bug in Php Language Server, Php Language Server does not provide superGlobals
+                // Variables as completion. so these variables are being explicity put in response objects
+                // below code should be removed if php server fix this bug.
+                for(var key in phpSuperGlobalVariables) {
```
shubhsnov: I don't think we should add the static hints here always, otherwise we'll show hints even when they are not required. It should only come when implicitChar is '$', or otherwise filtered by the current token's prefix in case the token starts with '$'.

niteskum: it will shown only if prefix is '$'. in below code "filterWithQueryAndMatcher" will filter it.

niteskum:
> > Php server does not support WorkspaceFolders change Notification. so this can not be tested with Php client. so I am not using this in PHP client. we should not ship any feature without having tested.
> > instead I have used servercapabilities.definitionProvider in PHP client this flag is true if Language Server support "go to definition". PHP Server support so i have used this PHP client.
> 
> @niteskum Don't worry, I have added tests for the same already, by creating a sample LSP server, which supports the event. What you need to do here is just check the existence of the server capability`(ServerCapabilities/workspace/workspaceFolders)`. If it's available then use the "didChangeWorkspaceFolder" event, otherwise do an explicit restart.

If we are going to use this event , then in initialize function we should also send workspaceFolders which is root_path, this require a change in core side in initialize function.

src/extensions/default/PhpTooling/main.js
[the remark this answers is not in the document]
niteskum: fixed

src/extensions/default/PhpTooling/main.js
niteskum: I don't see this gives us any benefits, I understand that I have changes only in getHints function but still I have to rewrite all Util functions which are defined in default code Hints provider and I have to access "this" object, If i use "this" variable in normal functions that will look weird. this seems to be an hack code not recommended coding practise. 
If you are going to externalise all util functions, then I can reuse your util functions from default provider.

src/extensions/default/PhpTooling/CodeHintsProvider.js
niteskum: see my comment below

niteskum:
> Another thing to note: To test the validation logic against PHP5 and PHP7. Both should give different results, based on syntax.

For LSP server to work we need PHP 7, with PHP 5 LSP server will not work

src/extensions/default/PhpTooling/main.js
```diff
@@ -0,0 +1,204 @@
...
+        description: Strings.DESCRIPTION_PHP_TOOLING_CONFIGURATION
+    });
+
+    PreferencesManager.on("change", "php", function () {
+        var newPhpConfig = PreferencesManager.get("php");
+
+        if ((newPhpConfig["executablePath"] !== phpConfig["executablePath"])
+                || (newPhpConfig["enablePhpTooling"] !== phpConfig["enablePhpTooling"])) {
+            phpConfig = newPhpConfig;
+            startPhpServer();
```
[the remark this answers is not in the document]
niteskum: fixed

niteskum:
> > > Php server does not support WorkspaceFolders change Notification. so this can not be tested with Php client. so I am not using this in PHP client. we should not ship any feature without having tested.
> > > instead I have used servercapabilities.definitionProvider in PHP client this flag is true if Language Server support "go to definition". PHP Server support so i have used this PHP client.
> > 
> > 
> > @niteskum Don't worry, I have added tests for the same already, by creating a sample LSP server, which supports the event. What you need to do here is just check the existence of the server capability`(ServerCapabilities/workspace/workspaceFolders)`. If it's available then use the "didChangeWorkspaceFolder" event, otherwise do an explicit restart.
> 
> If we are going to use this event , then in initialize function we should also send workspaceFolders which is root_path, this require a change in core side in initialize function.

fixed

src/extensions/default/PhpTooling/phpGlobals.json
```diff
@@ -0,0 +1,62 @@
...
+	},
+	"$argc": {
+		"description": "Contains the number of arguments passed to the current script when running from the command line.",
+		"type": "array"
+	},
+	"$argv": {
+		"description": "Contains an array of all the arguments passed to the script when running from the command line.",
+		"type": "array"
+	},
+	"$this": {
```
shubhsnov: Add the "parent" and "self" keyword as well.

src/extensions/default/PhpTooling/main.js
```diff
@@ -0,0 +1,244 @@
...
+                    });
+                }, 1500);
+            }).fail(function () {
+                console.log("Error encountered while stoping Php Server.");
+            });
+    };
+
+    function registerToolingProviders() {
+        var chProvider = new CodeHintsProvider(_client),
+            phProvider = new DefaultProviders.ParameterHintsProvider(_client),
```
niteskum: We need servercapabilities in client

src/extensions/default/PhpTooling/CodeHintsProvider.js
```diff
@@ -0,0 +1,160 @@
...
+                }
+            }
+
+            return searchResult;
+        });
+
+        return matchResults;
+    }
+
+    CodeHintsProvider.prototype.hasHints = function (editor, implicitChar) {
```
shubhsnov: I'll add the ability to add static hints in the Default Provider itself.

src/extensions/default/PhpTooling/main.js
shubhsnov: Sure.

src/extensions/default/PhpTooling/main.js
```diff
@@ -0,0 +1,252 @@
...
+                    }, 1500);
+                }).fail(function () {
+                    console.log("Error encountered while stoping Php Server.");
+                });
+        }
+    };
+
+    function registerToolingProviders() {
+        var chProvider = new CodeHintsProvider(_client),
+            phProvider = new DefaultProviders.ParameterHintsProvider(_client),
```
shubhsnov: @niteskum Please override the hasParameterHints function to handle <no parameter> bug. Use `TokenUtils` to check if the previous token is of type "variable", and return true only then.

src/extensions/default/PhpTooling/main.js
```diff
@@ -0,0 +1,252 @@
...
+        }
+
+        currentRootPath = ProjectManager.getProjectRoot()._path;
+        setTimeout(function () {
+            CodeInspection.requestRun("Diagnostics");
+        }, 1500);
+    }
+
+    function restart(project) {
+        var result = $.Deferred();
```
shubhsnov: Try using _client.restart directly.

src/extensions/default/PhpTooling/phpGlobals.json
shubhsnov: Also __destruct.

src/extensions/default/PhpTooling/main.js
niteskum: @shubhsnov  what is the bug?

shubhsnov:
@niteskum There is a bug with the php server, where it provides the parameter hints just outside the function after the closing parentheses  ')'. Can you take a look at this and see if we can do a targeted fix from Brackets side.

shubhsnov:
LGTM 👍

swmitra:
Please resolve the conflicts @niteskum.

niteskum:
> Please resolve the conflicts @niteskum.

resolved conflicts in strings.js

shubhsnov:
@gavankar I think the Travis CI env configuration should also be updated, for PHP 7 and composer 1.8.4. The CI is failing due to this and will continue to fail otherwise.

shubhsnov (approved):

gavankar:
Submitted a fix to

PR #14683, opened 2019-04-03
Sending Analytics Data for file Operation and LSP features

niteskum:
@shubhsnov @SnchitGrover please review
@narayani28

### DIFF
--- a/src/editor/EditorStatusBar.js
+++ b/src/editor/EditorStatusBar.js
@@ -465,6 +465,17 @@ define(function (require, exports, module) {
             var document = EditorManager.getActiveEditor().document,
                 fullPath = document.file.fullPath;
 
+            var fileType = (document.file instanceof InMemoryFile) ? "newFile" : "existingFile",
+                filelanguageName = lang ? lang._name : "";
+
+            HealthLogger.sendAnalyticsData(
+                "usagelanguageChange" + filelanguageName + fileType,
+                "usage",
+                "languageChange",
+                filelanguageName.toLowerCase(),
+                fileType
+            );
+
             if (lang === LANGUAGE_SET_AS_DEFAULT) {
                 // Set file's current language in preferences as a file extension override (only enabled if not default already)
                 var fileExtensionMap = PreferencesManager.get("language.fileExtensions");
--- a/src/editor/EditorStatusBar.js
+++ b/src/editor/EditorStatusBar.js
@@ -469,9 +469,10 @@ define(function (require, exports, module) {
                 filelanguageName = lang ? lang._name : "";
 
             HealthLogger.sendAnalyticsData(
-                "usagelanguageChange" + filelanguageName + fileType,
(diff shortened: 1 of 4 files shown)


src/editor/EditorStatusBar.js
```diff
@@ -465,6 +465,17 @@ define(function (require, exports, module) {
             var document = EditorManager.getActiveEditor().document,
                 fullPath = document.file.fullPath;
 
+            var fileType = (document.file instanceof InMemoryFile) ? "newFile" : "existingFile",
+                filelanguageName = lang ? lang._name : "";
+
+            HealthLogger.sendAnalyticsData(
+                "usagelanguageChange" + filelanguageName + fileType,
+                "usage",
```
SnchitGrover: @niteskum can all analytics strings be moved to another file as variables and use that variable everywhere needed?

SnchitGrover (changes requested):

src/utils/HealthLogger.js
```diff
@@ -165,6 +166,92 @@ define(function (require, exports, module) {
...
+     * Whenever a file is saved call this function.
+     * The function will send the analytics Data
+     * We only log the standard filetypes and fileSize
+     * @param {String} filePath The path of the file to be registered
+     */
+    function fileSaved(docToSave) {
+        if (!docToSave) {
+            return;
+        }
+        var fileType = docToSave.language ? docToSave.language._name : "";
```
shubhsnov: @SnchitGrover Do we want an empty string here or something like "unknown"? Empty strings are sometimes treated as null values while serialization depending on platform. Depends on your use case.

shubhsnov: @niteskum Is there ever a possibility of not having the `language` property in a document object?

niteskum: @shubhsnov  I think document will always have language. but since file save is critical operation , don't want to leave any space in code to trigger exception so I am doing  null check.

src/editor/EditorStatusBar.js
niteskum: moved all common strings constant to HealthLogger.js as earlier all constants were defined in this file

src/utils/HealthLogger.js
```diff
@@ -165,6 +174,92 @@ define(function (require, exports, module) {
...
+
+        function _sendData(fileSize) {
+            var subType = "";
+
+            if(fileSize/1024 < 1) {
+
+                if(fileSize === -1) {
+                    subType = "";
+                }
+                if(fileSize < 10) {
```
narayani28: shouldn't the check be <=? if the file size is 10kb, it will log in 10_50KB range.

niteskum: fixed

SnchitGrover:
LGTM

shubhsnov (approved):

PR #14684, opened 2019-04-03
Removing the old Brackets CLA check

shubhsnov:
Brackets has integrated Adobe CLA check by default now, so we don't need the old check anymore. This task can be updated if we ever have to create a custom Adobe CLA bot integration.
ping @swmitra for review


swmitra (approved):
LGTM 🥇

PR #14685, opened 2019-04-03
Fixed LSP Linting Synchronization Issues

niteskum:
@shubhsnov  please review.
@narayani28

### DIFF
--- a/src/extensions/default/PhpTooling/main.js
+++ b/src/extensions/default/PhpTooling/main.js
@@ -107,7 +107,7 @@ define(function (require, exports, module) {
         CodeHintManager.registerHintProvider(chProvider, ["php"], 0);
         ParameterHintManager.registerHintProvider(phProvider, ["php"], 0);
         CodeInspection.register(["php"], {
-            name: Strings.PHP_DIAGNOSTICS,
+            name: "",
             scanFileAsync: lProvider.getInspectionResultsAsync.bind(lProvider)
         });
 
@@ -123,7 +123,6 @@ define(function (require, exports, module) {
 
         if (phpConfig["validateOnType"] !== "false") {
             lProvider._validateOnType = true;
-            lProvider._linterName = Strings.PHP_DIAGNOSTICS;
         }
 
         _client.addOnProjectOpenHandler(handleProjectOpen);
--- a/src/extensions/default/PhpTooling/main.js
+++ b/src/extensions/default/PhpTooling/main.js
@@ -57,7 +57,11 @@ define(function (require, exports, module) {
         DEBUG_OPEN_PREFERENCES_IN_SPLIT_VIEW  = "debug.openPrefsInSplitView",
         phpServerRunning = false,
         serverCapabilities,
-        currentRootPath;
+        currentRootPath,
+        chProvider,
+        phProvider,
+        lProvider,
+        jdProvider;
 
     PreferencesManager.definePreference("php", "object", phpConfig, {
         description: Strings.DESCRIPTION_PHP_TOOLING_CONFIGURATION
@@ -66,6 +70,9 @@ define(function (require, exports, module) {
     PreferencesManager.on("change", "php", function () {
         var newPhpConfig = PreferencesManager.get("php");
 
+        if (lProvider && newPhpConfig["validateOnType"] !== phpConfig["validateOnType"]) {
+            lProvider._ValidateOnType = !(newPhpConfig["validateOnType"] === "false");
+        }
         if ((newPhpConfig["executablePath"] !== phpConfig["executablePath"])
                 || (newPhpConfig["enablePhpTooling"] !== phpConfig["enablePhpTooling"])) {
             phpConfig = newPhpConfig;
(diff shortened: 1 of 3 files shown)


src/languageTools/DefaultProviders.js
```diff
@@ -335,16 +335,21 @@ define(function (require, exports, module) {
 
     function LintingProvider() {
         this._results = new Map();
+        this._defferedLintingPromise = new Map();
```
shubhsnov: Nit: the member variables have issues:

- deffered => deferred (typo)
- change defferedLintingPromise to something which seems more intuitive for a map (suggestion: use 'promises' or 'promiseMap' in the name)
- V => v (lowercase, only Modules and Classed are capitalized by convention)
- lintingName => could be changed to 'providerName' or 'linterName'

src/languageTools/DefaultProviders.js
```diff
@@ -371,6 +376,23 @@ define(function (require, exports, module) {
         this._results.set(filePath, {
             errors: errors
         });
+        if(this._defferedLintingPromise.get(filePath)) {
+           this._defferedLintingPromise.get(filePath).resolve(this._results.get(filePath));
+           this._defferedLintingPromise.delete(filePath);
+        }
+        if (this._ValidateOnType) {
```
shubhsnov: Few issues: 

- Check if filePath === currentActiveDocumentPath? No need to run inspection for files which are not active.
- How is onSave being handled here?
- Why do we need a promiseMap here at all? The request is being run once we have already gotten the result.

niteskum: @shubhsnov  on File Save and File Open diagnostics would be run by CodeInspection Manager itself.
in case of file open Diagnostics result won't be available immediately in that case result will be return using promise

shubhsnov: Okay. If the promise is only for the file open scenario then why do we require a Map? We can just keep a promise for the current doc since we are already not running the code inspection request for any inactive docs.

niteskum: FileOpen scenario is the the sceanrio which I could think where result might not be available. there could be other scenarios as well.

shubhsnov: In any case, we only show the linting for the current file. Only having currentActiveDocPromise should suffice. No point in caching since the cached result will never be used.

shubhsnov: Also please add a check in the getInspectionResultsAsync for filePath === currentActiveDocumentPath as well.

src/extensions/default/PhpTooling/main.js
```diff
@@ -107,7 +107,7 @@ define(function (require, exports, module) {
         CodeHintManager.registerHintProvider(chProvider, ["php"], 0);
         ParameterHintManager.registerHintProvider(phProvider, ["php"], 0);
         CodeInspection.register(["php"], {
-            name: Strings.PHP_DIAGNOSTICS,
```
shubhsnov: Have something here to differentiate from any other provider which might also provide "php" related linting.

src/languageTools/DefaultProviders.js
niteskum: > In any case, we only show the linting for the current file. Only having currentActiveDocPromise should suffice. No point in caching since the cached result will never be used.

using just variable will cause problem. If you will see the code workflow It will not cache result if it is not to be used.  To me it looks perfectly ok. we can discuss this if you like.

src/extensions/default/PhpTooling/main.js
niteskum: As discussed we are not having any name for default PHP Lint Provider

src/languageTools/DefaultProviders.js
niteskum: fixed

src/languageTools/DefaultProviders.js
niteskum: > Also please add a check in the getInspectionResultsAsync for filePath === currentActiveDocumentPath as well.

already fixed

shubhsnov: > > Also please add a check in the getInspectionResultsAsync for filePath === currentActiveDocumentPath as well.
> 
> already fixed

I specifically meant for `getInspectionResultsAsync` not `setInspectionResultsAsync`. It would act as a guard and optimization. Suppose a file has hundred of errors, but we switched to any other file before actually getting the results. Why would we want those hundred of results to pass through another function callback and then get discarded, since the active file changed. Doesn't seem proper to me. If we have a check in set, we should have a check in get.

niteskum: Not Sure if you are looking at complete code workflow from CodeInspection side. It looks perfectly ok to me. the guard you are mentioning is already in place in codeInspection. CodeInspection always call getInspectionResultsAsync  for current File only.
and this function can't return null, It has to return promise.

If still not clear we can discuss this.

shubhsnov: It's an optimization.  Okay let me explain it to you more clearly here:
You open File A (Some large file which let's assume will cause some time for the diagnostics response) 
getInspectionResultsAsync will get called
Just as the function was called you switched the document
Now you are inside the getInspectionResultsAsync for File A but the active Document is File B

Now according to me, instead of creating a promise and waiting for the promise to resolve, we can instead do $.Deferred().resolve(null) because it won't matter, as CodeInspection will drop the promise as it is.

niteskum: not taking this change as I am not convinced that it is an optimization. we are just double checking if the file is current file.

shubhsnov:
LGTM 👍

PR #14687, opened 2019-04-05
switching from anonymous to pseudonymous to align with the definition…

niteskum:
switching from "anonymous" to "pseudonymous" to align with the definitions used in the GDPR.

@swmitra @narayani28 please review

### DIFF
--- a/src/nls/root/strings.js
+++ b/src/nls/root/strings.js
@@ -650,10 +650,10 @@ define({
     // extensions/default/HealthData
     "HEALTH_DATA_NOTIFICATION"                  : "Health Report Preferences",
     "HEALTH_FIRST_POPUP_TITLE"                  : "{APP_NAME} Health Report",
-    "HEALTH_DATA_DO_TRACK"                      : "Share anonymous information on how I use {APP_NAME}",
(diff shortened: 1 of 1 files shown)


niteskum:
Legal: Replace all references to "anonymous" with "pseudonymous" (including, but not limited to, the Brackets UI itself, Brackets Health Report, and the Health Data webpage - see attached screenshots)

Changes recommended by Jane Lee (PFA mailthread)
•	Proceed with proposed changes to language for next release (in July). As previously discussed, all references to "anonymous" should be changed to "pseudonymous" (including, but not limited to, the Brackets UI itself, Brackets Health Report, and the Health Data webpage - see attached screenshots)
•	Rename "Health Data" to "Project Health Data" on https://github.com/adobe/brackets/wiki/Health-Data 
•	Make it clear on https://github.com/adobe/brackets/wiki/Health-Data that Adobe is collecting the data on the Health Report (this is clear in the Brackets Health Report pop-up in the UI, but not on the webpage)
•	As for the proactive statement about the change from "anonymous" to "pseudonymous," we can simply say that we are switching from "anonymous" to "pseudonymous" to align with the definitions used in the GDPR. We also recommend adding a note that no change is being made to the type of data collected (assuming that is true for now; if/when we change the type of data we collect, we should revisit this topic) so as to limit potential concerns.

sobisht (commented):
LGTM 👍

narayani28:
how about the other changes to be done as recommended by Jane Lee?

niteskum:
> how about the other changes to be done as recommended by Jane Lee?

we have to update wiki page , I don't have access to edit wiki page.
@sobisht can you help me with this?

narayani28:
LGTM

PR #14691, opened 2019-04-09
Fix for "GetItNow" action behaviour

swmitra:
When user clicks on "GetItNow" it should perform one of these actions 

- Either initialises Auto-Update process after all the conditions verified. 
- Or if Auto-Update can't be initialised, then take the user to brackets.io.

However, we shouldn't perform both the actions. 

This PR fixes the faulty behaviour and ensures, we don't open brackets.io if we are able to start auto-update successfully. 

@shubhsnov @narayani28 Please review.

### DIFF
--- a/src/extensions/default/AutoUpdate/main.js
+++ b/src/extensions/default/AutoUpdate/main.js
@@ -483,6 +483,9 @@ define(function (require, exports, module) {
 
             //Initiate the auto update, with update params
             initiateAutoUpdate(updateParams);
+	    
+	    //Send a truthy value to ensure caller is informed about successful initialization of auto-update
+	    return true;
     }
 
 
--- a/src/utils/UpdateNotification.js
+++ b/src/utils/UpdateNotification.js
@@ -522,8 +522,8 @@ define(function (require, exports, module) {
      */
     function handleUpdateProcess(updates) {
         var handler = _updateProcessHandler || _defaultUpdateProcessHandler;
-        var success = handler(updates);
-        if (_updateProcessHandler && !success) {
+        var initSuccess = handler(updates);
+        if (_updateProcessHandler && !initSuccess) {
             // Give a chance to default handler in case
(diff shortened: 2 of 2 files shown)


src/extensions/default/AutoUpdate/main.js
```diff
@@ -483,6 +483,9 @@ define(function (require, exports, module) {
 
             //Initiate the auto update, with update params
             initiateAutoUpdate(updateParams);
+	    
+	    //Send a truthy value to ensure caller is informed about successful initialization of auto-update
+	    return true;
```
shubhsnov: This seems okay to me. 

Just one nit, in UpdateNotification.js the _updateProcessHandler is actually an async function(internally calls initiateAutoUpdate), but  _defaultUpdateProcessHandler would work synchronously. The way the code is written seems to imply that the update was successfully handled, which is not the case even if the initialization was successful. Should we change the variable name to `initSuccess` in UpdateNotification.js?
On the otherhand, we can return a promise from here and resolve it with true or false accordingly to indicate actual success, but that would make things a bit complicated.

swmitra: Understood. Will change the variable name in that case. It can't be async, or rather we can't wait till success of auto-update process itself, because that would break the continuity of user journey in the workflow. No point in launching brackets.io after 5 minutes if download has failed.

shubhsnov: Oh yes, definitely. What I meant was to handle the initialization failure scenario separately from update success/failure scenario, if required but through a promise since that will be more reliable. As for update failure, we're already telling the user to go to the site if the update fails [here](https://github.com/adobe/brackets/blob/master/src/extensions/default/AutoUpdate/main.js#L247).

swmitra: Changed `success` to `initSuccess` now.

shubhsnov (approved):

PR #14692, opened 2019-04-10
Fix for PHPCodeHints invocation failure scenarios

shubhsnov:
Fix for class-based invocation "::"
Fix for object-based invocation "->"
Fix for Explicit invocation "Ctrl-Space"
Fix for insertHints in the above scenarios

ping @niteskum @narayani28 @swmitra  for review.

### DIFF
--- a/src/extensions/default/PhpTooling/CodeHintsProvider.js
+++ b/src/extensions/default/PhpTooling/CodeHintsProvider.js
@@ -101,11 +101,15 @@ define(function (require, exports, module) {
 
             self.query = context.token.string.slice(0, context.pos.ch - context.token.start);
             if (msgObj) {
-                var res = msgObj.items || [];
+                var res = msgObj.items || [],
+                    trimmedQuery = self.query.trim(),
+                    hasIgnoreCharacters = self.ignoreQuery.includes(implicitChar) || self.ignoreQuery.includes(trimmedQuery),
+                    isExplicitInvokation = implicitChar === null;
+
                 // There is a bug in Php Language Server, Php Language Server does not provide superGlobals
                 // Variables as completion. so these variables are being explicity put in response objects
                 // below code should be removed if php server fix this bug.
-                if(self.query) {
+                if((isExplicitInvokation || trimmedQuery) && !hasIgnoreCharacters) {
                     for(var key in phpSuperGlobalVariables) {
                         res.push({
                             label: key,
@@ -115,7 +119,12 @@ define(function (require, exports, module) {
                     }
                 }
 
-                var filteredHints = filterWithQueryAndMatcher(res, self.query);
+                var filteredHints = [];
+                if (hasIgnoreCharacters || (isExplicitInvokation && !trimmedQuery)) {
+                    filteredHints = filterWithQueryAndMatcher(res, "");
+                } else {
+                    filteredHints = filterWithQueryAndMatcher(res, self.query);
+                }
 
                 StringMatch.basicMatchSort(filteredHints);
                 filteredHints.forEach(function (element) {
--- a/src/languageTools/DefaultProviders.js
+++ b/src/languageTools/DefaultProviders.js
@@ -46,6 +46,7 @@ define(function (require, exports, module) {
     function CodeHintsProvider(client) {
         this.client = client;
         this.query = "";
+        this.ignoreQuery = ["-", "->", ">", ":", "::", "(", "()", ")", "[", "[]", "]", "{", "{}", "}"];
     }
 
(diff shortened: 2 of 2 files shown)


src/extensions/default/PhpTooling/CodeHintsProvider.js
```diff
@@ -115,7 +119,12 @@ define(function (require, exports, module) {
                     }
                 }
 
-                var filteredHints = filterWithQueryAndMatcher(res, self.query);
+                var filteredHints = [];
+                if (hasIgnoreCharacters || (isExplicitInvokation && !trimmedQuery)) {
+                    filteredHints = filterWithQueryAndMatcher(res, "");
+                } else {
+                    filteredHints = filterWithQueryAndMatcher(res, self.query);
```
niteskum: do we need to filter hints. isn't server itself sending filtered hints based on the query??

niteskum (changes requested):

src/extensions/default/PhpTooling/CodeHintsProvider.js
shubhsnov: Filtering here isn't required for creating stringRanges, and matchGoodness, which are then used to highlight the matched text and sort the things accordingly as per the query.

src/languageTools/DefaultProviders.js
```diff
@@ -46,6 +46,7 @@ define(function (require, exports, module) {
     function CodeHintsProvider(client) {
         this.client = client;
         this.query = "";
+        this.ignoreQuery = ["-", "->", ">", ":", "::", "(", "()", ")", "[", "[]", "]", "{", "{}", "}"];
```
narayani28: Since you have added this list to the defaultprovider, the ignorequery list will be different for different languages. why should this be here?

shubhsnov: I could have made it specific to PHP but the CodeHintProvider for PHP was using the insertHint from the DefaultProvider, and we needed to know if any query needed to be ignored there as well.

Secondly, this was something that would affect most languages. For PHP I could have stopped at "->" and ":" but then decided to take a more generic approach by including the braces as well, as these things need to be ignored by most languages.

src/extensions/default/PhpTooling/CodeHintsProvider.js
niteskum: do we have any known scenario where hints returned from server is not a good match with query?

niteskum: if the purpose is to create  stringRanges, and matchGoodness, then why this function removes some of hints from hintlist retured from server??

niteskum: Can we make change here that none of the hints returned from server should be removed?
If we can do this then we will not have to worry about ignoreCharacters list

shubhsnov: I am not sure I understand. The characters in ignoreList are from the query, not the hints. In essence, I am actually doing the same thing you have suggested. The query is being saved on the Brackets side for a session. So when a user types "se" the results are then filtered by calculating the distance of the hint from the subsequence "se". The results are then sorted by the distance. When a user types something like "->" which is not part of the hint but still invokes hints, I am just not using "->" to filter out the hints, rather preserving the order of the hints as they come.

niteskum: @shubhsnov 
I feel same thing  can be done without maintaining an ignore characters list.

make change in filterWithQueryAndMatcher function.
Pass a boolean parameter(dontRemoveIfNoMatch) to this function. if this third parameter is true none of the hints should be removed even if hints has no match with query.
call this "filterWithQueryAndMatcher" for server hints with dontRemoveIfNoMatch as true.
and call this function separately for static hints with dontRemoveIfNoMatch  as false which is default.

lets think if we can do this?

swmitra: @niteskum `filterWithQueryAndMatcher` implementation shouldn't be changed as there are multiple extensions using this fn. This is more generic implementation which shouldn't be altered for code hint listing use cases.

src/languageTools/DefaultProviders.js
swmitra: IMO, `-` inclusion in ignore list would create problems.  `-` is a valid token character for a lot of languages. HTML5 attributes, CSS property names and values, where we need filtering.

swmitra (approved):
Barring some minor comments on implementation, rest of the code looks ok to me. I am marking this approved as those comments can be revisited in future. LGTM :100:

PR #14698, opened 2019-04-12
Fix : Error occurs when dismissing QuickOpen dialog when no document is open

shubhsnov:
Add null checks for the same.
ping @niteskum @narayani28 for review.


niteskum:
LGTM 👍

PR #14705, opened 2019-04-16
Fixed Find References Issue in release build

niteskum:
This PR fixes two issues:

- Makes the Find References menu contextual (Active only when we have a provider for the file type)
- Correct the order in which DOM is created for FindReference Bottom panel (After status bar has been inserted, since a bottom panel is prepended to it)

@shubhsnov @narayani28 @swmitra  Please review

### DIFF
--- a/src/brackets.js
+++ b/src/brackets.js
@@ -141,7 +141,6 @@ define(function (require, exports, module) {
     //load language features
     require("features/ParameterHintsManager");
     require("features/JumpToDefManager");
-    require("features/FindReferencesManager");
 
     // Load modules that self-register and just need to get included in the main project
     require("command/DefaultMenus");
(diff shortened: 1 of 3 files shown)


narayani28 (commented):
Can you describe how the code changes fixes the issue and what's the original issue was?

niteskum:
Below is the explanation why Find References was not working with Release build:

in htmlReady function of FindReferencesManager we create SearchResultsView Object, which internally call a function  createBottomPanel. in this function panel is inserted before div element "#status-bar"
 $panel.insertBefore("#status-bar");

so when htmlReady function was getting called "#status-bar" div element was not present and it used to fail to insert panel in DOM.  "#status-bar" div element is inserted in DOM by htmlReady function of "widgets/StatusBar". so ""widgets/StatusBar" must be required before FindReferences manager.

I have moved code "require("features/FindReferencesManager");" in brackets.js after "require("editor/EditorStatusBar");"

in DebugBuild it was working because in debug build files are loaded in order of file tree structure in Project root, not based on the order it is required in brackets.js

shubhsnov (approved):

PR #14706, opened 2019-04-16
Fixed UriToPath Utill Function

niteskum:
uriToPath function was converting network path and new file path incorrectly on windows.
@shubhsnov  please review

### DIFF
--- a/src/languageTools/PathConverters.js
+++ b/src/languageTools/PathConverters.js
@@ -36,7 +36,7 @@ define(function (require, exports, module) {
 
         let filePath = decodeURIComponent(url.pathname);
         if (brackets.platform === 'win') {
-            if (filePath && filePath.length > 1 && filePath[0] === '/' && filePath[1] !== '/') {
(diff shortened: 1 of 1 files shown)


src/languageTools/PathConverters.js
```diff
@@ -36,7 +36,7 @@ define(function (require, exports, module) {
 
         let filePath = decodeURIComponent(url.pathname);
         if (brackets.platform === 'win') {
-            if (filePath[0] === '/') {
+            if (filePath && filePath.includes(":/") && filePath[0] === '/') {
```
shubhsnov: Can you please describe the format of paths that are returned by the server in different scenarios?

That would help assess this change.

niteskum: @shubhsnov 
Below are the file format in different scenario
For local disk file: file:///C:/<file_path>
For Network FIle : file:////blrstore.corp.adobe.com/<file_path>
For Temp New File : file:///_brackets1234/<file_name>

shubhsnov (approved):

PR #14715, opened 2019-04-22
InApp Notifications

swmitra:
This PR adds the required infrastructure to display App notifications created by Brackets publisher with the ability to filter applicability of the notifications based on multiple parameters. A sample notifications file (localized) hosted from S3 would look like - 

```json
{
    "notifications": [
        {
            "sequence": "1555874975721",
            "silent": false,
            "filters": {
                "platforms": ["OSX"],
                "version": "1.14",
                "locales": ["en"],
                "builds": [],
                "extensions": [],
                "lastUsed": ""
            },
            "html": "<div><span>Web development and design trends are changing at a rapid pace. We want to hear how new paradigms and tools are impacting you and the way you work!</span><a id='banner-button' href='https://adobe.ly/bracketsiosurvey'>Take the Brackets Survey</a></div>",
            "actionables": "#banner-button",
            "expiry" : "1556044200000"
        }
    ]
}
```

### DIFF
--- a/src/brackets.config.prerelease.json
+++ b/src/brackets.config.prerelease.json
@@ -4,5 +4,6 @@
     "serviceKey"              : "brackets-service",
     "environment"             : "production",
     "update_info_url"         : "https://s3.amazonaws.com/files.brackets.io/updates/prerelease/<locale>.json",
+    "notification_info_url"         : "https://s3.amazonaws.com/files.brackets.io/notifications/prerelease/<locale>.json",
     "buildtype"               : "prerelease"
 }
--- a/src/brackets.config.prerelease.json
+++ b/src/brackets.config.prerelease.json
@@ -4,6 +4,6 @@
     "serviceKey"              : "brackets-service",
     "environment"             : "production",
     "update_info_url"         : "https://s3.amazonaws.com/files.brackets.io/updates/prerelease/<locale>.json",
-    "notification_info_url"         : "https://s3.amazonaws.com/files.brackets.io/notifications/prerelease/<locale>.json",
+    "notification_info_url"   : "https://s3.amazonaws.com/files.brackets.io/notifications/prerelease/<locale>.json",
     "buildtype"               : "prerelease"
 }
--- a/src/extensions/default/InAppNotifications/htmlContent/notificationContainer.html
+++ b/src/extensions/default/InAppNotifications/htmlContent/notificationContainer.html
@@ -1,7 +1,7 @@
 <div id="notification-bar" tabindex="0">
-    <div id="content-container">
+    <div class="content-container">
     </div>
-    <div id="close-icon-container">
-        <button type="button" id="close-icon" tabIndex="0">&times;</button>
+    <div class="close-icon-container">
+        <button type="button" class="close-icon" tabIndex="0">&times;</button>
     </div>
 </div>
--- a/src/extensions/default/InAppNotifications/htmlContent/notificationContainer.html
+++ b/src/extensions/default/InAppNotifications/htmlContent/notificationContainer.html
@@ -1,0 +1,7 @@
+<div id="notification-bar" {{#type}}class={{{type}}}{{/type}} tabindex="0">
+    <div id="content-container">
+    </div>
+    <div id="close-icon-container">
+        <button type="button" id="close-icon" tabIndex="0">&times;</button>
+    </div>
+</div>
--- a/src/extensions/default/InAppNotifications/htmlContent/notificationContainer.html
+++ b/src/extensions/default/InAppNotifications/htmlContent/notificationContainer.html
@@ -1,4 +1,4 @@
-<div id="notification-bar" {{#type}}class={{{type}}}{{/type}} tabindex="0">
+<div id="notification-bar" tabindex="0">
     <div id="content-container">
     </div>
     <div id="close-icon-container">
--- a/src/extensions/default/InAppNotifications/main.js
+++ b/src/extensions/default/InAppNotifications/main.js
@@ -88,8 +88,7 @@ define(function (require, exports, module) {
                 default:
                     locale = "en";
                 }
-                //return brackets.config.notification_info_url.replace("<locale>", locale);
-                return "file://./users/swamitra/Desktop/notifications/<locale>.json".replace("<locale>", locale);
+                return brackets.config.notification_info_url.replace("<locale>", locale);
             }
         }
         // PRERELEASE_END
@@ -120,9 +119,6 @@ define(function (require, exports, module) {
         if ((new Date()).getTime() > lastInfoURLFetchTime + ONE_DAY) {
             fetchData = true;
         }
-
-        // For dev testing, to be removed
-        fetchData = true;
 
         if (fetchData) {
             var lookupPromise = new $.Deferred(),
@@ -263,7 +259,9 @@ define(function (require, exports, module) {
     function _checkExtensions(filters) {
         var allExtensions = ExtensionManager.extensions,
             allExtnsMatched = true,
-            userExtensionKeys = Object.keys(allExtensions).filter(function(k) { return allExtensions[k].installInfo.locationType === 'user'; });
+            userExtensionKeys = Object.keys(allExtensions).filter(function(k) {
+                return allExtensions[k].installInfo.locationType === 'user';
+            });
 
         if (!filters.extensions) {
             allExtnsMatched = userExtensionKeys.size === 0;
@@ -325,6 +323,12 @@ define(function (require, exports, module) {
     function showNotification(msgObj) {
         var $htmlContent = $(msgObj.html),
             $notificationBarElement = $(NotificationBarHtml);
+
+        // Remove any SCRIPT tag to avoid secuirity issues
+        $notificationBarElement.find('script').remove();
+
+        // Remove any STYLE tag to avoid styling impact on Brackets DOM
+        $notificationBarElement.find('style').remove();
 
         cleanNotificationBar(); //Remove an already existing notification bar, if any
         $notificationBarElement.prependTo(".content");
--- a/src/extensions/default/InAppNotifications/main.js
+++ b/src/extensions/default/InAppNotifications/main.js
@@ -1,0 +1,362 @@
+/*
+ * Copyright (c) 2019 - present Adobe Systems Incorporated. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ *
+ */
+
+/**
+ *  module for displaying in-app notifications
+ *
+ */
+define(function (require, exports, module) {
+    "use strict";
+
+    var AppInit              = brackets.getModule("utils/AppInit"),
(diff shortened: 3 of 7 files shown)


src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
...
+     * Displays the Notification Bar UI
+     * @param   {object} msgObj - json object containing message info to be displayed
+     *
+     */
+    function showNotification(msgObj) {
+        var $htmlContent = $(msgObj.html),
+            $notificationBarElement = $(NotificationBarHtml);
+
+        // Remove any SCRIPT tag to avoid secuirity issues
+        $notificationBarElement.find('script').remove();
```
boopeshmahendran: Shouldn't we be sanitising `$htmlContent` as this is the network response?

Also, Should we think about more stricter checks for security as there are a lot of other ways to inject scripts in html?  **For Ex:**  Something like `<a onclick="some malicious script">`. 

For reference, https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet

src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
...
+    /**
+     * Check for notifications, notification overlays are always displayed
+     *
+     * @return {$.Promise} jQuery Promise object that is resolved or rejected after the notification check is complete.
+     */
+    function checkForNotification() {
+        var result = new $.Deferred();
+        var versionInfoUrl;
+
+        _getNotificationInformation(versionInfoUrl)
```
boopeshmahendran: Why are we calling with `undefined`? Couldn't we just write `_getNotificationInformation()`?

src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
...
+                    });
+                }
+                result.resolve();
+            })
+            .fail(function () {
+                // Error fetching the update data. If this is a forced check, alert the user
+                result.reject();
+            });
+
+        return result.promise();
```
boopeshmahendran: nit: I think we can directly return `_getNotificationInformation()` promise instead of creating a newer promise.

src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
...
+     */
+    function showNotification(msgObj) {
+        var $htmlContent = $(msgObj.html),
+            $notificationBarElement = $(NotificationBarHtml);
+
+        // Remove any SCRIPT tag to avoid secuirity issues
+        $notificationBarElement.find('script').remove();
+
+        // Remove any STYLE tag to avoid styling impact on Brackets DOM
+        $notificationBarElement.find('style').remove();
```
boopeshmahendran: Shouldn't we be sanitising `$htmlContent` as this is the network response?

src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
...
+                        if (_checkNotificationValidity(notificationObj)) {
+                            if (notificationObj.silent) {
+                                // silent notifications, to gather user validity based on filters
+                                HealthLogger.sendAnalyticsData("notification", notificationObj.sequence, "handled");
+                            } else {
+                                showNotification(notificationObj);
+                            }
+                            // Break, we have acted on one notification already
+                            return false;
+                        }
```
boopeshmahendran: Probably we will need a `return true` here since without it, it will return `undefined` thereby making the `every` to stop iteration. I think `notifications.find` will be more suited to this usecase.

src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
...
+    function _getVersionInfoUrl(locale, removeCountryPartOfLocale) {
+
+        locale = locale || brackets.getLocale();
+
+        if (removeCountryPartOfLocale) {
+            locale = locale.substring(0, 2);
+        }
+
+        // PRERELEASE_BEGIN
+        // The following code is needed for supporting notifications in prerelease,
```
boopeshmahendran: Can you please help me understand how does this work only for prerelease?

boopeshmahendran (changes requested):

src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
...
+        {
+            if (locale) {
+                if(locale.length > 2) {
+                    locale = locale.substring(0, 2);
+                }
+                switch(locale)  {
+                case "de":
```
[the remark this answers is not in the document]
shubhsnov: This will also work just fine here.
```
locales = ["fr", "de"...];
if (!locales.includes(locale)) {
    locale = "en";
}
```

src/extensions/default/InAppNotifications/main.js
shubhsnov: Nice catch. This is important.

src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
...
+                $.ajax({
+                    url: localVersionInfoUrl,
+                    cache: false,
+                    type: "HEAD"
+                }).fail(function (jqXHR, status, error) {
+                    // get rid of any country information from locale and try again
+                    var tmpUrl = _getVersionInfoUrl(brackets.getLocale(), true);
+                    if (tmpUrl !== localVersionInfoUrl) {
+                        $.ajax({
+                            url: tmpUrl,
```
boopeshmahendran: Is there some way to reduce the nesting of promises?

boopeshmahendran (changes requested):

src/extensions/default/InAppNotifications/main.js
swmitra: As the pre formatted html is provided by us, I don't wanted overwhelming checks and filter here. The checks included here are just to protect against accidental mistake and not from an actual attack.

src/extensions/default/InAppNotifications/main.js
swmitra: Explanation mentioned before.

src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
...
+    var AppInit              = brackets.getModule("utils/AppInit"),
+        Dialogs              = brackets.getModule("widgets/Dialogs"),
+        DefaultDialogs       = brackets.getModule("widgets/DefaultDialogs"),
+        PreferencesManager   = brackets.getModule("preferences/PreferencesManager"),
+        ExtensionUtils       = brackets.getModule("utils/ExtensionUtils"),
+        ExtensionManager     = brackets.getModule("extensibility/ExtensionManager"),
+        Strings              = brackets.getModule("strings"),
+        Mustache             = brackets.getModule("thirdparty/mustache/mustache"),
+        HealthLogger         = brackets.getModule("utils/HealthLogger"),
+        NotificationBarHtml  = require("text!htmlContent/notificationContainer.html");
```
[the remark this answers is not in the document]
swmitra: I would be using them as I have to add logic to display the notification as a dialog in next commit based on `display` configuration.

src/extensions/default/InAppNotifications/main.js
swmitra: Didn't have time to refactor the code, copied from Update notification module.

src/extensions/default/InAppNotifications/main.js
swmitra: Need to add a parameter in the wrapper function which can be provided for testing.

src/extensions/default/InAppNotifications/main.js
swmitra: Need to add a parameter in the wrapper function which can be provided for testing.

src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
+/*
+ * Copyright (c) 2019 - present Adobe Systems Incorporated. All rights reserved.
```
shubhsnov: Minor but necessary: You'll need to change this to `Copyright (c) 2019 - present Adobe. All rights reserved.` due to the recent name change for Adobe in all new documents.

src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
...
+        locale = locale || brackets.getLocale();
+
+        if (removeCountryPartOfLocale) {
+            locale = locale.substring(0, 2);
+        }
+
+        // PRERELEASE_BEGIN
+        // The following code is needed for supporting notifications in prerelease,
+        {
+            if (locale) {
```
shubhsnov: This code will run regardless of whether it's a prerelease or normal release. The comment above seems inaccurate.
Also, due to this we will never hit the return statement on line 96.

shubhsnov (commented):
Apart from a few issues that I and Boopesh have mentioned this looks functionally solid to me.

src/extensions/default/InAppNotifications/main.js
boopeshmahendran: Actually, I meant we should be sanitising `$htmlContent` variable instead of the `$notificationBarElement`.

src/extensions/default/InAppNotifications/main.js
swmitra: Done. no more "en-US" check first.

src/extensions/default/InAppNotifications/main.js
```diff
@@ -0,0 +1,372 @@
...
+        var data;
+
+        // If we don't have data saved in prefs, fetch
+        data = PreferencesManager.getViewState("notificationInfo");
+        if (!data) {
+            fetchData = true;
+        }
+
+        // If more than 24 hours have passed since our last fetch, fetch again
+        if ((new Date()).getTime() > lastInfoURLFetchTime + ONE_DAY) {
```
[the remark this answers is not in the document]
swmitra: Done.

src/extensions/default/InAppNotifications/main.js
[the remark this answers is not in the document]
swmitra: For now don't have time to bring in Dialogs, removed the imports.

src/extensions/default/InAppNotifications/main.js
swmitra: Removed the whole block.

src/extensions/default/InAppNotifications/main.js
swmitra: Handled.

src/extensions/default/InAppNotifications/main.js
swmitra: Done. Good catch!

src/extensions/default/InAppNotifications/main.js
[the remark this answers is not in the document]
swmitra: Removed.

src/extensions/default/InAppNotifications/htmlContent/notificationContainer.html
```diff
@@ -0,0 +1,7 @@
+<div id="notification-bar" tabindex="0">
+    <div id="content-container">
+    </div>
+    <div id="close-icon-container">
+        <button type="button" id="close-icon" tabIndex="0">&times;</button>
```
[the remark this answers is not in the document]
swmitra: Changed to `.` selector.

src/brackets.config.prerelease.json
```diff
@@ -4,5 +4,6 @@
     "serviceKey"              : "brackets-service",
     "environment"             : "production",
     "update_info_url"         : "https://s3.amazonaws.com/files.brackets.io/updates/prerelease/<locale>.json",
+    "notification_info_url"         : "https://s3.amazonaws.com/files.brackets.io/notifications/prerelease/<locale>.json",
```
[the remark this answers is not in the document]
swmitra: Fixed.

src/extensions/default/InAppNotifications/styles/styles.css
```diff
@@ -0,0 +1,56 @@
...
+    box-shadow: 0px 3px 6px rgba(0, 0, 0, 0.53);
+    padding: 5px 0px;
+    width: 100%;
+    min-height: 39px;
+    position: absolute;
+    z-index: 16;
+    left: 0px;
+    bottom: 25px;
+    outline: none;
+	overflow: hidden;
```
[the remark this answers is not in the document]
swmitra: Done.

swmitra:
@boopeshmahendran and @shubhsnov Thanks a ton for reviewing this PR minutely under short notice 💯 
Addressed all comments now. 
@shubhsnov Please test it once more whenever possible.

src/extensions/default/InAppNotifications/main.js
swmitra: Removed.

src/extensions/default/InAppNotifications/main.js
swmitra: Done.

src/extensions/default/InAppNotifications/styles/styles.css
```diff
@@ -0,0 +1,56 @@
...
+    width: 100%;
+}
+
+#notification-bar .close-icon-container {
+    height: auto;
+    position: absolute;
+    float: right;
+    text-align: center;
+    width: auto;
+	min-width: 66px;
```
boopeshmahendran: nit: some fomatting issues in this file.

boopeshmahendran (approved):

shubhsnov (approved):

shubhsnov:
LGTM 👍 Tested with multiple platforms, versions etc. Everything is working fine. With this, we can target notifications in the future. Thanks, @swmitra for this brilliant contribution.

PR #14719, opened 2019-04-23
Fixed Blank Error Dialog Issue

niteskum:
If there is any exception triggered in code it will not show blank dialog instead it will log the error in console.
@shubhsnov @narayani28  plzz review


shubhsnov (approved):

narayani28:
LGTM

PR #14722, opened 2019-04-24
Fixed Code/Parameter Hints Issue for large file

niteskum:
on editorChange event we have multiple listeners, each does one specific Task. 
regarding LSP on editor change , textSynchronization(didChange) notification is sent to lsp server and simultaneously requestHints/ParameterHints request are sent to server. 
For large file we observed that requestHints/ParameterHints reqests are sent before the textSynchronization notification message as all these listeners work parallel. this behaviour is observed for some smallersize file also.

since we would want requestHints/ParameterHints request to be sent at the end. so I have made change in this PR that requestHints/ParameterHints request will be called from callback in setTimeout function with 0 second delay. reason 0 second delay is chosen because we just want all the queued task to be finished and then call the callback function.
https://stackoverflow.com/questions/779379/why-is-settimeoutfn-0-sometimes-useful
I refered the this webpage for this solution.

@shubhsnov  @narayani28 @swmitra  please review

### DIFF
--- a/src/languageTools/DefaultProviders.js
+++ b/src/languageTools/DefaultProviders.js
@@ -109,51 +109,49 @@ define(function (require, exports, module) {
             $deferredHints = $.Deferred(),
             self = this;
 
-        setTimeout(function(){
-            self.client.requestHints({
-                filePath: docPath,
-                cursorPos: pos
-            }).done(function (msgObj) {
-                var context = TokenUtils.getInitialContext(editor._codeMirror, pos),
-                    hints = [];
+        this.client.requestHints({
+            filePath: docPath,
+            cursorPos: pos
+        }).done(function (msgObj) {
+            var context = TokenUtils.getInitialContext(editor._codeMirror, pos),
+                hints = [];
 
-                self.query = context.token.string.slice(0, context.pos.ch - context.token.start);
-                if (msgObj) {
-                    var res = msgObj.items,
-                        filteredHints = filterWithQueryAndMatcher(res, self.query);
+            self.query = context.token.string.slice(0, context.pos.ch - context.token.start);
+            if (msgObj) {
+                var res = msgObj.items,
+                    filteredHints = filterWithQueryAndMatcher(res, self.query);
 
-                    StringMatch.basicMatchSort(filteredHints);
-                    filteredHints.forEach(function (element) {
-                        var $fHint = $("<span>")
-                            .addClass("brackets-hints");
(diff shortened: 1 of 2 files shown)


swmitra (commented):
Can you please explain why a timeout is required? Is there a problem with current call context and you want to request hints after the current call context is over?

UPDATE: I guess every LS call has a confirmation event even though it doesn't really sends any data back. We should sequence our calls based on these events instead of `setTimeout`. SetTimeout with 0 is useful only when you want to do do something after the current call context is over and you have no other means of sequencing your calls. 

@swmitra The Problem is when when a document changes we send "didChange" notification to LS and  also we send Code/Parameter Hints request to LS. we want to sequence these two things. First "didChange" notifications should be sent then Code/Parameter Hints request should be sent. for Notifications LS does not send any confirmation. so I am using timeout function to make sure "didChange"Notifications should be sent to LS before Code/Parameter Hints request are sent to LS

For Notifications LS does not send any confirmation to client, only for request call LS sends response/confirmation. Text Synchronisation event can only be sent as Notifications.

shubhsnov:
@niteskum You can also try `$.Deferred().resolve().then(func).` to change the call context instead of setTimeout

@swmitra We don't get any confirmation back for text synchronization calls between Brackets and the server.

src/languageTools/DefaultProviders.js
```diff
@@ -109,49 +109,51 @@ define(function (require, exports, module) {
...
-                                    .append(_.escape(item.text))
-                                    .addClass("matched-hint"));
-                            } else {
-                                $fHint.append(_.escape(item.text));
-                            }
-                        });
-                    } else {
-                        $fHint.text(element.label);
-                    }
+        setTimeout(function(){
```
shubhsnov: @niteskum you should just send the document change event explicitly from here, which will ensure that the aync calls enter the event queue in the correct order. setTimeout doesn't guarantee that requestHints will always happen after notifyTextDocumentChanged. 

Just paste the snippet below above the requestHints call.
``` 
//Make sure the document is in sync with the server
this.defaultCodeHintProviders.client.notifyTextDocumentChanged({
    filePath: docPath,
    fileContent: editor.document.getText()
});
```

shubhsnov (changes requested):

src/languageTools/DefaultProviders.js
niteskum: changed , please check this once in your machine also.

shubhsnov (approved):

PR #14724, opened 2019-04-24
Reinitiate tooling service on Node process crash

shubhsnov:
All tooling extensions dependent on LanguageClient module in Brackets must only initiate tooling service post the initialization of the module on Node side. On cold launch, we were waiting for the initialization and only then loading the respective client domain. However, when the node process crashes due to any issue, both the language server, as well as the initialized state in the node process are lost. The client then becomes useless.

Solution: Broadcasting an event in the Brackets context to inform the client to reinitiate tooling service once the node process is back up and initialized with LanguageClient module information.

Adding reference handling of the event in the PhpTooling extension.

ping @narayani28 @swmitra @niteskum for review.


swmitra (approved):
Great job.

PR #14728, opened 2019-04-25
Fix for CodeHint Description CSS Issue

shubhsnov:
To calculate the offset for the code hint description I was using a dummy div earlier. This was causing the parent element (anchor for the codehint menu) to inherit the height of the dummy div and overlap with certain area of the editor outside of the codehint list. Due to this we were not able to put cursor on the overlapped area.

Solution: Fixing this issue by using a modified approach which doesn't require a dummy div. Instead, we rely on absolute positioning and adjust the top margin on the description box according to codehint menu height.

ping @swmitra @narayani28 @niteskum for review.

### DIFF
--- a/src/editor/CodeHintList.js
+++ b/src/editor/CodeHintList.js
@@ -266,16 +266,12 @@ define(function (require, exports, module) {
             // attach to DOM
             $parent.append($ul);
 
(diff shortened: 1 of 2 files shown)


swmitra (approved):
Good job, again!

PR #14732, opened 2019-05-02
Added vscode-languageserver-protocol Notice

niteskum:
@shubhsnov  please review


narayani28 (commented):
Have we checked this with the legal team for the usage?

PR #14784, opened 2019-05-07
Adding 64 bit binaries for CodeHelper.exe

g-217:
have changed the icon for this binary to be same as brackets, corresponding to brackets-shell/Windows-x64 branch [PR](https://github.com/adobe/brackets-shell/pull/664)

Brackets-process


gautam0217:
@swmitra @vickramdhawal @shubhsnov please have a look, this is corresponding to brackets-shell [PR](https://github.com/adobe/brackets-shell/pull/664)

gautam0217:
My changes are in [forked repo](https://github.com/gautam0217/vscode-filewatcher-windows)

vickramdhawal:
Should the forked repo be brought under or adobe org ? It will be difficult to track the source of this change. Even the readme or wiki needs to be updated to identify the source of the 64bit binary.

shubhsnov:
VSCode 64bit binary is still using the 32bit filewatcher, and since you haven't changed anything specific in the source, rather just the build platform, I'm not sure about this. 

A better approach would be to raise these forked changes as a PR against upstream, and see the response there.

gautam0217:
Yes "Any CPU" is good enough. So, don't need 64 bit build,

PR #14798, opened 2019-05-21
 Disable the JSRefactor Menu based on context

shubhsnov:
@swmitra for review

+@narayani28

### DIFF
--- a/src/extensions/default/JavaScriptRefactoring/main.js
+++ b/src/extensions/default/JavaScriptRefactoring/main.js
@@ -166,6 +166,31 @@ define(function (require, exports, module) {
             CommandManager.register(Strings.CMD_REFACTORING_GETTERS_SETTERS, REFACTORCREATEGETSET, _.partial(_handleRefactor, REFACTORCREATEGETSET));
             subMenu.addMenuItem(REFACTORCREATEGETSET);
             Menus.getMenu(menuLocation).addMenuItem(REFACTORCREATEGETSET);
+            
+            $("#editor-holder").on("contextmenu", function (e) {
+                console.log("here");
+                var editor = EditorManager.getActiveEditor();
+        
+                // Logging should be done only when the context is javascript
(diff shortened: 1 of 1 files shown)


navch:
Are we intending to disable only context menus, not from the menu bar?

shubhsnov:
There are few issues here: Until now we used to provide features for JS only, in which case it worked to have these shortcuts and triggers be a part of JS. But we're planning to incorporate other languages now, so the question is - what would happen when we want to provide refactoring for PHP or even TS.

These features should be handled centrally and the extension should only work as a provider.
So we might need to move these to the core.

As for menu bar, that is actually a bit complex. We do have capture event for the context menu but not the menu bar since that is shell based. I'll see what I can do.

navch:
Sounds good 👍  Agreed this should be handled centrally. Maybe we can also think of removing these menus from menus bar as after adding these menus Edit looks really odd.

src/extensions/default/JavaScriptRefactoring/main.js
```diff
@@ -166,6 +166,30 @@ define(function (require, exports, module) {
...
+                // Logging should be done only when the context is javascript
+                if (!editor || editor.getModeForSelection() !== "javascript") {
+                    CommandManager.get(REFACTOR_RENAME).setEnabled(false);
+                    CommandManager.get(EXTRACTTO_VARIABLE).setEnabled(false);
+                    CommandManager.get(EXTRACTTO_FUNCTION).setEnabled(false);
+                    CommandManager.get(REFACTORWRAPINTRYCATCH).setEnabled(false);
+                    CommandManager.get(REFACTORWRAPINCONDITION).setEnabled(false);
+                    CommandManager.get(REFACTORCONVERTTOARROWFN).setEnabled(false);
+                    CommandManager.get(REFACTORCREATEGETSET).setEnabled(false);
+                    return;
```
swmitra: Please handle using else to keep the code more logical.

swmitra (changes requested):

PR #14888, opened 2019-10-08
Hacktoberfest - Update setup_for_hacking.bat

anandprabhakar0507:
ECHO (in the folder containing Brackets.exe)


Wikunia:
Suggest to label this as invalid and close the PR as this is clearly only for a Hacktoberfest reward.

PR #14935, opened 2019-10-25
Reduce code size

tokirmanva22:
modify unuseful things in code reduce size of code


sprintr:
This is not clean code and it's unnecessary.

PR #14950, opened 2019-11-01
Complete translation of a section

evertonrbraga:
Section: "Strings for LanguageTools Preferences"

PR #14951, opened 2019-11-01
SHA of the last commit

evertonrbraga:


Lominean:
Add this change to #14950 instead.

PR #14961, opened 2019-11-11
Remove space between localhost: and port in infobar message

g-217:
Remove space between localhost: and port in infobar message.
This was pending from https://github.com/adobe/brackets/pull/14956


jha-g:
@shubhsnov

swmitra (approved):

PR #14965, opened 2019-11-13
Changing version to 1.14.1 for next release

g-217:
Will change the version back to 1.15.0 after the release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
     "name": "Brackets",
-    "version": "1.15.0-0",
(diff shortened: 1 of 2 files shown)


niteskum:
@jha-g I think a separate branch would be created for 1.14.1 release, This PR should be raised directly to that branch. Please check how it was done for 1.13.1 release

jha-g:
I think we're using tags for release.

niteskum:
for 1.13.1 release version was not changed but only build name was changed to 1.13.1

PR #14975, opened 2019-11-20
Moving command line port validation error message to infobar

g-217:
Moving command line port validation error messages to infobar.

Also CEF version used for Brackets, lack proper port validation logic(It is fixed from 3202), CEF relies on Operating System to block opening reserved ports:
This allows Brackets to open any positive number port less than 65535 on Windows, so for Win port range is [1,65534].
Mac and Linux do throw error while calling bind() on reserved port, so for Mac and Linux port range is [1024,65534].

Despite issues with Windows port range, we should continue showing valid port range to be [1024,65535] on info-bar.

### DIFF
--- a/src/brackets.js
+++ b/src/brackets.js
@@ -257,12 +257,19 @@ define(function (require, exports, module) {
             );
         }
 
-        brackets.app.getRemoteDebuggingPort(function (err, remote_debugging_port){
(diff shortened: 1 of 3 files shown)


jha-g:
As per discussion with @swmitra, removed port number suggestion from error message, as port suggestion for Windows will not be congruous with general understanding of network ports.

PR #15018, opened 2020-01-20
Fix for parameter hints not showing when using tabs instead of spaces

SietseB:
Parameter hints are not showing when using the tab character for indents.
In case of spaces a column number equals the character position in the line string, but in case of tabs that trick doesn't work.
This fix converts a column number to a character position, taking tabs into account.

### DIFF
--- a/src/JSUtils/Session.js
+++ b/src/JSUtils/Session.js
@@ -394,6 +394,43 @@ define(function (require, exports, module) {
                     lex.prev.info === "call")));
         }
 
+        /**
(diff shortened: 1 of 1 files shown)


swmitra (approved):
Interesting solution to handle tabs 👍   
Change looks good to me, please test JS intellisense features once  before merging.

yogo1212:
@SietseB i've squashed and merged your changes into https://github.com/brackets-cont/brackets

PR #15035, opened 2020-02-17
Adding value "legacy" to justify-items and removing "unset" from justify-items and justify-self as per w3 org recommendations.

narayani28:
Adding value "legacy" to justify-items and removing "unset" from justify-items and justify-self as per w3 org recommendations.


ghost (approved):

sobisht (approved):
LGTM 👍

PR #15038, opened 2020-02-17
Open .xd file in Adobe XD app on double click in FileTree.

niteskum:
in this PR change is on Brackets Launch It will detect If XD is installed. If it is installed then a Dialog will be shown to user asking If user likes to open .xd file in Adobe XD as aexternal editor. If users says yes then  double click on .xd file in file tree will open file in Adobe XD App. 

@swmitra  @shubhsnov  @jha-g  @sobisht  @narayani28  Please review.

### DIFF
--- a/src/brackets.js
+++ b/src/brackets.js
@@ -89,7 +89,7 @@ define(function (require, exports, module) {
     require("project/WorkingSetSort");
     require("search/QuickOpen");
     require("search/QuickOpenHelper");
-    var FileUtils = require("file/FileUtils");
+    require("file/FileUtils");
     require("project/SidebarView");
     require("utils/Resizer");
     require("LiveDevelopment/main");
@@ -101,6 +101,7 @@ define(function (require, exports, module) {
     require("language/XMLUtils");
     require("language/JSONUtils");
     require("widgets/InlineMenu");
+    var HealthLogger = require("utils/HealthLogger");
 
     // DEPRECATED: In future we want to remove the global CodeMirror, but for now we
     // expose our required CodeMirror globally so as to avoid breaking extensions in the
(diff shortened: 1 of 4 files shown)


src/project/FileTreeView.js
```diff
@@ -569,6 +571,10 @@ define(function (require, exports, module) {
                 if (this.state.clickTimer !== null) {
                     this.clearTimer();
                 }
+                if("xd" === FileUtils.getFileExtension(this.myPath()).toLowerCase() && PreferencesManager.getViewState("OpenXDFileInXDApp")) {
+                    brackets.app.OpenDocumentWithXDApp(this.myPath(), function () {});
```
shubhsnov: See my comment on brackets-shell, you can make this callback optional by using _dummyCallback

src/project/FileTreeView.js
```diff
@@ -39,7 +39,8 @@ define(function (require, exports, module) {
         LanguageManager   = require("language/LanguageManager"),
         FileTreeViewModel = require("project/FileTreeViewModel"),
         ViewUtils         = require("utils/ViewUtils"),
-        KeyEvent          = require("utils/KeyEvent");
+        KeyEvent          = require("utils/KeyEvent"),
```
shubhsnov: Why is this included?

src/brackets.js
```diff
@@ -371,6 +371,42 @@ define(function (require, exports, module) {
             });
         });
 
+        // Check If Adobe XD Application is installed on User Machine
```
shubhsnov: Should we really pollute `brackets.js` with this piece of code? Can we think about moving this somewhere else?
@swmitra for thoughts?

src/file/FileUtils.js
```diff
@@ -526,6 +531,23 @@ define(function (require, exports, module) {
...
+     */
+    function canBracketsOpenFileWithGivenExtensions(ext) {
+        return !extListNotSupportedInBrackets.includes(ext);
+    }
+
+    /**
+     * @param {string} ext File Extensions to be added Not Supported List
+     *
+     */
+    function addExtensionToNotSupportedList(ext) {
```
shubhsnov: This feels a bit unnecessary since only `xd` will be added to the list and added functionality will never be used again.

src/brackets.js
shubhsnov: Perhaps we can move all this to an extension without changing any code files?

src/file/FileUtils.js
niteskum: Fixed

PR #15088, opened 2020-03-05
Feature to associate file types with external editors

niteskum:
in this PR, added an extension to which read User preferences "externalApplications" wher users can associate an external Applications to any file type:

sample Prefernces:
"externalApplications": {
 "jpg": "default",
        "doc": "notepad"
 }

in this case double click on a jpg file in file tree, jpg file will be opened in its associated os default Application and ,doc file will be opened in notepad Application.

@swmitra @shubhsnov  @sobisht  @narayani28  Please review

### DIFF
--- a/src/extensions/default/OpenWithExternalApplication/node/package.json
+++ b/src/extensions/default/OpenWithExternalApplication/node/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "brackets-open-external_editor",
+  "name": "brackets-open-external_application",
   "dependencies": {
     "open": "0.0.5"
   }
--- a/src/file/FileUtils.js
+++ b/src/file/FileUtils.js
@@ -537,7 +537,7 @@ define(function (require, exports, module) {
      *
      */
     function shouldOpenInExternalApplication(ext) {
-        return !extListToBeOpenedInExtApp.includes(ext);
+        return extListToBeOpenedInExtApp.includes(ext);
     }
 
     /**
@@ -547,7 +547,7 @@ define(function (require, exports, module) {
     function addExtensionToExternalAppList(ext) {
 
         if(typeof ext !== 'string') {
-            extListToBeOpenedInExtApp.concat(ext);
+            extListToBeOpenedInExtApp = ext;
         } else {
             extListToBeOpenedInExtApp.push(ext);
         }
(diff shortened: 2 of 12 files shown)


src/extensions/default/OpenWithExternalApplication/node/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "brackets-open-external_application",
+  "dependencies": {
+    "open": "0.0.5"
```
jha-g: open module does not seem to be latest.
Quoting https://www.npmjs.com/package/open
```
Note: The original open package was previously deprecated in favor of this package, and we got the name, so this package is now named open instead of opn. If you're upgrading from the original open package (open@0.0.5 or lower), keep in mind that the API is different.
```

src/file/FileUtils.js
```diff
@@ -526,6 +531,28 @@ define(function (require, exports, module) {
...
+        return extListToBeOpenedInExtApp.includes(ext);
+    }
+
+    /**
+     * @param {string} ext File Extensions to be added in External App List
+     *
+     */
+    function addExtensionToExternalAppList(ext) {
+
+        if(typeof ext !== 'string') {
```
jha-g: Please add check for list instead.

g-217 (changes requested):
Please have a look at node_module named open.

src/extensions/default/OpenWithExternalApplication/node/package.json
niteskum: above 0.0.5 , open module version starts with 6.0.0. Open module starting 6.0.0 is not compatible with node version 6.11.0.

It gives below error  when requiring module:
 const wslToWindowsPath = async path => {
                               ^^^^

SyntaxError: Unexpected identifier
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:549:28)
    at Object.Module._extensions..js (module.js:586:10)
    at Module.load (module.js:494:32)
    at tryModuleLoad (module.js:453:12)
    at Function.Module._load (module.js:445:3)
    at Module.require (module.js:504:17)
    at require (internal/module.js:20:19)

src/file/FileUtils.js
niteskum: Fixed.

g-217 (approved):

PR #15092, opened 2020-03-13
Detect Graphics file in project and associate it with OS default application

niteskum:
in this PR Graphics File with type ["jpg", "jpeg", "png", "svg", "xd", "psd", "ai"] are checked in project , if files are present , a dialog is shown to user if user would like to associate these graphics file type to os default app. If user click yes, Brackets check OS Default app for each type mentioned above and associate it.

@jha-g @shubhsnov @sobisht @swmitra @narayani28 please review


jha-g:
LGTM

g-217 (approved):

PR #15096, opened 2020-03-19
Path is converted to windows style when opening with external editor in win

niteskum:
some Applications like Illustrator can not open file represented in Posix style. so converting path to windows style for windows platform.

@sobisht  @narayani28  @jha-g  please review.

### DIFF
--- a/src/extensions/default/OpenWithExternalApplication/main.js
+++ b/src/extensions/default/OpenWithExternalApplication/main.js
@@ -48,9 +48,16 @@ define(function (require, exports, module) {
 
     var extensionToExtApplicationMap = {};
 
+    function convertUnixPathToWindowsPath(path) {
+        if (brackets.platform === "win") {
+            path = path.split("/").join("\\");
+        }
+        return path;
+    }
+
     function _openWithExternalApplication(event, path) {
         _nodeDomain.exec("open", {
-            path: path,
+            path: convertUnixPathToWindowsPath(path),
             app: extensionToExtApplicationMap[FileUtils.getFileExtension(path).toLowerCase()]
         });
     }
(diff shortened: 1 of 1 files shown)


src/extensions/default/OpenWithExternalApplication/main.js
```diff
@@ -48,9 +48,16 @@ define(function (require, exports, module) {
 
     var extensionToExtApplicationMap = {};
 
+    function convertUnixPathToWindowsPath(path) {
+        if (brackets.platform === "win") {
+            path = path.split("/").join("\\");
```
jha-g: I think we should write
path.replace(/\//g, '\\')
Also since we are converting path to Windows, should we ensure first char is a drive letter?

niteskum: path = path.replace(/\//g, '\\'); this don't work , tried it.

Other suggestion to check first char is a drive done,

jha-g: Check this:
    FileSystem.isAbsolutePath = function (fullPath) {
        return (fullPath[0] === "/" || (fullPath[1] === ":" && fullPath[2] === "/"));
    };

jha-g: Also checked, and this is working:
``
'C:/Users/gjha/Desktop/test.js'.replace(/\//g, '\\')
"C:\Users\gjha\Desktop\test.js"
'C:/Users/gjha/Desktop/test.js'.replace(RegExp('/','g'), '\\')
"C:\Users\gjha\Desktop\test.js"
```

niteskum: @jha-g  made changes as above info.

g-217 (approved):

PR #15098, opened 2020-03-19
upgrading lodash node module to 4.17.15 from 4.17.4

niteskum:
to address reported security vulnerability, Upgrading lodash node module version  to 4.17.15 from 4.17.4.

@jha-g  @narayani28  please review


g-217 (approved):
LGTM

PR #15100, opened 2020-03-23
Windows path correction

niteskum:
Windows network path also need to be converted to windows style path
@jha-g  @narayani28  please review

### DIFF
--- a/src/extensions/default/OpenWithExternalApplication/main.js
+++ b/src/extensions/default/OpenWithExternalApplication/main.js
@@ -49,7 +49,8 @@ define(function (require, exports, module) {
     var extensionToExtApplicationMap = {};
 
     function convertUnixPathToWindowsPath(path) {
-        if (brackets.platform === "win" && path && path[1] === ":" && path[2] === "/") {
+        if (brackets.platform === "win"
+        && path && ((path[1] === ":" && path[2] === "/") || (path[0] === "/" && path[1] === "/"))) {
            path = path.replace(RegExp('/','g'), '\\');
         }
         return path;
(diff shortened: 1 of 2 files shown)


src/extensions/default/OpenWithExternalApplication/main.js
```diff
@@ -49,7 +49,8 @@ define(function (require, exports, module) {
     var extensionToExtApplicationMap = {};
 
     function convertUnixPathToWindowsPath(path) {
-        if (brackets.platform === "win" && path && path[1] === ":" && path[2] === "/") {
+        if (brackets.platform === "win"
```
jha-g: What is the reason for this change? Windows network paths?

niteskum: when we open network file with some external application, it does not accept path with forward-slash "/", so need to converted to windows style.

e.g

Illustrator //<network_path_with_forward_slash>/example.ai    this does not work

so have to give like : Illustrator \\<network_path_with_backward_slash>\example.ai

jha-g: Can't we use the condition  `brackets.platform === "win" && FileSystem.isAbsolutePath()`.

jha-g: Also are we planning to remove leading slashes from paths like `/C:/notepad/notepad.exe`?
And more than two leading slashes do not work for network path, so we are covered there.

niteskum: Here we will never have local path with leading slashes so not required.

g-217 (approved):

PR #15194, opened 2020-08-04
Update Korean version index.html

choibyeol:
The existing translation is very good, but there are some grammatical errors, so I corrected it.


KJ-honey:
번역 검수 고생하셨습니다ㅎㅎ

PR #15254, opened 2020-11-26
Add a value for justify-content property

PraneetDixit:
Add "space-evenly" as value for "justify-content" property


yogo1212:
@PraneetDixit  i've merged your changes into https://github.com/brackets-cont/brackets

PR #15303, opened 2021-01-08
Update README.md

praphullagogula:


iamjatinchauhan (commented):
There's no need to include a newline here. The description is already enough.

iamjatinchauhan (commented):
Readme is already Excellent. Addition of new line is not required here. Issue #15303 can be closed.

